PR #3, opened 2015-07-24
Fixing syntax error in actionScript sdk generator

pgilmorepf:
Vectors/arrays had an incorrect syntax when used within the copy-constructor.
This fixes the issue where ActionScriptSDK won't compile.


hungwunfai:
LGTM

PR #4, opened 2015-07-31
Adding a testing framework and API tests to CSharpSDK

pgilmorepf:
Adding a testing framework and API tests to CSharpSDK


pgilmorepf:
From IM conversation, again, I'll do this in a future commit.  It's not a huge problem as long as it's fixed "soon".

PR #12, opened 2015-08-07
Http optimization update unity sdk

MarcoWilliamsPF:


pgilmorepf:
Unity.StartCoroutine - This is incompatible with my current automated testing framework.  Coroutines still run on the main thread, so they really don't serve a purpose, since it won't separate the work to another thread.  They will not improve performance, they only serve to offload the work to a future unity-tick (Which makes my testing system pointless).  The callback receivers can be running in a coroutine, but really that's the only coroutine that should be running.

Overall, I just don't think that this thing is structured correctly to achieve optimal performance.  You haven't really changed the original thread-structure.  We'll get a lot more performance out of threading if we change how we're calling and using the threads.  (We discussed that before, and how I'll probably take that on later, but I'm re-iterating the point)

I don't think the synchronous and non-synchronous calls need to be two big copy-paste blocks.  The main api-call logic could be 1 sub-function.

There's inconsistent space/tab usage in files, making it not line up correctly.  We should do a mass replace at some point: "/t"->"_ _ _ _".  Some new json files have this problem.

This new json library uses System.Linq.  This is not a good sign for performance.

For now, LGTM, but we still have plenty of performance work to go, as well as some revisions to make it compatible with automated testing.

PR #13, opened 2015-08-07
Updating the build scripts that build the SDKs

pgilmorepf:


hungwunfai:
Why hardcoded path? What if someone else want to run the script and build SDK?

pgilmorepf:
I have tried using relative paths with Jenkins, and it does not like it.  Many of the environment variables also fail to work correctly.
For now, my priority is making these files work for Jenkins.

PR #15, opened 2015-08-10
API update requires rebuilding all SDKs

pgilmorepf:


hungwunfai:
If we have to use absolute path, please do leave comment to explain why and how to workaround for customers, because this repo is public.

pgilmorepf:
I will do that in my next checkin.

PR #17, opened 2015-08-11
Fixing C++ crash exception: PFWORKBIN-261

pgilmorepf:
The C++ project was not safely checking all error possibilities, and one of those possibilities caused a crash.  This is now fixed.  As well, some error conditions didn't always set the error outputs, and so specific errors would be hidden.  All of this is fixed.


pgilmorepf:
I fail at git again.
By accident, this fork request also includes: PFWORKBIN-293.
However, they're both C++ crash fixes, so it's probably all good.

MarcoWilliamsPF:
LGTM

PR #20, opened 2015-08-14
Added Project Files for NodeJS Tools For Visual Studio

MarcoWilliamsPF:
Update Unity Client API Generator to allow pass-through of Request & CustomData objects


pgilmorepf:
I need a better understanding about why ModifyUserVirtualCurrency requires a special case.

pgilmorepf:
.ntvs_analysis.dat is probably a temporary file that doesn't need to be checked in.
You didn't remove all the Currency exceptions.
 targets/csharp-unity/templates/API.cp.ejs:
<% if(apiCall.result != "ModifyUserVirtualCurrencyResult") { %>

pgilmorepf:
Spectat bonum mihi

PR #24, opened 2015-08-20
Fixing: PFWORKBIN-302 and PFWORKBIN-445

pgilmorepf:
Region (type) and Region (variable) have been adjusted so they don't have the same name - This is a requirement for android builds.
A bunch of calls to memcpy() and strncpy() have been replaced with sprintf(), which allows the Android build to compile.


MarcoWilliamsPF:
LGTM

PR #27, opened 2015-08-24
Pfworkbin 235

pgilmorepf:
The PlayFabError object in api-callbacks will now contain better information, and more consistently show an error state when there is a problem - specifically when there is no connection, the error information will be populated appropriately.


MarcoWilliamsPF:
LGTM

PR #29, opened 2015-08-25
Fixing GMFB-327

pgilmorepf:
where the UnitySDK dies if the timestamp it receives is in an alternate format.

Making the timestamp format in util.cs more consistent.
Beefing up the unit-testing around timestamp formats.
Removing some gameObject references from logic that is running outside the unity thread.


MarcoWilliamsPF:
LGTM

PR #35, opened 2015-09-23
Updating CSharp tests

pgilmorepf:
Updating the LeaderBoard test to be more reliable, adding the AccountInfo-Enum test.


zac-playfab:
LGTM, although we may want to look at future versions having better assert messages. Having more explicit messages will be self-documenting also.

PR #39, opened 2015-09-24
Adding a CloudScript test to WindowsSDK.

pgilmorepf:
This test discovers that CloudScript calls are incomplete in the last decode step.
This bug has been filed for a future date: PFICEBOX-566.


melissa-playfab:
LGTM. As an aside-- is there a bug attached to that known issue about Results.Results not populated in cloudscript? Can that be added into the comment if possible?

pgilmorepf:
Added a few more details to the Jira bug

PR #40, opened 2015-09-25
Java CloudScript test

pgilmorepf:
And other associated fixes:
Fixing cloudscript itself, which was totally busted
Fixing the error output from the SDK (so that sdk-users can get good messages when they do something wrong)


MarcoWilliamsPF:
LGTM

PR #42, opened 2015-10-11
Fixing issue where session ticket could never be set

sivakatir:
[Review on Reviewable](https://reviewable.io/reviews/playfab/sdkgenerator/42)


zac-playfab:
LGTM

PR #44, opened 2015-10-14
Addition of UE4 SDKGenerator

JLyons1985:
I would like to add an unreal-cpp target to the SDKGenerator. A SDKBuildScript can be found in the scripts folder as well. This generator will build all the necessary plugin files as well as create an example project with the plugin included.


pgilmorepf:
This pull was temporarilly redirected into the https://github.com/PlayFab/SDKGenerator/tree/UnrealBlueprints branch.

PR #55, opened 2015-11-09
Bug T45

pgilmorepf:
Fixing the WinC++ spacing to more closely match the visual studio default/automatic spacing.
Fixing the T45 issue where C++ could not remove keys from UserData (among others, like sharedGroupData, itemCustomData, etc)
Adding a unit test to verify that UserData specifically works (which was the original reported bug).


zac-playfab:
LGTM

PR #56, opened 2015-11-11
Merging the UnrealCPP sdk generator into master

pgilmorepf:


pgilmorepf:
Brendan on Slack:
And I've finished the review, by the way - not seeing any obvious issues. I'd say let's ship it, and I can work with the folks who are using C++ Unreal on digging in on it.

PR #61, opened 2015-11-23
Fixing a minor issue with apis that were not hidden properly.

pgilmorepf:
Fixing an issue where ActionScript is submitting excessive files.
The Xbox apis aren't live yet, and they were pulled out of API-SPECs.  AS3 creates files for each request and result, and the old script didn't clean them up.
We can more rigorously handle api deletions with this adjustment.  (Deletions should only happen for apis that shouldn't be released yet, never published ones)


MarcoWilliamsPF:
LGTM

PR #63, opened 2015-12-11
google advertisingId and ios IDFA work

pgilmorepf:


pgilmorepf:
Zac already accepted this in:
https://phab.playfabdev.com/D358

But arcanist can't land the diff, so I'm doing it manually here.

PR #65, opened 2016-01-14
Update cocos2dx version from mingle games

pgilmorepf:


pgilmorepf:
Master is the wrong branch.

PR #68, opened 2016-01-18
Fix the oop for the post-response callbacks and data flow.

pgilmorepf:
Big surprise, arcanist won't land this one either.


pgilmorepf:
Approved by zac:
https://phab.playfabdev.com/D504
Merged manually because arcanist is stupid.

PR #78, opened 2016-02-12
Simplify PlayFabAPI blueprints and minimize json decoding steps

tgraupmann:
The optimized blueprints use delegates which send the decoded results to custom events. This reduces the number of steps when using the API.


pgilmorepf:
I have executed the equivalent of this pull request into a temporary testing branch.  I'm closing this pull request since I don't accept merges directly into master.  You can see the completed merged here:
https://github.com/PlayFab/SDKGenerator/pull/80

I will begin testing it immediately, and push it to master when I'm done.  Thank you for your submission.

PR #90, opened 2016-04-01
Json multi serializer

MarcoWilliamsPF:


pgilmorepf:
Please remove the .idea folder.

pgilmorepf:
My first reaction to makeStrangeIoC in generate.js is very negative.  I'm against putting sdk-specific code in generate.js.  That's the entire point of the make.js files is to be that customization.

targets/csharp-unity/source/Internal/ISerializer.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        string SerializeObject(object json);
+        string SerializeObject(object json, object jsonSerializerStrategy);
+    }
+
+    public class SimpleJson
+    {
+        private static ISerializer _instance = new SimpleJsonInstance();
+
+        public static ISerializer Instance
```
pgilmorepf: I don't really see the need for this.

pgilmorepf: This needs a docstring to indicate that this is the global reset for changing out the serializer.

generate.js
```diff
@@ -134,6 +141,22 @@ function generate(args) {
...
+
+        if (targetMaker.makeGameServerAPI) {
+            apiOutputDir = targetMaker.putInRoot ? sdkOutputDir : path.resolve(sdkOutputDir, "PlayFabSDK");
+            console.log(" + Generating GameServer to " + apiOutputDir);
+            if (!fs.existsSync(apiOutputDir))
+                mkdirParentsSync(apiOutputDir);
+            targetMaker.makeGameServerAPI(gameServerApis, targetSourceDir, apiOutputDir);
+        }
+
+        if (targetMaker.makeStrangeIoC) {
```
pgilmorepf: My first reaction to makeStrangeIoC in generate.js is negative. This seems sdk-specific, which is the entire point of the make.js files.

.idea/.name
```diff
@@ -0,0 +1 @@
+SDKGenerator
```
pgilmorepf: Please remove the .idea folder, or add it to .gitignore.

pgilmorepf:
3 requested changes in specific locations.
Also, just because we should, but not because I'm any better:  Please put distinct features in separate checkins.

targets/csharp-unity-gameserver/make.js
```diff
@@ -0,0 +1,419 @@
+var path = require("path");
+
+exports.putInRoot = true;
+exports.makeServerAPI = function (apis, sourceDir, apiOutputDir) {
+    var gameServerApis = [];
+    for(var i=0; i<apis.length; i++){
+        var api = apis[i];
+        if(apis.name != "Admin"){
```
pgilmorepf: It would be nicer long term to filter positively instead of negatively.
If === "Server" and if === "Matchmaker".

pgilmorepf:
LGTM

PR #93, opened 2016-05-17
Pass developer key by command line

FineRedMist:
We have made some updates to our implementation that removes the need to have the developer key (and other parameters) in source. They can be passed by command line instead.


pgilmorepf:
We evaluated your changes to UE-cpp.
Strictly speaking, they're good changes for this specific SDK given the way this SDK is built.

But, we're declining the developerKey-commandLine change for two reasons:
1) It diverges further from the established pattern in the other SDKs.  The other SDKs solve this problem by having a separate client/server/combined instance.
2) This forces users to change the way they build/start their existing servers.  We now have 5 studios with titles running this SDK, so we have to consider all studios when evaluating a change.  Those other 4 studios would have to redesign the way they start their servers if we accept this change.

This is a rough decision, especially since you guys authored the SDK in the first place.

Sorry guys.  :(

PR #94, opened 2016-05-17
C#-flat SDK test fixes

FineRedMist:
Fixes the test cases for the C# flat SDK.

Picks up additional code gen and test cases from the base C# SDK.

Note: 
* I was missing the helloWorld cloud script resulting in that test failing
* The user that was created had NeedsAttribution set to false causing the advertising test to fail. I tweaked the test to check the NeedsAttribution value before asserting matching IDs.


pgilmorepf:
Please see:
https://github.com/PlayFab/SDKGenerator/pull/97
and
https://github.com/PlayFab/SDKGenerator/pull/98

I pulled your changes into a local branch, and resolved them with our updates to SdkGenerator before I pushed to master.  I'm closing this request under the assumption that it's covered by the two pull requests above.

FineRedMist:
I'll double check, they may not be...

PR #101, opened 2016-07-22
Fix for null ref issue

MarcoWilliamsPF:
Fix for corrupted ProductionEnvironmentUrl

### DIFF
--- a/targets/unity-v2/templates/PlayFabSettings.cs.ejs
+++ b/targets/unity-v2/templates/PlayFabSettings.cs.ejs
@@ -50,7 +50,7 @@ namespace PlayFab
 
         public static string ProductionEnvironmentUrl
         {
-            get { return PlayFabShared.ProductionEnvironmentUrl; }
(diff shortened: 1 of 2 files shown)


targets/unity-v2/templates/PlayFabSettings.cs.ejs
```diff
@@ -50,7 +50,7 @@ namespace PlayFab
 
         public static string ProductionEnvironmentUrl
         {
-            get { return PlayFabShared.ProductionEnvironmentUrl; }
+            get { return !string.IsNullOrEmpty(PlayFabShared.ProductionEnvironmentUrl) ? PlayFabShared.ProductionEnvironmentUrl : ".playfabapi.com"; }
```
pgilmorepf: return PlayFabShared.ProductionEnvironmentUrl ?? ".playfabapi.com";

pgilmorepf:
Please revise.

pgilmorepf:
I was wrong about the comment above, but we've decided to try another solution that's probably better.  Closing pull request.

PR #102, opened 2016-08-09
Disable Unique Identifier - Removes required phone state permission

MarcoWilliamsPF:


targets/unity-v2/templates/PlayFabSettings.cs.ejs
```diff
@@ -48,6 +48,14 @@ namespace PlayFab
         }
 #endif
 
+        private static string _uniqueId;
+        public static string DeviceUniqueIdentifier
+        {
+            get { return _uniqueId; }
+            set { _uniqueId = value; }
+        }
```
pgilmorepf: This shouldn't be a property, and the underlying field shouldn't exist.
It should just be:
public static String DeviceUniqueIdentifier;

targets/unity-v2/source/Shared/Public/PlayFabDataGatherer.cs
```diff
@@ -97,7 +97,11 @@ public void GatherData()
             DeviceName = SystemInfo.deviceName;
             DeviceType = SystemInfo.deviceType;
 
-            DeviceUniqueId = SystemInfo.deviceUniqueIdentifier;
+#if !DISABLE_UNIQUEIDENTIFIER
+            DeviceUniqueId = PlayFabSettings.DeviceUniqueIdentifier ?? SystemInfo.deviceUniqueIdentifier;
+#else
+            DeviceUniqueId = PlayFabSettings.DeviceUniqueIdentifier;
+#endif
```
pgilmorepf: I'd rather have an ENABLE style #define, rather than a DISABLE, given the perceived risk of the permission it requires.
And I think the PlayFabSettings.DeviceUniqueIdentifier is clunky and deceptive.
Like many other fields in this file, it should simply not exist at all, if it's not available.

pgilmorepf:
Denied

targets/unity-v2/templates/PlayFabSettings.cs.ejs
```diff
@@ -48,6 +48,25 @@ namespace PlayFab
...
+        {
+            get
+            {
+                var deviceId = "";
+#if UNITY_ANDROID && !UNITY_EDITOR
+                 AndroidJavaClass up = new AndroidJavaClass ("com.unity3d.player.UnityPlayer");
+                 AndroidJavaObject currentActivity = up.GetStatic<AndroidJavaObject> ("currentActivity");
+                 AndroidJavaObject contentResolver = currentActivity.Call<AndroidJavaObject> ("getContentResolver");
+                 AndroidJavaClass secure = new AndroidJavaClass ("android.provider.Settings$Secure");
+                 string deviceId = secure.CallStatic<string> ("getString", contentResolver, "android_id");
```
pgilmorepf: Compiler error, you've defined "deviceId" twice.

pgilmorepf:
Denied.  Try again.

pgilmorepf:
LGTM

PR #105, opened 2016-10-18
CloudScriptTypings target.

JoshuaStrunk:
Creates a new target(cloudscript-typings) which generates a typescript declarations file for the CloudScript environment.


JoshuaStrunk:
Technically if accepted would also require the addition of the target to API_Specs/SdkManualNotes.json

pgilmorepf:
Accepted!  I need some time to get a Jenkins job and all the downstream stuff set up but it'll definitely happen.

PR #106, opened 2016-10-27
CloudScriptTypings update

JoshuaStrunk:
This is an update to focused on making the code in the newly added CloudScriptTypings target easy to read and understand. This should allow the code to be more easily translated into generating typings for the js and node sdk. 

In addition the way in which descriptions and summaries are inserted into the generated file has been tuned to be better suited to TS tools such as  atom-typscript.

### DIFF
--- a/targets/cloudscript-typings/make.js
+++ b/targets/cloudscript-typings/make.js
@@ -56,7 +56,7 @@ function MakeDatatype(datatype, api, sourceDir) {
 }
 
 /** Handles generating a property field for inside an interface */
-function GetProperty(property, api, sourceDir) {
+function GetProperty(property, indentSpaces) {
 
   var type = property.jsontype.toLowerCase();
 
@@ -77,7 +77,7 @@ function GetProperty(property, api, sourceDir) {
     postColon += "[]";
   }
 
-  return GetDescription(property.description, "   ") + "   "+ preColon + " : " + postColon + ",\n";
+  return GetDescription(property.description, indentSpaces) + indentSpaces+ preColon + " : " + postColon + ",\n";
 
 }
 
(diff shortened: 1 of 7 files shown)


targets/cloudscript-typings/make.js
```diff
@@ -26,3 +28,122 @@ exports.makeServerAPI = function (apis, sourceDir, apiOutputDir) {
...
 
 }
+
+/**
+  Handles selecting the correct datatype template to use for a datatype.
+  enum, interface
+*/
+function MakeDatatype(datatype, api, sourceDir) {
+  var stringLiteralTemplate = compileTemplate(sourceDir,"StringLiteral");
+  var interfaceTemplate = compileTemplate(sourceDir,"Interface");
```
pgilmorepf: Calls to EJS are reasonably expensive (It loads and compiles an entire file)
I would not fetch both templates here but rather...

targets/cloudscript-typings/make.js
```diff
@@ -26,3 +28,122 @@ exports.makeServerAPI = function (apis, sourceDir, apiOutputDir) {
...
+  var postColon = type;
+
+  if(property.collection === "map") {
+    postColon = "{ [key:string] : "+ type +"}";
+  }
+  if(property.collection === "array") {
+    postColon += "[]";
+  }
+
+  return GetDescription(property.description, "   ") + "   "+ preColon + " : " + postColon + ",\n";
```
pgilmorepf: You've hard coded your tab-spacing here.
I'd add a parameter that makes the spacing an input parameter of GetProperty, the same way as GetDescription.

targets/cloudscript-typings/make.js
```diff
@@ -26,3 +28,122 @@ exports.makeServerAPI = function (apis, sourceDir, apiOutputDir) {
...
+
+  var locals = {
+    name: datatype.name,
+  };
+
+  if(datatype.isenum) {
+    locals.enumvalues = datatype.enumvalues;
+    return  GetDescription(datatype.description, "")+ stringLiteralTemplate(locals);
+  }
+  else {
```
pgilmorepf: ... and the datatype template here.

targets/cloudscript-typings/make.js
```diff
@@ -26,3 +28,122 @@ exports.makeServerAPI = function (apis, sourceDir, apiOutputDir) {
...
+*/
+function MakeDatatype(datatype, api, sourceDir) {
+  var stringLiteralTemplate = compileTemplate(sourceDir,"StringLiteral");
+  var interfaceTemplate = compileTemplate(sourceDir,"Interface");
+
+  var locals = {
+    name: datatype.name,
+  };
+
+  if(datatype.isenum) {
```
pgilmorepf: ... fetch the enum template here...

pgilmorepf: (Alternately, you could fetch the templates once, and store them as a global since you're going to call them so many times)

pgilmorepf:
Minor comments.  If you make the revisions, awesome.  If not, I'll wait a day or so for you to have time to read them, otherwise I'll accept it and make the revisions myself tomorrow.

JoshuaStrunk:
Thank you for the comments. I have changed my compileTemplate function to have a compiledTemplates property on which I cache compiled templates. This should solve the issue of needlessly recompiling the templates. I have also removed the hard coded tab-spacing.

PR #107, opened 2017-02-05
Fix a typo in IPlayFabEnvironment definition. 

JoshuaStrunk:
The typo causes IPlayFabEnvionment to have an extra property `string:any` and implicitly types the titleId property to any. This is instead of having titleId be OF type string.


pgilmorepf:
Accepted with gratitude.

PR #108, opened 2017-02-05
A proposed change which creates an interface to type the handlers object.

JoshuaStrunk:
The IPlayFabHandlers interface types handlers object to more closely match its intended function.(A mapping of functions)

This would cause a semi-breaking change in that attaching a handler using `handlers.MyHandler = function(args) {}` would cause a compiler error. Instead a developer would need to use the mapping notation `handlers["MyHandler"] = function(args) {}` or use [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) to merge the new handlers declaration into the Handlers interface. 
```
interface Handlers {
   MyHandler(args:IMyHandlerRequest):void
}
handlers.MyHandler = function(args){ }
```

### DIFF
--- a/targets/SdkTestingCloudScript/templates/CloudScript.d.ts.ejs
+++ b/targets/SdkTestingCloudScript/templates/CloudScript.d.ts.ejs
@@ -1,5 +1,9 @@
 /** Static object you add your CloudScript endpoints to */
-declare var handlers: any;
+declare var handlers: Handlers;
+interface IPlayFabHandlers {
+  [handlerId:string]: (args?:any, context?:IPlayFabContext) => any;
+}
+
 /** The playfab id for the user who called into CloudScript */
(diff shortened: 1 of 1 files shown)


JoshuaStrunk:
TypeScript 2.2 actually has support for ["Dotted property for types with string index signatures"](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#typescript-22). 2.2 is currently only a [release candidate](https://blogs.msdn.microsoft.com/typescript/2017/02/02/announcing-typescript-2-2-rc/), but once it is accepted as the current stable release this would no longer be a breaking change to anyone using the latest TypeScript compiler.

pgilmorepf:
I lost track of this pull request. I will try to merge it soon.
I need to resolve it against the current test project, but I'll bump it up on my personal priority list.

PR #110, opened 2017-08-19
LumberyardSdk - Case sensitivity fixes, const correctness and conversion to AZ standard types

tkgdbowen:
Hi @pgilmorepf 

I've been looking at using the Sdk generator to reflect the PlayFab API to Lumberyard's behavior context and scripting system. As part of this change (which isn't ready yet) I've accumulated a few fixes to the generator which would be good to get back into your master, and are supplied in this pull request.

Note - a couple of these we'd already supplied in non-generator form in our last pull request.

There's a few fixes for inconsistent filename case, mainly API/Api, Sdk/SDK and bus/Bus. In this change I've standardised on "Api", "Sdk" and "Bus", though Graham mentioned you might be making similar changes your end.


pgilmorepf:
Aha!  I did not think to look in the generator for changes from you guys. I was just porting over changes from the actual repo. You caught several here, that I did not catch over there, so thanks!

I'll check these out and merge them asap.

targets/LumberyardSdk/make.js
```diff
@@ -751,7 +751,7 @@ function GetAuthParams(apiCall) {
     if (apiCall.auth === "SecretKey")
         return "\"X-SecretKey\", PlayFabSettings::playFabSettings->developerSecretKey";
     else if (apiCall.auth === "SessionTicket")
-        return "\"X-Authorization\", *mUserSessionTicket";
+        return "\"X-Authorization\", mUserSessionTicket";
```
pgilmorepf: Aha, this one returns.  I'm confused about this one.
You guys are the ones who requested it become a pointer, and not a static.
I thought the previous change was a typo, because it didn't appear to be complete (it didn't change all uses).

Could you guys please confirm whether you actually intend to change this to a static?

targets/LumberyardSdk/templates/Code/Source/PlayFabAPI.cpp.ejs
```diff
@@ -5,17 +5,22 @@
...
 {
-    return mUserSessionTicket != nullptr && mUserSessionTicket->length() != 0;
+    return mUserSessionTicket.length() != 0;
 }
 
 // #THIRD_KIND_PLAYFAB_SHUTDOWN_FIXES: - Added a logout function so that the statics can be destroyed before the system allocator.
 void PlayFabClientApi::ForgetClientCredentials()
 {
-    SAFE_DELETE(PlayFabClientApi::mUserSessionTicket);
+    // #THIRD_KIND_PLAYFAB_CRASH_WHEN_PC_IS_OFFLINE: szaluga (2017/08/05)
```
pgilmorepf: Aha, I finally see it.
OK, I'll have to double check this one, because the conversion was scattered in various pieces, but I can finally see someplace where it seems that the change was deliberate.

targets/LumberyardSdk/make.js
tkgdbowen: Hi @pgilmorepf.

Apologies for the confusion, our latest version does indeed revert the session ticket back to a static and still allows the session ticket to be unset. Overall it's a cleaner solution and safer as the pointer version was causing a crash. 

It was previously converted to a pointer because Aws::String has memory manager problems with shutting down statics. However now that we're using AZStd::string, the original pointer change is  no longer needed.

PR #111, opened 2017-09-12
UnrealCPP Changes to upgrade to 4.17

mtuska:
UnrealCPP Changes to upgrade to 4.17, removed the FMultitypeVar, removed the OnlineSubsystem(it was the main thing causing errors each engine version), and moved proxy to its own standalone plugin. The change to FJsonKeeper from FMultitypeVar is the biggest change, causing anything that used the Multitype before to be broken. Now users will have to include Json as a dependency in their build.cs, and use a Json object. To set the value of the FJsonValue, simply set the variable equal to the value you are providing. For example, if you want to send an FJsonObject to ExecuteCloudScript, simply create an FJsonObject and then set the FunctionParameter to that object. I.e. Request.FunctionParameter = JsonObject; To read from the FJsonKeeper, simply call the GetJsonValue method. From there you can convert back to a JsonObject via the AsObject method or get the value as any type that should be provided. If you don't know the type/value names simply refer back to the PlayFab API documentation for that API call. Just trying to explain that as best as I can so that you can make a guide from there.


SDKBuildScripts/ue4_CPP_build.bat
```diff
@@ -1,24 +1,24 @@
 rem === Cleaning existing files from UnrealCppSdk ===
 pushd ..\..\sdks\UnrealCppSdk
-pushd PlayFabSDK\Plugins\PlayFab\Source\PlayFabProxy\Private\Proxy
+pushd PlayFabSDK\Plugins\PlayFabProxy\Source\PlayFabProxy\Private\Proxy
```
pgilmorepf: What does the proxy do, and why move it?

targets/cpp-ue4/Plugins/PlayFab/Source/PlayFab/Public/Core/PlayFabBaseModel.h
```diff
@@ -52,81 +52,50 @@ namespace PlayFab
...
-        operator double() const { return mNumber; }
-        operator FString() const { return mString; }
-
-
-        ~FMultitypeVar() {}
-        void writeJSON(JsonWriter& writer) const override;
-        bool readFromValue(const TSharedPtr<FJsonObject>& obj) override;
-        bool readFromValue(const TSharedPtr<FJsonValue>& value) override;
-    };
+	struct PLAYFAB_API FJsonKeeper : public FPlayFabBaseModel
```
pgilmorepf: Awesome!

targets/cpp-ue4/Plugins/PlayFab/Source/PlayFab/Public/Core/PlayFabJsonHelpers.h
```diff
@@ -1,13 +1,16 @@
...
+
 namespace PlayFab
 {
 	class FPlayFabJsonHelpers
 	{
 	public:
 
 		// Returns the array named Key or nullptr if it is missing or the wrong type
-		static const TArray< TSharedPtr<FJsonValue> >& ReadArray(TSharedPtr<class FJsonObject> Item, const FString& Key);
+		static const TArray< TSharedPtr<class FJsonValue> >& ReadArray(TSharedPtr<class FJsonObject> Item, const FString& Key);
```
pgilmorepf: What does this do?

SDKBuildScripts/ue4_CPP_build.bat
mtuska: Proxy, as provided in the previous UnrealCPP SDK, implements some BP functions, however wasn't feature complete in the previous version. I moved it to its own plugin so people can throw it in plugins if wanted. Technically the proxy could be removed as well, I'm just providing the same things as before.

targets/cpp-ue4/make-bp.js
```diff
@@ -191,8 +191,7 @@ function GetBpPropertyDefinition(property, api, datatype) {
         return "TArray<" + GetPropertyUe4ToOpaqueType(property, api, datatype) + ">";
     }
     else if (property.collection === "map") {
-        //return "TMap<FString, " + getProperyUE4Type(property, api) + ">";
-        //return ""; // TODO: handle map properly, by wrapping it into a structure somehow
+        return "TMap<FString, " + GetPropertyUe4ToOpaqueType(property, api, datatype) + ">";
```
pgilmorepf: WOOOOOOO!

targets/cpp-ue4/Plugins/PlayFab/Source/PlayFab/Public/Core/PlayFabJsonHelpers.h
mtuska: I had to make changes as UE4 moves away from its monolithic way of loading classes. I am just defining the class without the headers here, and then include the headers in CPP to avoid addition overhead.

targets/cpp-ue4/templates/PlayFab.uplugin.ejs
```diff
@@ -2,7 +2,7 @@
     "FileVersion": 3,
     "FriendlyName": "<%- friendlyName %>",
     "Version": 0,
-    "EngineVersion" : "4.15.0",
+    "EngineVersion" : "4.17.0",
```
pgilmorepf: +1

targets/cpp-ue4/templates/PlayFabProxy.uplugin.ejs
```diff
@@ -0,0 +1,27 @@
...
+    "FriendlyName": "<%- friendlyName %>",
+    "Version": 0,
+    "EngineVersion" : "4.17.0",
+    "VersionName": "<%- sdkVersion %>",
+    "CreatedBy": "PlayFab and Phoenix Labs",
+    "CreatedByURL": "https://playfab.com/",
+    "DocsURL": "https://api.playfab.com/",
+    "SupportURL": "https://community.playfab.com/index.html",
+    "Category": "PlayFab",
+    "Description": "PlayFab Blueprint SDK for Unreal Engine 4. Current API version: <%- sdkVersion %>",
```
pgilmorepf: This is not the blueprint Sdk.

mtuska: Sorry, must've copied that. That is for the Proxy however, which again handles exposing the CPP to blueprint.

SDKBuildScripts/ue4_CPP_build.bat
pgilmorepf: OIC, ok.  So the proxy would be the theoretical blueprints tie-in.  NICE. That seems like a good plan for a theoretical future where we have a consistent plugin/set-of-plugins that supports both CPP and Blueprint.

targets/cpp-ue4/Plugins/PlayFab/Source/PlayFab/Public/Core/PlayFabJsonHelpers.h
pgilmorepf: Optimization.  Got it.  Sounds good.

targets/cpp-ue4/templates/PlayFabProxy.uplugin.ejs
pgilmorepf: No worries, don't fix it here, I'll fix it after.

SDKBuildScripts/ue4_CPP_build.bat
mtuska: Well, again, it was already provided before. It's still not feature complete but I plan to continue to edit it, flush it out more.

PR #112, opened 2017-09-17
Proxy Upgrade

mtuska:
Mainly upgrading the proxy to a much more usable state. Overall has a bit of weight in size, but I will continue to look at that at a later date. Other changes are mainly for packaged builds, making sure that the plugins can actually be used in shipping builds... and one or two previous errors. Lot's of newline/space changes, due to how Atom.io reads files, sorry about that, but didn't want to take all the time trying to remove those... Main thing is that, blueprints are now viable with this SDK and that it _should_ build just fine. Just drop me a line if we run into issues with the SDK and I'll take a look.


mtuska:
Ran into a strange problem with proxy and I didn't realize it, trying to correct...

mtuska:
Sorry ran into some complications and thought the plugin might have an issue. Was just on my local project, and so I reopened the pull request.

PR #113, opened 2017-11-12
Improved exceptions in PlayFab SDK for Unity

Si1ver:
Most throwed exceptions are now derived from PlayFabException and can be explicitly caugth by type in client code.

### DIFF
--- a/targets/unity-v2/make.js
+++ b/targets/unity-v2/make.js
@@ -379,11 +379,11 @@ function getRequestActions(tabbing, apiCall, api) {
 
     if (api.name === "Client" && (apiCall.result === "LoginResult" || apiCall.request === "RegisterPlayFabUserRequest"))
         return tabbing + "request.TitleId = request.TitleId ?? PlayFabSettings.TitleId;\n"
-            + tabbing + "if (request.TitleId == null) throw new Exception(\"Must be have PlayFabSettings.TitleId set to call this method\");\n";
+            + tabbing + "if (request.TitleId == null) throw new TitleIdNotSetException();\n";
     if (api.name === "Client" && apiCall.auth === "SessionTicket")
-        return tabbing + "if (!IsClientLoggedIn()) throw new Exception(\"Must be logged in to call this method\");\n";
+        return tabbing + "if (!IsClientLoggedIn()) throw new ClientNotLoggedInException();\n";
     if (apiCall.auth === "SecretKey")
-        return tabbing + "if (PlayFabSettings.DeveloperSecretKey == null) throw new Exception(\"Must have PlayFabSettings.DeveloperSecretKey set to call this method\");\n";
+        return tabbing + "if (PlayFabSettings.DeveloperSecretKey == null) throw new DeveloperSecretKeyNotSetException();\n";
     return "";
 }
 
--- a/targets/unity-v2/source/Shared/Public/PlayFabExceptions.cs
+++ b/targets/unity-v2/source/Shared/Public/PlayFabExceptions.cs
@@ -1,0 +1,51 @@
+using System;
+using System.Runtime.Serialization;
+
+namespace PlayFab
+{
+    /// <summary>
+    /// Base class for PlayFab exceptions
+    /// </summary>
+    [Serializable]
+    public class PlayFabException : Exception
+    {
+        public PlayFabException() { }
+
+        public PlayFabException(string message)
+            : base(message) { }
+
+        public PlayFabException(string message, Exception inner)
+            : base(message, inner) { }
+
+        protected PlayFabException(SerializationInfo info, StreamingContext context)
+            : base(info, context) { }
+    }
+
+    /// <summary>
+    /// Base class for PlayFab exceptions caused by incorrect settings
+    /// </summary>
+    [Serializable]
+    public class PlayFabSettingsException : PlayFabException
+    {
+        public PlayFabSettingsException(string message)
+            : base(message) { }
+
+        protected PlayFabSettingsException(SerializationInfo info, StreamingContext context)
+            : base(info, context) { }
+    }
+
+    /// <summary>
+    /// Exception that occurs when call to PlayFab APIs is made while PlayFabSettings.DeveloperSecretKey is not set
+    /// </summary>
+    [Serializable]
+    public class DeveloperSecretKeyNotSetException : PlayFabSettingsException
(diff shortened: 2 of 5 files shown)


Si1ver:
Hey, @pgilmorepf !

Please consider reviewing my PR.

The problem that I'm trying to solve is that System.Exception is used in a lot of places.
System.Exception is masking all other exception types making hard to identify the meaning and origin of catched exception.

The goal is to introduce PlayFab specific exception types in places where it makes sense. Custom base type allows to catch all PlayFab specific exceptions without catching unwanted ones. Custom types for different situations makes possible for developer to identify the cause of exception and properly handle it. 

There is also a place where standard exception type InvalidOperationException is reasonably to use.

This patch is for Unity API only, but similar changes could also be applied to C# sources.

If you are not satisfied with solution, I can improve it. Just point out what's not good enough.

targets/unity-v2/source/Shared/Public/PlayFabExceptions.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        protected PlayFabSettingsException(SerializationInfo info, StreamingContext context)
+            : base(info, context) { }
+    }
+
+    /// <summary>
+    /// Exception that occurs when more than one PlayFabSharedSettings asset is found in resources
+    /// </summary>
+    [Serializable]
+    public class MultipleSharedSettingsException : PlayFabSettingsException
```
pgilmorepf: This one conflicts with a change we're going to make very soon in the SharedSettings object. You were right to include it, but I want to preemptively avoid that conflict.  Please remove this one, and leave the place it's used as a normal exception that we'll fix on our end.

targets/unity-v2/make.js
```diff
@@ -379,11 +379,11 @@ function getRequestActions(tabbing, apiCall, api) {
 
     if (api.name === "Client" && (apiCall.result === "LoginResult" || apiCall.request === "RegisterPlayFabUserRequest"))
         return tabbing + "request.TitleId = request.TitleId ?? PlayFabSettings.TitleId;\n"
-            + tabbing + "if (request.TitleId == null) throw new Exception(\"Must be have PlayFabSettings.TitleId set to call this method\");\n";
+            + tabbing + "if (request.TitleId == null) throw new TitleIdNotSetException();\n";
```
pgilmorepf: I like the idea of PlayFabException minimally wrapping Exception, but I'm not a fan of each of these individual specific exceptions. I'd rather change this:
throw new Exception(\"Must be have PlayFabSettings.TitleId set to call this method\");
to this:
throw new PlayFabException(\"Must be have PlayFabSettings.TitleId set to call this method\");

I think it accomplishes what you want, but it is still easy for me to add new exceptions where needed, without having to define new exception types all over.

targets/unity-v2/templates/PlayFabSettings.cs.ejs
```diff
@@ -41,7 +41,7 @@ namespace PlayFab
             var settingsList = Resources.LoadAll<PlayFabSharedSettings>("PlayFabSharedSettings");
             if (settingsList.Length != 1)
             {
-                throw new Exception("The number of PlayFabSharedSettings objects should be 1: " + settingsList.Length);
+                throw new MultipleSharedSettingsException(settingsList.Length);
```
pgilmorepf: Yeah, here it is.  Please undo this one, because it'll conflict with another change we're making.

targets/unity-v2/source/Shared/Public/PlayFabExceptions.cs
```diff
@@ -0,0 +1,66 @@
+using System;
+using System.Runtime.Serialization;
+
+namespace PlayFab
+{
+    /// <summary>
+    /// Base class for PlayFab exceptions
+    /// </summary>
+    [Serializable]
+    public class PlayFabException : Exception
```
pgilmorepf: Per the comment above: I'd keep this one, and remove the more specific ones.

pgilmorepf:
Please note, if you wish your contribution to show up in the history for this project, I'd like the requested changes fairly soon.

Let's say, Friday, Dec 22, I'll do the changes myself, and commit it directly. Please respond before then if you wish to preserve your commit history.

nitreo:
Just a note: PlayFabException should also expose code, so that 4.6 users may benefit from "when" construct, rough example:
```
try
{
    // ...
}
catch(PlayFabException ex ) when (ex.Code == Codes.TitleNotSet)
{
    // ...       
}
```

Si1ver:
Hi!
Glad to hear back from you.

I've partially implemented changes requested by you.
I've also updated PR with latest changes from upstream master branch.

I'll implement remaining changes after a short discussion.

targets/unity-v2/source/Shared/Public/PlayFabExceptions.cs
Si1ver: Done!

targets/unity-v2/templates/PlayFabSettings.cs.ejs
Si1ver: Done!

targets/unity-v2/make.js
Si1ver: Sure, throwing PlayFabException is a lot better than throwing System.Exception.

The thing that I wath to achieve is also to be able to recover from some of exceptions that are not breaking internal state of PlayFabSdk.
For example,
```C#
try
{
  DoPlayfabRequest(arguments);
}
catch (TitleIdNotSetException)
{
  FixPlayfabTilleId();
}
catch (ClientNotLoggedInException)
{
  LoginPlayfabClient();
}
catch (Exception)
{
  SendReport();
  throw;
}
```

If you want, I can remove all strings from exception classes to make them as simple as possible.

If I still have not convince you to accept this change, I will leave only PlayFabException.

Si1ver: The othe way suggested by @nitreo is to add exception code to PlayFabException. This will keep the only one class for exceptions and also provide a way to differentiate them in code.

pgilmorepf: I like nitreo's suggestion, let's do that one.

pgilmorepf:
The spirit of your request has been accepted here:
https://github.com/PlayFab/SDKGenerator/pull/120
Thanks for your suggestion.

PR #125, opened 2018-02-08
GoLang Target

JoshuaStrunk:
I started this before Entity API was a thing and only just found it so it currently only generates v1 SDK. I would like to get it working with the EntityAPI asap as well as clean up any other issues. Looking for comments on how to do this.


pgilmorepf:
This pull request will need to wait until after GDC.
After GDC I'll come back and review this.
Thanks for your submission.

blackboneworks:
any news on this one?

PR #128, opened 2018-05-13
XPlatCppSdk Bugfixes and test additions.

pgilmorepf:
Http.Update will report the correct number of active API calls.
Numbers in json will [de]serialize correctly again.
Fixed timestamp [de]serialization in Windows.
Fixed some resharper complaints.

### DIFF
--- a/targets/XPlatCppSdk/source/cppsdk/include/playfab/PlayFabBaseModel.h
+++ b/targets/XPlatCppSdk/source/cppsdk/include/playfab/PlayFabBaseModel.h
@@ -85,7 +85,7 @@ namespace PlayFab
     // Utilities for [de]serializing time_t to/from json
     inline void ToJsonUtilT(const time_t input, Json::Value& output)
     {
-        struct tm timeInfo;
+        tm timeInfo;
 #ifdef _WIN32
         gmtime_s(&timeInfo, &input);
 #endif
@@ -99,12 +99,10 @@ namespace PlayFab
     inline void FromJsonUtilT(const Json::Value& input, time_t& output)
     {
         if (input == Json::Value::null) return;
-        std::string enumStr = input.asString();
+        const std::string timeStr = input.asString();
         tm timeStruct = {};
-        std::istringstream iss(enumStr.c_str());
+        std::istringstream iss(timeStr);
         iss >> std::get_time(&timeStruct, "%Y-%m-%dT%T");
-        timeStruct.tm_year -= 1900;
-        timeStruct.tm_mon -= 1;
         output = mktime(&timeStruct);
     }
     inline void ToJsonUtilT(const Boxed<time_t>& input, Json::Value& output)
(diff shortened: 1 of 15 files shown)


targets/XPlatCppSdk/source/cppsdk/include/playfab/PlayFabBaseModel.h
```diff
@@ -99,12 +99,10 @@ namespace PlayFab
     inline void FromJsonUtilT(const Json::Value& input, time_t& output)
     {
         if (input == Json::Value::null) return;
-        std::string enumStr = input.asString();
+        const std::string timeStr = input.asString();
         tm timeStruct = {};
-        std::istringstream iss(enumStr.c_str());
+        std::istringstream iss(timeStr);
         iss >> std::get_time(&timeStruct, "%Y-%m-%dT%T");
```
debhaldarMS: Should we check what happens if the parse of "iss" fails ? Can this bubble up as an exception in the game ?

targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp
```diff
@@ -226,20 +229,23 @@ namespace PlayFab
...
             reqContainer = pendingResults[pendingResults.size() - 1];
             pendingResults.pop_back();
+            activeRequestCount--;
         } // UNLOCK httpRequestMutex
 
         HandleResults(*reqContainer);
         delete reqContainer;
-        return resultCount;
+
+        // activeRequestCount can be altered by HandleResults, so we have to re-lock and return an updated value
```
debhaldarMS: I'm not seeing how HandleResults can modify activeRequestsCount - can you please help me understand ?

targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp
```diff
@@ -226,20 +229,23 @@ namespace PlayFab
...
+            if (pendingResults.empty())
+                return activeRequestCount;
 
             reqContainer = pendingResults[pendingResults.size() - 1];
             pendingResults.pop_back();
+            activeRequestCount--;
         } // UNLOCK httpRequestMutex
 
         HandleResults(*reqContainer);
         delete reqContainer;
```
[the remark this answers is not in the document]
pgilmorepf: reqContainer is entirely internal. ONLY the PlayFabHttp class touches it. It's never passed outside of PlayFabHttp. reqContainer contains the PlayFab result object, which is the only thing sent to the customer's code.

targets/XPlatCppSdk/source/cppsdk/include/playfab/PlayFabBaseModel.h
pgilmorepf: I think this is a larger discussion. The entire serialization scheme basically tries its best, and defaults to something (hopefully reasonable) if it fails. I'm open to discussing the proposal, but we would need to handle it in ALL deserializations, not just this one. It's certainly not a trivial change, and shouldn't be in this checkin.

targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp
pgilmorepf: HandleResults invokes customer code.
Customer code can do anything, including start another API call.

targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp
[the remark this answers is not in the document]
debhaldarMS: As discussed, let's log a bug for this!

targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp
debhaldarMS: agreed

PR #131, opened 2018-05-23
List of requirements to the common SDK infrastructure

Vicodex:
* Please note, this is just a list of general requirements gathered from our recent meetups/whiteboarding
* Design doc will go into deeper technical details of how to address these requirements, I will also be coding a working prototype that will go along with a design doc
* Please feel free to leave comments as in a usual code review PR

### DIFF
--- a/docs/common-sdk-requirements.md
+++ b/docs/common-sdk-requirements.md
@@ -8,10 +8,24 @@ The information listed here was primarily sourced in a few meetings/discussions/
 * _Service_ is a single building unit (module) of the new modular SDK. TODO: please let me know if another term is desired for this.
 
 ## Requirements
-### General requirements
-* text
-* text
+### Requirements to the SDK
+* Non-monolithic, modular structure that allows a greater degree of flexibility and scalability in development, maintenance, customization and extensibility due to smaller building blocks (services) that can be worked independently
+* Consists of services
+* Configurable composition:
+    * Services can be selected
+    * Services can be optional
+    * Services can be required
+    * Composition rules/restrictions apply based on each service dependencies and configuration
+    * Configuration errors must be prevented/communicated at the time of configuring
+* Can be composed and configured by a customer
+* Tracked SDK generation process for acceptable customer experience
+* Support of PlayFab-built and custom services (created by customers)
 
 ### Requirements to a service
-* text
-* text
+* Contains metadata (description of itself)
+* Configurable
+* Can be developed independently
+* Can interact with other services using:
+    * Specific service identifiers
+    * Configurable aliases (via configuration)
+* Not limited by nature (not only a part of existing SDKs)
--- a/docs/common-sdk-requirements.md
+++ b/docs/common-sdk-requirements.md
@@ -8,7 +8,7 @@ The information listed here was primarily sourced in a few meetings/discussions/
 * _Service_ is a single building unit (module) of the new modular SDK. TODO: please let me know if another term is desired for this.
 
 ## Requirements
-### Requirements to the SDK
+### Requirements for the SDK
 * Greater degree of flexibility and scalability in development, maintenance, customization and extensibility due to:
     * Non-monolithic, modular structure
     * Smaller building blocks (services) that can be worked on independently
@@ -29,7 +29,7 @@ The information listed here was primarily sourced in a few meetings/discussions/
 * Well-defined and easy new service development process
 * Well-defined and easy new service registration process
 
-### Requirements to a service
+### Requirements for a service
 * Contains metadata (description of itself that can be read by the system)
 * Configurable (a set of properties that can be set/modified by the system or customer):
     * When selecting the service before generating and downloading SDK (SDK design time)
--- a/docs/common-sdk-requirements.md
+++ b/docs/common-sdk-requirements.md
@@ -1,0 +1,17 @@
+# Requirements for the Common SDK infrastructure
+The purpose of this doc is to have a place to list and discuss all requirements to the new modular SDK infrastructure (also informally called "the Common SDK infrastructure") before creating an official design.
+
+The information listed here was primarily sourced in a few meetings/discussions/whiteboardings we had before.
+
+## Terminology
+* _Common SDK_ is the new modular SDK. 
+* _Service_ is a single building unit (module) of the new modular SDK. TODO: please let me know if another term is desired for this.
+
+## Requirements
+### General requirements
+* text
+* text
+
+### Requirements to a service
+* text
+* text
--- a/docs/common-sdk-requirements.md
+++ b/docs/common-sdk-requirements.md
@@ -4,40 +4,49 @@ The purpose of this doc is to have a place to list and discuss all requirements 
 The information listed here was primarily sourced in a few meetings/discussions/whiteboardings we had before.
 
 ## Terminology
-* _Common SDK_ is the new modular SDK. 
-* _Service_ is a single building unit (module) of the new modular SDK. TODO: please let me know if another term is desired for this.
+* _Common SDK infrastructure_ is the infrastructure necessary to build the new modular SDK.
+* _Plugin_ is a minimal building block (module) of the new modular SDK (in our prior discussions we referred to it as a _service_).
 
 ## Requirements
-### Requirements for the SDK
+### Requirements for the SDK and the infrastructure
 * Greater degree of flexibility and scalability in development, maintenance, customization and extensibility due to:
     * Non-monolithic, modular structure
-    * Smaller building blocks (services) that can be worked on independently
+    * Smaller building blocks (plugins) that can be worked on independently
     * High configurability (configuration-driven design)
 * Configurable composition:
-    * Services can be selected
-    * Services can be optional
-    * Services can be required
-    * Composition rules/restrictions apply based on each service dependencies and configuration
+    * Plugins can be selected
+    * Plugins can be optional
+    * Plugins can be required
+    * Composition rules/restrictions apply based on each plugin dependencies and configuration
     * Configuration errors must be prevented/communicated at the time of configuring
 * Ability to tailor SDK to specific customer needs
 * Can be composed and configured by a customer
-* Trackable SDK generation and download process to provide acceptable customer experience
-* Can be used to support all existing SDK platforms
 * Can be integrated into existing SDK generation process (eventually replace generation of monolithic SDKs)
-* Support of built-in (authored by PlayFab) and custom services (created by customers)
-* Service management/administration system (for PlayFab admins)
-* Well-defined and easy new service development process
-* Well-defined and easy new service registration process
+* Can be used to support all existing SDK platforms:
+    * Design is agnostic to any programming language used for plugin/SDK implementation
+    * Interaction of a plugin with the Common SDK infrastructure should be very simple and supported on all platforms
+* Plugin management:
+    * Support of built-in (authored by PlayFab) and custom plugins (created by customers)
+    * Plugin management/administration system (for PlayFab admins)
+    * Public plugin distrubution/publishing channels (future PlayFab plugin marketplace, popular package databases like nuget, npm, maven, etc)
+    * Well-defined and easy new plugin development process
+    * Well-defined and easy new plugin registration process. External parties/customers should be able to do that.
+* Customer experience:
+    * Trackable SDK generation and download process to provide acceptable customer experience
+    * If a customer is registered at PlayFab and logged in, their SDK configuration profile(s) will be saved in their account. If not, they should still be able to generate and download the SDK but their configuration profile will not be persisted on PlayFab web site (we can show a warning)
 
-### Requirements for a service
-* Contains metadata (description of itself that can be read by the system)
+### Requirements for a plugin
+* Contains metadata (description that can be read by the system)
 * Configurable (a set of properties that can be set/modified by the system or customer):
-    * When selecting the service before generating and downloading SDK (SDK design time)
-    * When adjusting the service locally after SDK is downloaded (client code at build time or runtime)
-* Can be developed independently of other services
-* Can interact with other services using:
-    * Specific unique service identifiers (reference to a specific service)
+    * When selecting the plugin before generating and downloading SDK (SDK design time)
+    * When adjusting the plugin locally after SDK is downloaded (client code at build time or runtime)
+* Can be developed independently of other plugins
+* Can interact with other plugins using:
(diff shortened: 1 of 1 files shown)


docs/common-sdk-requirements.md
```diff
@@ -0,0 +1,43 @@
...
+    * High configurability (configuration-driven design)
+* Configurable composition:
+    * Services can be selected
+    * Services can be optional
+    * Services can be required
+    * Composition rules/restrictions apply based on each service dependencies and configuration
+    * Configuration errors must be prevented/communicated at the time of configuring
+* Ability to tailor SDK to specific customer needs
+* Can be composed and configured by a customer
+* Trackable SDK generation and download process to provide acceptable customer experience
```
debhaldarMS: We'd like to remember player profile for SDK re-generation. Currently we do not require folks to be logged in to download SDKs - are we going to need that in future ? Might create a barrier to entry - if not how do we remember profiles ? Can they save their initial configuartion and re-use that ?

docs/common-sdk-requirements.md
```diff
@@ -0,0 +1,43 @@
...
+The purpose of this doc is to have a place to list and discuss all requirements to the new modular SDK infrastructure (also informally called "the Common SDK infrastructure") before creating an official design.
+
+The information listed here was primarily sourced in a few meetings/discussions/whiteboardings we had before.
+
+## Terminology
+* _Common SDK_ is the new modular SDK. 
+* _Service_ is a single building unit (module) of the new modular SDK. TODO: please let me know if another term is desired for this.
+
+## Requirements
+### Requirements to the SDK
```
debhaldarMS: nitpick - Requirements for :)

debhaldarMS: We should support programatic generation and download of the SDK -- something like : 
GenerateSDK.exe <configJson> 

The Config json could be sropped with the SDK such that if someone includes it in their build chain/ teamcity -- we can regenerate the SDK with same configurations for them.

docs/common-sdk-requirements.md
Vicodex: Option 1: We require registration at PlayFab to be able to use the new modular SDK platform (then everything will be saved in user's PlayFab account, including SDK configurations, SDK generation/download process tracking, etc)

Option 2: We could potentially provide the functionality of service selection and SDK configuration for anonymous users (public web access) and allow them to either upload/download a configuration profile as a JSON file or maybe store it in web cookies (though it may have some length limitation). Without the actual registration at PlayFab though they won't be able to do much with downloaded SDK (but it may be ok).

What do you guys think?

docs/common-sdk-requirements.md
Vicodex: This is an interesting idea! We should think about this. While SDK generation is already a scriptable command-line friendly process, making it a product that can be downloaded and used by client as a script or library so that they can use it, for example, in their automation/build systems sounds powerful.

docs/common-sdk-requirements.md
```diff
@@ -0,0 +1,43 @@
+# Requirements for the Common SDK infrastructure
+The purpose of this doc is to have a place to list and discuss all requirements to the new modular SDK infrastructure (also informally called "the Common SDK infrastructure") before creating an official design.
+
+The information listed here was primarily sourced in a few meetings/discussions/whiteboardings we had before.
+
+## Terminology
+* _Common SDK_ is the new modular SDK. 
+* _Service_ is a single building unit (module) of the new modular SDK. TODO: please let me know if another term is desired for this.
+
+## Requirements
```
debhaldarMS: Another requirement is that the Code Generated should feel as close to the platform/engine as possible. For example, while the design should be agnostic of language specifics -- the final implementation of core components and plugability model should follow familiar patterns in the language. For example, Interfaces might be the way to go in C#/Unity - but C++ might be suited to abstract base classes to work around ABI issues ( because the vtable format seldom differs among compilers) etc.

docs/common-sdk-requirements.md
```diff
@@ -0,0 +1,43 @@
...
+    * Configuration errors must be prevented/communicated at the time of configuring
+* Ability to tailor SDK to specific customer needs
+* Can be composed and configured by a customer
+* Trackable SDK generation and download process to provide acceptable customer experience
+* Can be used to support all existing SDK platforms
+* Can be integrated into existing SDK generation process (eventually replace generation of monolithic SDKs)
+* Support of built-in (authored by PlayFab) and custom services (created by customers)
+* Service management/administration system (for PlayFab admins)
+* Well-defined and easy new service development process
+* Well-defined and easy new service registration process
```
pgilmorepf: PUBLIC process.  3PP process. I want King.com to arbitrarily start uploading plugins into our infrastructure.

docs/common-sdk-requirements.md
debhaldarMS: The SDK core should use the least common denominator for the specific platform and minimize dependencies on additional packages/libraries. In C# land, this might mean we use .Net Core / Mono runtime only while in C++ we should avoid STL data passing in favor of flat C APIs for cross compatibility

docs/common-sdk-requirements.md
```diff
@@ -0,0 +1,43 @@
...
+* Contains metadata (description of itself that can be read by the system)
+* Configurable (a set of properties that can be set/modified by the system or customer):
+    * When selecting the service before generating and downloading SDK (SDK design time)
+    * When adjusting the service locally after SDK is downloaded (client code at build time or runtime)
+* Can be developed independently of other services
+* Can interact with other services using:
+    * Specific unique service identifiers (reference to a specific service)
+    * Configurable aliases (via configuration)
+* Can be used by other services (by a unique service identifier)
+* Can provide implementations for multiple programming languages/platforms
```
pgilmorepf: This is where 3pp gets really tricky. Registered services are going to include which languages/SDKs they support, and which ones they don't.  And while that's always existed, it will exaggerate the feature differences between our SDKs. (Can't be helped I guess...)

docs/common-sdk-requirements.md
```diff
@@ -0,0 +1,43 @@
...
+* Can be composed and configured by a customer
+* Trackable SDK generation and download process to provide acceptable customer experience
+* Can be used to support all existing SDK platforms
+* Can be integrated into existing SDK generation process (eventually replace generation of monolithic SDKs)
+* Support of built-in (authored by PlayFab) and custom services (created by customers)
+* Service management/administration system (for PlayFab admins)
+* Well-defined and easy new service development process
+* Well-defined and easy new service registration process
+
+### Requirements to a service
```
debhaldarMS: Should  we have a registry for plugin discoverability/ plugin registry ?

docs/common-sdk-requirements.md
```diff
@@ -0,0 +1,43 @@
...
+* Configurable (a set of properties that can be set/modified by the system or customer):
+    * When selecting the service before generating and downloading SDK (SDK design time)
+    * When adjusting the service locally after SDK is downloaded (client code at build time or runtime)
+* Can be developed independently of other services
+* Can interact with other services using:
+    * Specific unique service identifiers (reference to a specific service)
+    * Configurable aliases (via configuration)
+* Can be used by other services (by a unique service identifier)
+* Can provide implementations for multiple programming languages/platforms
+* Not limited by nature (not necessarily to be part of any existing SDKs, it can be anything)
```
pgilmorepf: Whoa.... Cool new idea: 3pp Customers don't just upload a new plugin for an existing SDK... they can upload a whole new core sdk... OOOoooOOOooohhh... I mean they already do it, it's just a process that bottlenecks through me. Maybe we can make that totally open too.

docs/common-sdk-requirements.md
```diff
@@ -0,0 +1,43 @@
...
+* Well-defined and easy new service development process
+* Well-defined and easy new service registration process
+
+### Requirements to a service
+* Contains metadata (description of itself that can be read by the system)
+* Configurable (a set of properties that can be set/modified by the system or customer):
+    * When selecting the service before generating and downloading SDK (SDK design time)
+    * When adjusting the service locally after SDK is downloaded (client code at build time or runtime)
+* Can be developed independently of other services
+* Can interact with other services using:
```
debhaldarMS: Do you envision plugins communicating with each others independently or is the communication always brokered by the SDK Core/Plugin manager ?

docs/common-sdk-requirements.md
debhaldarMS: Should we have a requirement for versioning - how do we know which plugin is compatiable with which version of SDK ? Will the always be backwards and forward compatible ?

docs/common-sdk-requirements.md
pgilmorepf: Much of our previous discussion was "get a direct reference to other service", and then "access whatever public options are available on that service. I don't really want to bother with restrictions, because I assure you, they're not useful, and they only make the system harder for a customer to use.

Vicodex: I envision it brokered by some kind of the SDK plugin manager, but it will have a platform-specific implementation more suitable for each specific platform.

docs/common-sdk-requirements.md
pgilmorepf: Agreed. And honestly... Impossible to avoid. There's a certain amount of "that's just the only way to do it" that applies to many of these wildly different languages.

pgilmorepf: ABSOLUTELY agree. This is a SUPER important detail, and I think you're right to describe it explicitly. (this document should as well)

docs/common-sdk-requirements.md
debhaldarMS: Subclassing -- should we prevent it by sealing the plugin classes ?

pgilmorepf: Absolutely. I think Nuget is the model to use here. Hell, C# and C++ could piggyback DIRECTLY on Nuget...

pgilmorepf: That's a really good point. But in "PlayFab SDK Style" I think a VERY high assumption of reverse compatibility should be deeply encouraged. Going back to my "Piggy back on nuget" approach, I think mostly, if Nuget/Npm/Maven/etc exist for that SDK, we should just implement versioning as part of those systems.

docs/common-sdk-requirements.md
pgilmorepf: Agreed. But in general, I want a pretty open and simple system. 1 line of code for the customer, which does exactly what they expect. No magic.

docs/common-sdk-requirements.md
Vicodex: Don't we want to have some kind of governing process around that?

docs/common-sdk-requirements.md
Vicodex: Yep, it will be as simple as that :)

docs/common-sdk-requirements.md
```diff
@@ -0,0 +1,43 @@
...
+    * Services can be required
+    * Composition rules/restrictions apply based on each service dependencies and configuration
+    * Configuration errors must be prevented/communicated at the time of configuring
+* Ability to tailor SDK to specific customer needs
+* Can be composed and configured by a customer
+* Trackable SDK generation and download process to provide acceptable customer experience
+* Can be used to support all existing SDK platforms
+* Can be integrated into existing SDK generation process (eventually replace generation of monolithic SDKs)
+* Support of built-in (authored by PlayFab) and custom services (created by customers)
+* Service management/administration system (for PlayFab admins)
```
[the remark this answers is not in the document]
Vicodex: I meant some kind of a management system that would allow PlayFab to control their plugin cloud.

docs/common-sdk-requirements.md
Vicodex: I was thinking to propose Semantic Versioning concept when we get there, that is used by many modern popular systems including Nuget and other package managers. More about it here:
https://semver.org/

Vicodex: Sealing might be a language-specific concept. Some languages may not even have a thing like classes. We could leave it up to nuget authors for now.

docs/common-sdk-requirements.md
debhaldarMS: Let's capture this as a requirement and say that we'll address this later when we hot builds the sdks. But let's capture your and Paul's thoughts as an appendix(?) or any markdowny way you deem fit.

There are a few things we should avoid:
1. Making login mandatory to get an SDK - this just puts more barrier to entry for onboarders.
2. Cookie idea is difficult to implement and it's probabably going to be weeks between subsequent regenerations. Plus now we might need to make it GDPR compliant.
3. A safer bet is informing them via a dialog that either prompts them to login while explaining the benefit or has a "download anyways"
4. I love the idea of giving them a json file they can save and later upload to bypass the web UI selection process if not logged in. We'll need to explore and think about it more though.

docs/common-sdk-requirements.md
```diff
@@ -0,0 +1,43 @@
...
+* Configurable composition:
+    * Services can be selected
+    * Services can be optional
+    * Services can be required
+    * Composition rules/restrictions apply based on each service dependencies and configuration
+    * Configuration errors must be prevented/communicated at the time of configuring
+* Ability to tailor SDK to specific customer needs
+* Can be composed and configured by a customer
+* Trackable SDK generation and download process to provide acceptable customer experience
+* Can be used to support all existing SDK platforms
```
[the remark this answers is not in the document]
debhaldarMS: Agreed -- this'll become increasingly important as we onboard other SDKs like .. bumbelion(?!) :) Let's capture this as a requirement and we can prioritize it.

docs/common-sdk-requirements.md
debhaldarMS: So the eventual goal is to have a marketplace for vetting plugins that we distribute. Similar to WACK or XBox Cert or Unreal marketplace - because when we distribute, we take the legal responsibility for quality/security etc.
Till we have that, I'd like to see a system where if a third party(king.com) can create a copy of our infrastructure/ tool and add their own plugins as necessary. Alternatively, they can upload their plugin to our infrasturute and get a custom generated SDK for "them" only. We don't distribute those plugins to other parties.

docs/common-sdk-requirements.md
debhaldarMS: FWIW -- the thing to remember is "no magic" ! :)

docs/common-sdk-requirements.md
debhaldarMS: As much as i'd like to have symmetry between all sdks - the final decision needs to be governed by the use cases and facilities offered by each platform.

PR #133, opened 2018-06-06
Fixing some issues in WindowsSDK and XPlatCppSdk

pgilmorepf:
Specifically, a typo: "HTMLAllCollection" should have been "collection" in two make.js files.
I suspect that it was an auto-complete mishap that got copy pasted and wasn't found until this edge case was invoked by the API.
Also, just since I was here, most of the U("xxx") strings have been converted to L"xxx" strings, which is ultimately better for a hundred reasons, and was requested by Arun. Only 1 remains which wasn't trivial to remove.


targets/WindowsSdk/make.js
```diff
@@ -172,41 +172,41 @@ function getPropertyDefinition(tabbing, property, datatype) {
...
     var safePropName = getPropertySafeName(property);
     if (property.jsontype === "Object" && property.actualtype === "object")
-        return tabbing + "output[U(\"" + property.name + "\")] = " + safePropName + ";";
+        return tabbing + "output[L\"" + property.name + "\"] = " + safePropName + ";";
     if (property.jsontype === "Object")
-        return tabbing + "web::json::value each_" + safePropName + "; ToJsonUtilO(" + safePropName + ", each_" + safePropName + "); output[U(\"" + property.name + "\")] = each_" + safePropName + ";";
-    if (property.isenum && (property.HTMLAllCollection || property.optional))
-        return tabbing + "web::json::value each_" + safePropName + "; ToJsonUtilE(" + safePropName + ", each_" + safePropName + "); output[U(\"" + property.name + "\")] = each_" + safePropName + ";";
+        return tabbing + "web::json::value each_" + safePropName + "; ToJsonUtilO(" + safePropName + ", each_" + safePropName + "); output[L\"" + property.name + "\"] = each_" + safePropName + ";";
+    if (property.isenum && (property.collection || property.optional))
```
pgilmorepf: This is the change that actually fixes the compile issue in WindowsSDK

targets/XPlatCppSdk/make.js
```diff
@@ -197,7 +197,7 @@ function getPropertyToJson(tabbing, property, datatype) {
         return tabbing + "output[\"" + property.name + "\"] = " + safePropName + ";";
     if (property.jsontype === "Object")
         return tabbing + "Json::Value each_" + safePropName + "; ToJsonUtilO(" + safePropName + ", each_" + safePropName + "); output[\"" + property.name + "\"] = each_" + safePropName + ";";
-    if (property.isenum && (property.HTMLAllCollection || property.optional))
+    if (property.isenum && (property.collection || property.optional))
```
pgilmorepf: This is the change that actually fixes the compile issue in XPlatCppSdk

targets/WindowsSdk/source/source/PlayFabSettings.cpp.ejs
```diff
@@ -2,14 +2,14 @@
...
-    const utility::string_t PlayFabSettings::versionString = U("WindowsSDK-<%- sdkVersion %>");
+    const utility::string_t PlayFabSettings::sdkVersion = L"<%- sdkVersion %>";
+    const utility::string_t PlayFabSettings::buildIdentifier = L"<%- buildIdentifier %>";
+    const utility::string_t PlayFabSettings::versionString = L"WindowsSDK-<%- sdkVersion %>";
 
     bool PlayFabSettings::useDevelopmentEnvironment = false;
     utility::string_t PlayFabSettings::serverURL;
-    utility::string_t PlayFabSettings::developmentEnvironmentURL = U(".playfabsandbox.com");
-    utility::string_t PlayFabSettings::productionEnvironmentURL = U(".playfabapi.com");
+    utility::string_t PlayFabSettings::developmentEnvironmentURL = L".playfabsandbox.com";
```
benkatt: Should this be const?

targets/WindowsSdk/source/source/PlayFabSettings.cpp.ejs
```diff
@@ -2,14 +2,14 @@
...
+    utility::string_t PlayFabSettings::productionEnvironmentURL = L".playfabapi.com";
```
benkatt: Should this be const?

targets/WindowsSdk/source/source/PlayFabSettings.cpp.ejs
```diff
@@ -2,14 +2,14 @@
 
 namespace PlayFab
 {
-    const utility::string_t PlayFabSettings::sdkVersion = U("<%- sdkVersion %>");
-    const utility::string_t PlayFabSettings::buildIdentifier = U("<%- buildIdentifier %>");
-    const utility::string_t PlayFabSettings::versionString = U("WindowsSDK-<%- sdkVersion %>");
+    const utility::string_t PlayFabSettings::sdkVersion = L"<%- sdkVersion %>";
```
benkatt: I think utility::string_t comes from CppRestSDK. It is the same as std::wstring, right? Any reason we are not just using std::wstring?

targets/WindowsSdk/make.js
```diff
@@ -103,10 +103,10 @@ function getAuthParams(apiCall) {
     if (apiCall.url === "/Authentication/GetEntityToken")
         return "authKey, authValue";
     switch (apiCall.auth) {
-        case "None": return "U(\"\"), U(\"\")";
-        case "EntityToken": return "U(\"X-EntityToken\"), PlayFabSettings::entityToken";
-        case "SessionTicket": return "U(\"X-Authorization\"), PlayFabSettings::clientSessionTicket";
-        case "SecretKey": return "U(\"X-SecretKey\"), PlayFabSettings::developerSecretKey";
+        case "None": return "L\"\", L\"\"";
```
benkatt: Nonblocking - FWIW, I think mixing single/double quotes would make this a bit more readable. Not worth changing just the one line here by itself, but it is something to consider in the future.

```
case "None": return 'L"", L""';
```

pgilmorepf: I would agree with you... except that Resharper complains if you do that.
And I'm a bit OCD about those stupid marks all over my screen distracting me.

targets/WindowsSdk/source/source/PlayFabSettings.cpp.ejs
pgilmorepf: CppRestSdk, correct. I just use the string it wants because I don't want to constantly re-cast it to the string type it wants. I also worry about memory use/efficiency of constantly copying strings. If I use the string it wants, there's no conversion.
And for the most part, only CppRestSdk sees... MOST of the utility strings.

In the end I only really care about not abusing the memory with lots of conversions. If that's better than my worst-case-assumptions about the situation, then I don't care.

targets/WindowsSdk/source/source/PlayFabSettings.cpp.ejs
pgilmorepf: I don't think that even exists anymore. It's just an ancient hanger-on.
Truthfully we should just delete it, and make sure there's easy private cluster redirect support (For all sdks).

targets/WindowsSdk/source/source/PlayFabSettings.cpp.ejs
pgilmorepf: Probably! 😄

PR #135, opened 2018-06-12
TOC update

pgilmorepf:
Clarity, more controllable, and multi-phase roll out.


generate.ts
```diff
@@ -34,7 +34,7 @@ var sdkGeneratorGlobals: SdkGlobals = {
 };
 
 const defaultApiSpecFilePath = "../API_Specs"; // Relative path to Generate.js
-const defaultApiSpecGitHubUrl = "https://raw.githubusercontent.com/PlayFab/API_Specs/master";
+const defaultApiSpecGitHubUrl = "https://raw.githubusercontent.com/PlayFab/API_Specs/TOC";
```
Vicodex: >TOC [](start = 85, length = 3)

Just curious, why are we switching to another branch instead of using master?

generate.ts
```diff
@@ -34,7 +34,7 @@ var sdkGeneratorGlobals: SdkGlobals = {
...
 const defaultApiSpecPlayFabUrl = "https://www.playfabapi.com/apispec";
 const tocFilename = "TOC.json";
```
Vicodex: >tocFilename [](start = 6, length = 11)

May i suggest to also add a constant tocCacheKey = "TOC" here so it's defined only in one place? :)  It will make it easier to change/replace (if ever needed) and also ensure that someone won't write it incorrectly somewhere in the code as a literal

generate.ts
pgilmorepf: That was a mistake. I was testing.
I thought I reverted it.
Thanks for catching it.

generate.ts
pgilmorepf: Yeah, agreed.

pgilmorepf:
I removed the checkmark to the repo/branch settings that makes old approvals stale.

PR #139, opened 2018-06-15
Update UE Cpp plugin

benkatt:
- Delete blueprint code (PlayFabProxy) from cpp plugin. We will ultimately use the code from the blueprint plugin in some way.
- Delete PlayFabCore.h - it was empty and didn't serve a purpose
- Add copyright documentation to the top of all h and cpp files
- Move WhitelistPlatforms line in plugin to inside the module (it is supposed to be at the module level)
- Update plugin EngineVersion to 4.19.0 per UE marketplace guidelines


targets/cpp-ue4/Plugins/PlayFab/Source/PlayFab/Public/Core/PlayFabCore.h
```diff
@@ -1,13 +0,0 @@
-#pragma once
```
pgilmorepf: Wait... really? This isn't needed at all?

targets/cpp-ue4/make.js
```diff
@@ -8,23 +7,29 @@ if (typeof (templatizeTree) === "undefined") templatizeTree = function () { };
...
     var locals = {
         apis: apis,
         buildIdentifier: exports.buildIdentifier,
+        copyright: copyright,
         errorList: apis[0].errorList,
         errors: apis[0].errors,
         friendlyName: "PlayFab Cpp Sdk",
         sdkVersion: exports.sdkVersion,
-        ueTargetVersion: "4.19"
+        ueTargetVersion: "4.19.0"
```
pgilmorepf: There was an important reason why this wasn't specified.
I think it gives you a lot of warnings if you have 4.19.1, and you're loading a 4.19.0 plugin.
It's not a blocker, but it's a bad customer experience.

benkatt: I actually get warnings if it only specifies 4.19 and not 4.19.0. FWIW, the marketplace guidelines indicate the EngineVersion is supposed to be 4.x.0, not just 4.x. https://www.unrealengine.com/en-US/marketplace-guidelines

targets/cpp-ue4/Plugins/PlayFab/Source/PlayFab/Public/Core/PlayFabCore.h
benkatt: As far as I can tell, the file is empty and isn't included anywhere. The plugin compiles fine without it.

PR #142, opened 2018-06-25
added specialization support via an optional -- command line parameter in generate.ts

Vicodex:
Addition of PluginGenerator functionality in near future requires some minor refactoring of generate.ts. The idea is to keep all shared functionality between different generation processes in generate.ts and then also have "specializations" - scripts that will contain functions/overrides specified to a particular generation process. The default specialization will be "sdk" - i.e. the process we already have today.
Specialization files will be named using pattern "generate-<specialization>.ts". 
This PR introduces basic support of this. It has been tested to make sure that it doesn't break the current generate functionality.

### DIFF
--- a/SDKGenerator.njsproj
+++ b/SDKGenerator.njsproj
@@ -989,6 +989,12 @@
     <None Include="targets\javascript\testingFiles\PlayFabApiTest.ts" />
   </ItemGroup>
   <ItemGroup>
+    <TypeScriptCompile Include="generate-plugins.ts">
+      <SubType>Code</SubType>
+    </TypeScriptCompile>
+    <TypeScriptCompile Include="generate-sdk.ts">
+      <SubType>Code</SubType>
+    </TypeScriptCompile>
     <TypeScriptCompile Include="generate.ts" />
     <TypeScriptCompile Include="node.d.ts" />
     <TypeScriptCompile Include="targets\js-node\source\Scripts\typings\PlayFab\PlayFab.d.ts" />
--- a/SDKGenerator.njsproj
+++ b/SDKGenerator.njsproj
@@ -989,12 +989,8 @@
     <None Include="targets\javascript\testingFiles\PlayFabApiTest.ts" />
   </ItemGroup>
   <ItemGroup>
-    <TypeScriptCompile Include="generate-plugins.ts">
-      <SubType>Code</SubType>
-    </TypeScriptCompile>
-    <TypeScriptCompile Include="generate-sdk.ts">
-      <SubType>Code</SubType>
-    </TypeScriptCompile>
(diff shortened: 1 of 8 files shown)


SDKGenerator.njsproj
```diff
@@ -989,6 +989,12 @@
     <None Include="targets\javascript\testingFiles\PlayFabApiTest.ts" />
   </ItemGroup>
   <ItemGroup>
+    <TypeScriptCompile Include="generate-plugins.ts">
+      <SubType>Code</SubType>
+    </TypeScriptCompile>
+    <TypeScriptCompile Include="generate-sdk.ts">
+      <SubType>Code</SubType>
+    </TypeScriptCompile>
```
pgilmorepf: Visual Studio occasionally does excessive tagging of these files in the proj file for some reason. I don't know why it does this vs the format below, and how to trigger one or the other.
I usually manually convert it to the single-line format when it happens.
I'm willing to start ignoring it if we decide I'm being too OCD about it though.

generate.js
```diff
@@ -36,7 +38,7 @@ function reportErrorsAndExit(errorMessages) {
     if (errorMessages.length === 0)
         return; // No errors to report, so continue
     // Else, report all errors and exit the program
-    console.log("Synatax: node generate.js\n" +
+    console.log("Syntax: node generate.js\n" +
```
pgilmorepf: HAHA, good catch. Thanks.

generate.js
```diff
@@ -57,7 +59,21 @@ function reportErrorsAndExit(errorMessages) {
...
     extractArgs(args, argsByName, targetOutputPathList, errorMessages);
-    // Apply defaults 
+    // Apply defaults
+    var specialization = defaultSpecialization;
+    if (argsByName.specialization) {
+        specialization = argsByName.specialization;
+    }
+    var specializationFile = path.resolve("generate-" + specialization + ".ts");
+    try {
+        specializationContent = require(specializationFile);
```
pgilmorepf: You can require a .ts file?
NICE!
I would have instinctively required the .js file.

SDKGenerator.njsproj
Vicodex: I will update them manually to make one-liners as below, good point. I like minimality, too :)

---
In reply to: [198236509](https://github.com/PlayFab/SDKGenerator/pull/142#discussion_r198236509) [](ancestors = 198236509)

generate.js
Vicodex: Apparently it works, or it looks like it does :)

---
In reply to: [198237250](https://github.com/PlayFab/SDKGenerator/pull/142#discussion_r198237250) [](ancestors = 198237250)

PR #143, opened 2018-06-26
Toddbell/python

toddbellMSFT:
Adding setup.py and __init__.py files to the python generation repo


targets/PythonSdk/make.js
```diff
@@ -1,10 +1,27 @@
 var path = require("path");
+var fs = require("fs");
 
 // Making resharper less noisy - These are defined in Generate.js
 if (typeof (templatizeTree) === "undefined") templatizeTree = function () { };
 if (typeof (getCompiledTemplate) === "undefined") getCompiledTemplate = function () { };
 
+function copyOverScripts(sourceDir, apiOutputDir)
+{
+    var srcDir = sourceDir + "\\source";
```
pgilmorepf: Use path.resolve(a, b) to combine path elements

targets/PythonSdk/make.js
```diff
@@ -1,10 +1,27 @@
...
+    var outDir = apiOutputDir.substr(0, 18);
```
pgilmorepf: Uh... whut?
There's no explanation here, and this seems unusual.
What are you doing here?
This will be very bizarre if your root-repo-path is different from mine (and it probably is)

targets/PythonSdk/make.js
```diff
@@ -1,10 +1,27 @@
...
+    console.log("copying scripts from: " + srcDir + " to: " + apiOutputDir);
+
+    var filesInDir = fs.readdirSync(srcDir);
```
pgilmorepf: You don't have to do this. Use templatizeTree(), which is defined in generate.js, and automatically available global function.

targets/PythonSdk/make.js
```diff
@@ -35,15 +54,17 @@ exports.makeServerAPI = function (apis, sourceDir, apiOutputDir) {
     };
 
     console.log("Generating Server api from: " + sourceDir + " to: " + apiOutputDir);
-    templatizeTree(locals, path.resolve(sourceDir, "source"), apiOutputDir);
+    templatizeTree(locals, path.resolve(sourceDir, "source/playfab"), apiOutputDir);
```
pgilmorepf: See, copy a tree of files.  Easy. (re above)

targets/PythonSdk/make.js
```diff
@@ -35,15 +54,17 @@ exports.makeServerAPI = function (apis, sourceDir, apiOutputDir) {
...
 
     for(var i=0; i< apis.length; i++)
         makeApi(apis[i], sourceDir, apiOutputDir);
 
     generateSimpleFiles(apis, sourceDir, apiOutputDir);
+    copyOverScripts(sourceDir, apiOutputDir);
```
pgilmorepf: What are "OverScripts"  Potentially rename func.

targets/PythonSdk/make.js
```diff
@@ -35,15 +54,17 @@ exports.makeServerAPI = function (apis, sourceDir, apiOutputDir) {
...
 }
 
 exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
+    apiOutputDir += "\\playfab"
```
pgilmorepf: Use path.resolve(a, b) to combine path elements

targets/PythonSdk/source/setup.py
```diff
@@ -0,0 +1,20 @@
+import setuptools
+with open("README.md", "r") as fh:
+        long_description = fh.read()
+        
+setuptools.setup(
+    name="playfab",
+    version="0.0.1",
```
pgilmorepf: this needs to be a template for sdkVersion
Rename file, append ".ejs" to the end, and replace "0.0.1" with "<- sdkVersion >"
And add
sdkVersion = exports.sdkVersion
to the locals dict for the call to templatizeTree

targets/PythonSdk/source/setup.py
```diff
@@ -0,0 +1,20 @@
...
+    author="Todd Bello",
+    author_email="toddbell@microsoft.com",
```
pgilmorepf: This will get you a large number of direct emails from customers.
Generally we don't put an author, and add information directing them to the forums.
https://community.playfab.com/questions/ask.html

targets/PythonSdk/source/setup.py
```diff
@@ -0,0 +1,20 @@
...
+    description="PlayFab SDK for Python",
+    long_description=long_description,
```
pgilmorepf: Should probably set this.
For reference, this is what we put in Corona marketplace:
https://marketplace.coronalabs.com/corona-plugins/playfab-combo

targets/PythonSdk/source/setup.py
```diff
@@ -0,0 +1,20 @@
...
+    long_description_content_type="text/markdown",
+    url="https://github.com/PlayFab/PythonSdk",
+    packages=setuptools.find_packages(),
+    classifiers=(
+        "Programming Language :: Python :: 3",
+        "License :: OSI Approved :: Apache Software License",
+        "Operating System :: OS Independent",
```
pgilmorepf: Look in the unreal playfab.uplugin for a list of typical marketplace/package tags.

targets/PythonSdk/source/uploadPython.ps1
```diff
@@ -0,0 +1,26 @@
+Param(
```
pgilmorepf: I don't see any reason this needs to be powershell.
I'd prefer bash.
I'm trying to consolidate all the Jenkins scripts to bash.

pgilmorepf:
Let's only do "makeCombo" don't do clieint/server at all.
remove the other two and update accordingly.

targets/PythonSdk/source/uploadPython.ps1
toddbellMSFT: doesn't NEED to be powershell, but it's 1.) what I know 2.) works

just tried converting this to a .sh and ran into various problems (powershell has to launch git bash which just closes and doesn't tell you what's wrong, launching from git bash doesn't allow me to type in my password for pypi? at least that's what was happening yesterday, launching from ubuntu is saying it doesn't understand my cd commands). Uploaded an equivalent .sh file for you to look at.

targets/PythonSdk/source/uploadPython.sh
```diff
@@ -0,0 +1,24 @@
...
+cd /mnt/f/sdks/PythonSDK/
+
+python setup.py sdist bdist_wheel
+
+if [ -p "$1" ]
+then
+    twine upload dist/*
+    exit 1
+fi
+    # last test version was 0.2.8 any earlier version uploaded will probably get stomped over by pypi's cache
```
toddbellMSFT: I think this can probably be un-indented.

targets/PythonSdk/templates/setup.py.ejs
```diff
@@ -0,0 +1,22 @@
+import setuptools
+with open("README.md", "r") as fh:
+        long_description = fh.read()
+        
+setuptools.setup(
+    name="playfab",
+    version="<%- sdkVersion %>",
+    author="PlayFab Dev Tools team",
+    author_email="playfabdt@microsoft.com",
```
pgilmorepf: This is honestly worse than your own work email.
It lets the public spam us ALL directly (or more hopefully, it's a private alias that outsiders won't be able to mail at all, so it'll just bounce).

If an email is absolutely required, let's use the public zendesk email:
devrel@playfab.com
But, if the email field is not required, then it should just be deleted.

targets/PythonSdk/source/uploadPython.sh
pgilmorepf: Indeed, tabbing in bash means nothing, and thus you can use it however you want to make your script more readable.

targets/PythonSdk/source/uploadPython.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+# cd F:/sdks/PythonSDK/
```
pgilmorepf: Relative directories are important. My absolute path is different from yours.

targets/PythonSdk/source/uploadPython.sh
pgilmorepf: Bash also has functions, which syntactically are basically C-like syntax.
Use them however makes sense to you.

targets/PythonSdk/source/uploadPython.ps1
toddbellMSFT: Adding a comment to continue this work for Jekins. There are several issues with Git Bash (pypi doesn't accept username) and Ubuntu on Windows (pip isn't installed?) that are acting SUPER lame right now

PR #144, opened 2018-06-27
added reading specialization TOC

Vicodex:
* Moving the block of constants up a little so that we can use them in SdkGeneratorGlobals initialization
* Added reading from optional specialization TOC (see a separate PR in API_Specs) https://github.com/PlayFab/API_Specs/pull/55

### DIFF
--- a/generate.ts
+++ b/generate.ts
@@ -5,6 +5,11 @@ var path = require("path");
 
 ejs.delimiter = "\n";
 
+interface SpecializationTocRef {
+    name: string;
+    path: string;
+}
+
 interface SdkDoc {
     funcName: string;
     apiDocKeys: string[];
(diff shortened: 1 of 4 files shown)


generate.ts
```diff
@@ -18,8 +23,17 @@ interface SdkGlobals {
...
+    specialization: string;
 }
 
+const defaultApiSpecFilePath = "../API_Specs"; // Relative path to Generate.js
+const defaultApiSpecGitHubUrl = "https://raw.githubusercontent.com/PlayFab/API_Specs/master";
+const defaultApiSpecPlayFabUrl = "https://www.playfabapi.com/apispec";
+const tocFilename = "TOC.json";
+const tocCacheKey = "TOC";
+const specializationTocCacheKey = "specializationTOC";
+const defaultSpecialization = "sdk";
```
pgilmorepf: I was initially confused by this diff, because I thought these vars existed already. Then I scrolled and I saw you moved them. Did you move them up for readability?
I don't oppose the change, but I'd like to have some insight as to why you did it.

pgilmorepf: I guess also, how do you feel about maybe loading those default paths from the TOC for me?  😄 😄 😄  (In a future checkin)

Vicodex: Yep!

---
In reply to: [198675155](https://github.com/PlayFab/SDKGenerator/pull/144#discussion_r198675155) [](ancestors = 198675155)

Vicodex: Moved them up a little so that we can use them in SdkGeneratorGlobals initialization below

---
In reply to: [198674114](https://github.com/PlayFab/SDKGenerator/pull/144#discussion_r198674114) [](ancestors = 198674114)

PR #149, opened 2018-07-01
Proposal: Add Coroutine wrappers for API methods in Unity SDK

iomac:
Having used the Unity SDK for a while now, I have to write a lot of nested PlayFab related API calls to handle scenarios where you need to make a sequence of dependent calls (login and purchasing flows, for example). I have felt that an official set of wrappers for PlayFab calls that use Coroutines would make for cleaner and easier to read code in a number of cases -- having written a bunch of wrappers myself in my own projects to simplify problematic areas.

This PR contains the code to do just that. I implemented the methods as wrappers around the existing closure-based methods. I added a CustomYieldInstruction to SharedModels, and put the wrapper methods into their own #if/#endif block because CustomYieldInstruction was added in Unity 5.3.

The net effect is that it allows you to write more linear code, and if you are already using Coroutines for something else (quite common in Unity) it makes it easier to integrate call to PlayFab in those instances (mixing and matching gets ugly quickly).

As a quick example:

```
PlayFabClientAPI.LoginWithCustomID(
    new LoginWithCustomIDRequest {
        CustomId = "af786sa09f8ysd9fds9p8f",
        CreateAccount = true,
        TitleId = "FOO"
    }, loginResult => {
        PlayFabClientAPI.ExecuteCloudScript(
                new ExecuteCloudScriptRequest {
                FunctionName = "SyncClient",
                GeneratePlayStreamEvent = true
            }, 
            syncResult => {
                // handle result
            }, 
            error => {
                // handle error
            }
        );
    }, error => {
        // handle error
    }
);
```
becomes
```
var login = PlayFabClientAPI.LoginWithCustomID(new LoginWithCustomIDRequest {
    CustomId = "af786sa09f8ysd9fds9p8f",
    CreateAccount = true,
    TitleId = "FOO"
});

yield return login;

// handle result/error

var sync = PlayFabClientAPI.ExecuteCloudScript(new ExecuteCloudScriptRequest{ 
    FunctionName = "SyncClient",
    GeneratePlayStreamEvent = true
});

yield return sync;

// handle result/error
```

### DIFF
--- a/targets/unity-v2/source/Shared/Models/SharedModels.cs
+++ b/targets/unity-v2/source/Shared/Models/SharedModels.cs
@@ -16,4 +16,27 @@ namespace PlayFab.SharedModels
         public PlayFabRequestCommon Request;
         public object CustomData;
     }
+
+    public class WaitForPlayFabResponse<T> : UnityEngine.CustomYieldInstruction where T:PlayFabResultCommon
+    {
+        public T Result;
+        public PlayFabError Error;
+
+        bool complete = false;
+
+        public override bool keepWaiting { get { return !complete; } }
+        public bool IsError { get { return Error != null; } }
+
+        public void Complete(T r) {
+            Result = r;
+            Error = null;
+            complete = true;
+        }
+
+        public void Complete(PlayFabError e) {
+            Result = null;
+            Error = e;
+            complete = true;
+        }
+    }
 }
--- a/targets/unity-v2/source/Shared/Models/SharedModels.cs
+++ b/targets/unity-v2/source/Shared/Models/SharedModels.cs
@@ -17,6 +17,7 @@ namespace PlayFab.SharedModels
         public object CustomData;
     }
 
+#if UNITY_5_3 || UNITY_5_3_OR_NEWER
     public class WaitForPlayFabResponse<T> : UnityEngine.CustomYieldInstruction where T:PlayFabResultCommon
     {
         public T Result;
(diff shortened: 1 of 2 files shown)


pgilmorepf:
This looks like it could be very impressive.
I also want to take the time to review it properly before we decide to accept it as-is, or as a separate file/module that can be enabled.
I'm partially concerned about the 65k function limit in Android, and how many function calls this will add.

I can't guarantee a speedy review, but I'll take a look as soon as I can.

Also, it is much appreciated that you sent this in the SdkGenerator. That greatly improves the chances that this is accepted.

iomac:
Would it help/speed things up if I wrapped the new APIs in a PLAYFAB_COROUTINES define so that they are not enabled by default?

pgilmorepf:
I definitely like your suggestion, but I'd also want them separated out into separate code-files.

pgilmorepf:
The last thing I really need to do is ensure that the coroutine method can properly see all the important error information if an ApiCall fails.
I haven't actually generated this yet, apologies.
As long as a coroutine caller can see all the PlayFab error information, and errorDetails (when appropriate), then it's probably good. It's just something I want to explicitly check.

Thanks again for your submission, and apologies for slowness.

MarcoWilliamsPF:
I love the idea of this,  great addition.

joakimkm:
Any progress? With such a solution callback hell is avoided. 
Please do something.

jorisdg:
I've adopted this code, so much nicer with coroutines!

toddbellMSFT:
We are unable to take this change into the codebase at this time. Please submit a feature request through the developer forums if you would like to see this in future releases.

PR #159, opened 2018-07-05
C++11 support

Micha0:
std::make_unique is a C++14 feature. The added header includes std::make_unique if the project is set to C++11.

### DIFF
--- a/targets/XPlatCppSdk/source/cppsdk/include/StdMakeUnique.h
+++ b/targets/XPlatCppSdk/source/cppsdk/include/StdMakeUnique.h
@@ -1,0 +1,42 @@
+#pragma once
+//from https://stackoverflow.com/questions/17902405
+
+#if __cplusplus == 201103L
+
+#include <cstddef>
+#include <memory>
+#include <type_traits>
+#include <utility>
+
+namespace std {
+    template<class T> struct _Unique_if {
+        typedef unique_ptr<T> _Single_object;
+    };
+
+    template<class T> struct _Unique_if<T[]> {
+        typedef unique_ptr<T[]> _Unknown_bound;
+    };
+
+    template<class T, size_t N> struct _Unique_if<T[N]> {
+        typedef void _Known_bound;
+    };
+
+    template<class T, class... Args>
+        typename _Unique_if<T>::_Single_object
+        make_unique(Args&&... args) {
+            return unique_ptr<T>(new T(std::forward<Args>(args)...));
+        }
+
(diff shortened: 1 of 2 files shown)


pgilmorepf:
I have to ask legal about this one. Stay tuned.

pgilmorepf:
**Question:**
You should be able to create and include this in your local project, importing it before you import PlayFab, right?
**My basic point being:**
We are not going to be testing C++11, and we have plans to use smart pointers more across the package.
This might fix the problem today, but we will likely break it again in the near future, and we won't even know when we've broken it for you again.
I don't think PlayFab can/should even suggest that this package should work on C++11 now, because we intend to use more C++14 features later.
However, if you build a set of workarounds for yourself, which makes PF work for C++11, and maintain it yourself, we don't mind. I'm willing to make some minor changes to make that more possible for you, but I don't want to directly accept the C++11 workarounds into our project.

pgilmorepf:
Please respond within a few days.
My current intent is to close this request, as rejected.
Make your case, but bear in mind, we have no plans to support C++11 long term.

pgilmorepf:
I think we have accomplished your goal with 100% less legal complications, and with less ongoing maintenance concerns.
https://github.com/PlayFab/SDKGenerator/pull/163

I will still repeat myself when I say we're not planning to test C++11 in any automated way, and there's no guarantee that it'll stay unbroken. But we're willing to work with you if we break it to try and be flexible.

If you don't mind, please let us know what compiler/version you're using, and why, so that we can better understand your needs.

Thanks.

Micha0:
I'm using XCode 9.3 with Clang: `Apple LLVM version 9.1.0 (clang-902.0.39.1)`. Our project is has set the Language Dialect to `Compiler default`, with the C++11 standard library.

We use the same solution for all our projects. Which means this works for Visual Studio 2017 and Android Studio 3.x.x, NDK 15.x too. The header is guarded by a C++ version check, so even when C++14 (or higher) is enabled, it still works. (`#if __cplusplus == 201103L`)

So far, the only C++14 (or higher) this SDK is using, is one `std::make_unique` call. Which can also be replaced by `std::unique_ptr<T>(new T)` to "solve" this issue. I have no overview of your roadmap and what other C++14 features you might add in the future.

pgilmorepf:
Check out the pull request that I linked earlier:
https://github.com/PlayFab/SDKGenerator/pull/163/files
It does pretty much exactly as you say, using the same replacement you describe.

PR #160, opened 2018-07-05
ExecuteRequest is now able to handle large request

Micha0:
Curl's `CURLOPT_WRITEFUNCTION` is designed in such a way that it calls that function multiple times for large requests. Which means that the json parser parses uncomplete data. Because `reqContainer->responseString` was using `assign` instead of `append` the buffer was overwritten which results in invalid json.

`CurlReceiveData` has been refactored to only collect and buffer data while `ExecuteRequest` handles and parses the JSON result.


pgilmorepf:
Very nice, thanks!

PR #164, opened 2018-07-06
dynamically including api's in python's package init

toddbellMSFT:


SDKGenerator.njsproj
```diff
@@ -506,7 +506,7 @@
     <Content Include="targets\PythonSdk\source\playfab\PlayFabHTTP.py" />
     <Content Include="targets\PythonSdk\source\playfab\PlayFabManualTest.py" />
     <Content Include="targets\PythonSdk\source\playfab\PlayFabSettings.py" />
-    <Content Include="targets\PythonSdk\source\playfab\__init__.py" />
+    <Content Include="targets\PythonSdk\templates\__init__.py.ejs" />
```
pgilmorepf: You don't need to move this into the templates folder.
You can just rename it in-place to append ".ejs" on the end, and it automatically gets templatized where it is.

targets/PythonSdk/make.js
```diff
@@ -30,6 +30,10 @@ exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
     console.log("Generating Combined Client/Server api from: " + sourceDir + " to: " + apiOutputDir);
 
     templatizeTree(locals, path.resolve(sourceDir, "source"), apiOutputDir);
+
+    var iniTemplate = getCompiledTemplate(path.resolve(sourceDir, "templates/__init__.py.ejs"));
```
pgilmorepf: This is not necessary if you leave it in source/playfab and let templatizeTree do the work.

targets/PythonSdk/templates/__init__.py.ejs
```diff
@@ -0,0 +1,12 @@
+<% for (var a = 0; a < apis.length; a++) { var api = apis[a];
```
pgilmorepf: This file looks good except the part where it's been moved.

targets/PythonSdk/make.js
```diff
@@ -30,6 +30,7 @@ exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
     console.log("Generating Combined Client/Server api from: " + sourceDir + " to: " + apiOutputDir);
 
     templatizeTree(locals, path.resolve(sourceDir, "source"), apiOutputDir);
+
```
toddbellMSFT: I'll get rid of the extra whitespace.

PR #168, opened 2018-07-09
Toddbell/cpp hardcode fix

toddbellMSFT:


targets/XPlatCppSdk/source/LibUnitTest/PlayFabEntityTest.cpp
```diff
@@ -216,6 +216,7 @@ namespace UnittestRunner
         }
         static void SetObjectsCallback(const SetObjectsResponse& result, void*)
```
pgilmorepf: Don't do the UNREFERENCED_VARIABLE(result); thing below.
Fix result here by removing the result keyword.

targets/XPlatCppSdk/source/LibUnitTest/stdafx.h
```diff
@@ -3,6 +3,8 @@
 // Headers for CppUnitTest
 #include <CppUnitTest.h>
 
+#define UNREFERENCED_VARIABLE(X) (void) (X);
```
pgilmorepf: Don't do this

targets/XPlatCppSdk/source/LibUnitTest/PlayFabEntityTest.cpp
pgilmorepf: So literally:
static void SetObjectsCallback(const SetObjectsResponse&, void*)
That's correct syntax and it'll remove that warning/error.

targets/XPlatCppSdk/source/cppsdk/XPlatCppWindows.vcxproj.filters.ejs
```diff
@@ -17,45 +17,22 @@
     </Filter>
   </ItemGroup>
   <ItemGroup>
-    <ClInclude Include="include\playfab\PlayFabEntityDataModels.h">
-      <Filter>Header Files\playfab</Filter>
-    </ClInclude>
-    <ClInclude Include="include\playfab\PlayFabHttp.h">
```
pgilmorepf: You should not have removed this one.
Are you sure this compiled and ran?
You can't run the unit tests, but you CAN run the Windows Example Project.

toddbellMSFT: It's still there (check out line 30).

pgilmorepf: resolved. Thanks. Just missed it.

PR #169, opened 2018-07-10
Fixes for sending logic and for teardown process

alex-xbox:
- fixed sending logic, now won't send if no events;
- added teardown logic, will try to send last batch before game/app closes;
- namespace change;
- using same DateTime value for fields;
- typo fix.

### DIFF
--- a/targets/unity-v2/source/Entity/ScreenTimeTracker.cs
+++ b/targets/unity-v2/source/Entity/ScreenTimeTracker.cs
@@ -1,4 +1,4 @@
-#if ENABLE_PLAYFABENTITY_API
+﻿#if ENABLE_PLAYFABENTITY_API
 using System;
 using System.Collections.Generic;
 using PlayFab.EntityModels;
@@ -31,13 +31,14 @@ namespace PlayFab.Public
         private Guid focusId;
         private Guid gameSessionID;
         private bool initialFocus = true;
+        private bool isSending = false;
         private DateTime focusOffDateTime = DateTime.UtcNow;
         private DateTime focusOnDateTime = DateTime.UtcNow;
 
         private Queue<EventContents> eventsRequests = new Queue<EventContents>();
 
         private EntityModels.EntityKey entityInfo = new EntityModels.EntityKey();
-        private const String eventNamespace = "com.playfab.events.";
+        private const String eventNamespace = "com.playfab.events.analytics";
         private const int maxBatchSizeInEvents = 10;
 
         /// <summary>
@@ -53,9 +54,10 @@ namespace PlayFab.Public
 
             EventContents eventInfo = new EventContents();
 
-            eventInfo.Name = "cient_session_start";
+            eventInfo.Name = "client_session_start";
             eventInfo.EventNamespace = eventNamespace;
             eventInfo.Entity = entityInfo;
+            eventInfo.OriginalTimestamp = DateTime.UtcNow;
 
             var payload = new Dictionary<string, object>
                 {
@@ -78,6 +80,7 @@ namespace PlayFab.Public
         public void OnApplicationFocus(bool isFocused)
         {
             EventContents eventInfo = new EventContents();
+            DateTime currentUtcDateTime = DateTime.UtcNow;
 
             eventInfo.Name = "client_focus_change";
             eventInfo.EventNamespace = eventNamespace;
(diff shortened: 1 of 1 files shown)


targets/unity-v2/source/Entity/ScreenTimeTracker.cs
```diff
@@ -31,13 +31,14 @@ public class ScreenTimeTracker : IScreenTimeTracker
...
         private bool initialFocus = true;
+        private bool isSending = false;
         private DateTime focusOffDateTime = DateTime.UtcNow;
         private DateTime focusOnDateTime = DateTime.UtcNow;
 
         private Queue<EventContents> eventsRequests = new Queue<EventContents>();
 
         private EntityModels.EntityKey entityInfo = new EntityModels.EntityKey();
-        private const String eventNamespace = "com.playfab.events.";
+        private const String eventNamespace = "com.playfab.events.analytics";
```
pgilmorepf: Please confirm this line with Siva and/or Brent.
The left is what I was asked to use earlier today.
It seems unclear what people actually expect here.

targets/unity-v2/source/Entity/ScreenTimeTracker.cs
```diff
@@ -137,11 +146,14 @@ public void OnApplicationFocus(bool isFocused)
...
         {
-            if (PlayFabClientAPI.IsClientLoggedIn())
+            if ((PlayFabClientAPI.IsClientLoggedIn()) && (isSending == false))
             {
                 int eventsInTheBatch = 0;
+                isSending = true;
+
                 WriteEventsRequest request = new WriteEventsRequest();
                 request.Events = new List<EventContents>();
+                request.FlushToPlayStream = true;
```
pgilmorepf: https://api.playfab.com/documentation/entity/method/WriteEvents
FlushToPlayStream is not a valid request parameter for this event.
Please make sure you're actually compiling and executing this using the published assetPackage:
https://github.com/PlayFab/UnitySDK/blob/master/Packages/UnitySDK.unitypackage
FlushToPlayStream causes a compile error in the project.

targets/unity-v2/source/Entity/ScreenTimeTracker.cs
```diff
@@ -173,7 +190,7 @@ private void EventSentErrorCallback(PlayFabError response)
             Debug.LogWarning("Failed to send session data. Error: " + response.GenerateErrorReport());
         }
 
-#region Unused MonoBehaviour compatibility  methods
+        #region Unused MonoBehaviour compatibility  methods
```
pgilmorepf: Please use the Visual Studio "Ctrl+k+d" standard spacing.

targets/unity-v2/source/Entity/ScreenTimeTracker.cs
```diff
@@ -151,7 +163,12 @@ public void Send()
                     eventsInTheBatch++;
```
pgilmorepf: I think eventsInTheBatch is redundant.
It's just a mirror of:
request.Events.Count()

targets/unity-v2/source/Entity/ScreenTimeTracker.cs
alex-xbox: Used VS standard spacing, no difference. I'd assume "standard" is not so standard and we need to sync our settings.

targets/unity-v2/source/Entity/ScreenTimeTracker.cs
alex-xbox: fixed

targets/unity-v2/source/Entity/ScreenTimeTracker.cs
alex-xbox: fixed

targets/unity-v2/source/Entity/ScreenTimeTracker.cs
alex-xbox: fixed

targets/unity-v2/source/Entity/ScreenTimeTracker.cs
```diff
@@ -78,25 +80,31 @@ public void ClientSessionStart(EntityModels.EntityKey entityKey, string playFabU
...
             EventContents eventInfo = new EventContents();
+            DateTime currentUtcDateTime = DateTime.UtcNow;
 
             eventInfo.Name = "client_focus_change";
             eventInfo.EventNamespace = eventNamespace;
             eventInfo.Entity = entityInfo;
 
             double focusStateDuration = 0.0;
 
+            if (initialFocus)
```
kasharoo: When do you set initialFocus to false? Also, are there any threading concerns that would make you want to change initialFocus to static?

targets/unity-v2/source/Entity/ScreenTimeTracker.cs
```diff
@@ -137,21 +146,25 @@ public void OnApplicationFocus(bool isFocused)
         /// </summary>
         public void Send()
         {
-            if (PlayFabClientAPI.IsClientLoggedIn())
+            if ((PlayFabClientAPI.IsClientLoggedIn()) && (isSending == false))
             {
-                int eventsInTheBatch = 0;
+                isSending = true;
```
kasharoo: If multi-threading is an issue, consider making isSending static.

pgilmorepf: Unity is inherently single threaded. There's no real benefit of going out of our way to take this out of the Unity main-thread, so it's unlikely that this will ever have to deal with multi-thread issues.

PR #172, opened 2018-07-11
Fix: re-throwing an exception after logging so that it's not eaten

Vicodex:


Vicodex:
Paul reviewed this offline

PR #175, opened 2018-07-12
Added PluginManager tests for:

Vicodex:
- Custom Serializer plugin
- Custom Transport plugin
- Multiple transport plugins used at the same time

(custom plugins are declared in another C# project - probably not the right one - as they can't be declared in the same project/assembly with SDK itself)


targets/csharp/templates/PlayFabSDK.csproj.ejs
```diff
@@ -54,6 +54,7 @@
     <Compile Include="source\PluginContract.cs" />
     <Compile Include="source\PluginManager.cs" />
     <Compile Include="source\Uunit\PlayFabApiTest.cs" />
+    <Compile Include="source\Uunit\PluginManagerTest.cs" />
```
Vicodex: >PluginManagerTest [](start = 35, length = 17)

I know we talked about adding this to UnitTestRunner project instead, but there are complications with that. I can show offline, it will be faster.

targets/csharp/UnittestRunner/UnittestRunner.csproj
```diff
@@ -42,7 +42,7 @@
     <DebugType>pdbonly</DebugType>
     <Optimize>true</Optimize>
     <OutputPath>bin\Release\</OutputPath>
-    <DefineConstants>TRACE;NETFX_CORE;SIMPLE_JSON_TYPEINFOENABLE_PLAYFABADMIN_API;ENABLE_PLAYFABSERVER_API;ENABLE_PLAYFABMATCHMAKER_API;ENABLE_PLAYFABENTITY_API</DefineConstants>
```
toddbellMSFT: These look like Unity pre-compiler defines. Are these csharp tests expected to run in that environment?

targets/csharp/source/source/Uunit/PluginManagerTest.cs
```diff
@@ -0,0 +1,353 @@
...
+
+        // Fixed values provided from testInputs
+        private static string USER_EMAIL;
+        private static Dictionary<string, string> extraHeaders;
+
+        /// <summary>
+        /// PlayFab Title cannot be created from SDK tests, so you must provide your titleId to run unit tests.
+        /// (Also, we don't want lots of excess unused titles)
+        /// </summary>
+        public static void SetTitleInfo(TestTitleData testInputs)
```
pgilmorepf: You shouldn't need this.

targets/csharp/source/source/Uunit/PluginManagerTest.cs
```diff
@@ -0,0 +1,353 @@
...
+            extraHeaders = testInputs.extraHeaders;
+
+            // Verify all the inputs won't cause crashes in the tests
+            TITLE_INFO_SET &= !string.IsNullOrEmpty(PlayFabSettings.TitleId)
+                && !string.IsNullOrEmpty(USER_EMAIL);
+        }
+
+        public override void SetUp(UUnitTestContext testContext)
+        {
+            if (!TITLE_INFO_SET)
```
pgilmorepf: You shouldn't need this function-body, and once the body is empty you may not need to implement SetUp().

targets/csharp/source/source/Uunit/PluginManagerTest.cs
```diff
@@ -0,0 +1,353 @@
...
+                testContext.Skip(); // We cannot do client tests if the titleId is not given
+        }
+
+        public override void Tick(UUnitTestContext testContext)
```
pgilmorepf: Empty override functions aren't actually necessary.
You can keep them if you think you will use them, but if you know you won't need them, you should probably just delete it.

targets/csharp/source/source/Uunit/PluginManagerTest.cs
```diff
@@ -0,0 +1,353 @@
...
+        {
+            // No work needed, async tests will end themselves
+        }
+
+        public override void TearDown(UUnitTestContext testContext)
```
pgilmorepf: ditto.

targets/csharp/source/source/Uunit/PluginManagerTest.cs
```diff
@@ -0,0 +1,353 @@
...
+            var playFabSerializer = PluginManager.GetPlugin(PluginContract.PlayFab_Serializer) as ISerializerPlugin;
+            var customSerializer = new CustomSerializerPlugin();
+            try
+            {
+                // Set a custom serializer plugin
+                PluginManager.SetPlugin(customSerializer, PluginContract.PlayFab_Serializer);
+
+                // Call some PlayFab API 
+                var getRequest = new GetUserDataRequest();
+                var getDataTask = PlayFabClientAPI.GetUserDataAsync(getRequest, null, extraHeaders);
```
pgilmorepf: I still feel like it's unnecessary to make PlayFab API calls as part of testing PluginManager.
It puts too much on the test, and too many disconnected things can cause the test to fail.

targets/csharp/source/source/Uunit/PluginManagerTest.cs
```diff
@@ -0,0 +1,353 @@
...
+        }
+
+        private class CustomApi
+        {
+            public string PluginName;
+
+            public async Task<object> SomeApiMethod()
+            {
+                var transport = (ITransportPlugin)PluginManager.GetPlugin(PluginContract.PlayFab_Transport, this.PluginName);
+                return await transport.DoPost(null, null, null);
```
pgilmorepf: For this test, I wouldn't add the extra complication of using the "real" transport plugin, nor the async.
Both of those add complexity, but don't really improve the "test N plugin clones" case.
I would create two totally custom plugin types: CustomApi and CustomTransport, and not reuse ANY existing plugins, expectations, or interfaces.

For example, if we change the signature of DoPost, we have to change this test, when in reality, this test has no required connection to the DoPost function signature.

targets/csharp/source/source/Uunit/PluginManagerTest.cs
```diff
@@ -0,0 +1,353 @@
...
+                    Task<object> task = api.SomeApiMethod();
+                    resultTasks.Add(task);
+                    totalNumberOfCalls++;
+                });
+
+                var resultTasksArray = resultTasks.ToArray();
+                Task.WaitAll(resultTasksArray);
+
+                // Verify
+                testContext.True(totalNumberOfCalls == pluginNumber);
```
pgilmorepf: This line doesn't actually test anything about your plugins.
You set totalNumberOfCalls totally independently of anything your plugins do, and thus the test condition doesn't actually contribute to what you're trying to test.

targets/csharp/source/source/Uunit/PluginManagerTest.cs
```diff
@@ -0,0 +1,353 @@
...
+                for (int i = 0; i < pluginNumber; i++)
+                {
+                    var pluginName = "Plugin_" + i;
+                    var transportPlugin = PluginManager.GetPlugin(PluginContract.PlayFab_Transport, pluginName);
```
pgilmorepf: Right here is a great example. I feel like GetPlugin should be a generic function, so that you don't have to cast it yourself in the next line.
PluginManager should do all the otherwise copy-paste work, so that the caller doesn't have to repeatedly do anything like recasting.

targets/csharp/templates/PlayFabSDK.csproj.ejs
pgilmorepf: No worries, the tests do what they're meant to do, and they work. So good enough!  :D

targets/csharp/source/source/PluginManager.cs
```diff
@@ -25,9 +25,9 @@ private PluginManager()
         /// <param name="contract">The plugin contract.</param>
         /// <param name="instanceName">The optional plugin instance name. Instance names allow to have mulptiple plugins with the same contract.</param>
         /// <returns>The plugin instance.</returns>
-        public static IPlayFabPlugin GetPlugin(PluginContract contract, string instanceName = "")
+        public static T GetPlugin<T>(PluginContract contract, string instanceName = "") where T : IPlayFabPlugin
         {
-            return Instance.GetPluginInternal(contract, instanceName);
+            return (T)Instance.GetPluginInternal(contract, instanceName);
```
pgilmorepf: Two options here, and I'm not sure either one is "right":
return (T)GetPluginInternal(); // Which will throw an exception if the cast fails.
return GetPluginInternal() as T; // Which will silently return null if the cast fails.
I could see an argument for either one, and I haven't thought through the customer experience to really have an opinion. But perhaps you should consider which option you think is best before landing.

targets/csharp/source/source/Uunit/PluginManagerTest.cs
```diff
@@ -0,0 +1,125 @@
...
+            public string Name;
+
+            public async Task<object> DoPost(string urlPath, object request, Dictionary<string, string> headers)
+            {
+                throw new NotImplementedException();
+            }
+        }
+
+        /// <summary>
+        /// CLIENT AND SERVER API
```
pgilmorepf: CLIENT AND SERVER API doesn't apply to your tests.
the whole line is irrelevant since you're not testing APIs

targets/csharp/source/source/Uunit/PluginManagerTest.cs
```diff
@@ -0,0 +1,125 @@
...
+            try
+            {
+                // Set a custom serializer plugin
+                PluginManager.SetPlugin(customSerializer, PluginContract.PlayFab_Serializer);
+
+                // Get serializer plugin from manager
+                var serializerPlugin = PluginManager.GetPlugin<ISerializerPlugin>(PluginContract.PlayFab_Serializer);
+
+                // Verify
+                testContext.True(object.ReferenceEquals(serializerPlugin, customSerializer));
```
pgilmorepf: I like to use the phrasing "expected" and "actual" as much as possible during tests.
I need to read the above lines pretty carefully to decide which is which in this case.

targets/csharp/source/source/Uunit/PluginManagerTest.cs
```diff
@@ -0,0 +1,125 @@
...
+                testContext.EndTest(UUnitFinishState.PASSED, null);
+            }
+            catch (Exception e)
```
pgilmorepf: What manner of exception is possible here?
Is an exception part of the normal customer experience?

targets/csharp/source/source/Uunit/PluginManagerTest.cs
pgilmorepf: with a bunch of "ditto's" all through this test file.

targets/csharp/source/source/Uunit/PluginManagerTest.cs
pgilmorepf: I like tests to match the customer's experience as much as possible, because we should be testing that the customer's experience matches what we expect.

targets/csharp/source/source/PluginManager.cs
Vicodex: I thought about it too, and decided that explicit cast failure exception here is better than null ref exception somewhere else

---
In reply to: [202452655](https://github.com/PlayFab/SDKGenerator/pull/175#discussion_r202452655) [](ancestors = 202452655)

targets/csharp/source/source/Uunit/PluginManagerTest.cs
Vicodex: The finally block below is needed in case if anything goes wrong (including any exceptions that may happen inside PluginManager - in the future, if not now - we can't guarantee it will never happen). But you are right, the catch block itself seems redundant since the test will fail on exception anyway

---
In reply to: [202453615](https://github.com/PlayFab/SDKGenerator/pull/175#discussion_r202453615) [](ancestors = 202453615)

PR #177, opened 2018-07-13
Changes to support engine build for the ue bp sdk

sanchitgarg:


targets/cpp-unreal/PluginSource/Source/PlayFab/Private/PlayFabBaseModel.cpp
```diff
@@ -2,8 +2,8 @@
 // This files holds the code for the play fab base model.
 //////////////////////////////////////////////////////////////////////////////////////////////
 
-#include "PlayFabPrivatePCH.h"
 #include "PlayFabBaseModel.h"
+#include "PlayFabPrivatePCH.h"
```
benkatt: It's been a while since I looked at this. I can't remember - do you get build errors if the PCH is first? Usually PCH is listed before other headers.

sanchitgarg: Yes, Unreal needs the <CppFileName>.h to be the first header to be included. Im not sure if we are actually using PCH since we do not define that in the build.cs file.

PR #179, opened 2018-07-13
Separate the data-collection step and the data-processing step

pgilmorepf:
…of OnPlayFabLogin, to avoid future issues with LoginResult/RegisterPlayFabUserResult differences.

### DIFF
--- a/targets/unity-v2/source/Shared/Internal/PlayFabHttp/PlayFabUnityHttp.cs
+++ b/targets/unity-v2/source/Shared/Internal/PlayFabHttp/PlayFabUnityHttp.cs
@@ -106,7 +106,12 @@ namespace PlayFab.Internal
             };
 
             foreach (var headerPair in reqContainer.RequestHeaders)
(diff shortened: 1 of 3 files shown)


targets/unity-v2/source/Shared/Internal/PlayFabHttp/PlayFabUnityHttp.cs
```diff
@@ -155,7 +160,7 @@ private IEnumerator Post(CallRequestContainer reqContainer)
                                 var jsonResponse = streamReader.ReadToEnd();
                                 //Debug.Log(jsonResponse);
                                 OnResponse(jsonResponse, reqContainer);
-                                Debug.Log("Successful UnityHttp decompress for: " + www.url);
+                                //Debug.Log("Successful UnityHttp decompress for: " + www.url);
```
alex-xbox: Not sure we need commented code here. Feel free to keep it if necessary

PR #187, opened 2018-07-26
Change Blueprint module name from PlayFab to PlayFabBp

sanchitgarg:


targets/cpp-unreal/PluginSource/Source/PlayFabBp/Private/PlayFabPrivate.h
```diff
@@ -18,4 +18,4 @@ DECLARE_LOG_CATEGORY_EXTERN(LogPlayFab, Log, All);
 
 #include "IPlayFab.h"
 
-#include "PlayFabClasses.h"
+// #include "PlayFabClasses.h"
```
sanchitgarg: Cannot find this header file. Was this removed?
Will delete it instead of the comment then.

pgilmorepf: I... want to say that Ben removed it because it was unused... but now we find it's used?
I think you'll just have to go look at the commit history and try to figure it out.
Otherwise, I don't know.

pgilmorepf: (Delete seems fine?)

sanchitgarg: No, dont think we need it as everything builds without the header.
Ill remove the line then.

PR #191, opened 2018-07-27
Test for PUT

alex-xbox:
Added test for PUT verb

### DIFF
--- a/targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs
+++ b/targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs
@@ -16,10 +16,8 @@ namespace PlayFab.UUnit
         private const string TEST_OBJ_NAME = "testCounter";
         // Test variables
         private EntityKey _entityKey;
-        private string _testFileData;
         private const string TEST_FILE_NAME = "testfile";
         private const string _testPayload = "{123456789}";
-        private UUnitTestContext _testContext;
         private int _testInteger;
 
         public override void SetUp(UUnitTestContext testContext)
@@ -214,27 +212,22 @@ namespace PlayFab.UUnit
 
             if (!testFileFound)
             {
-                UploadFile(TEST_FILE_NAME, testContext);
+                UploadFile(testContext, TEST_FILE_NAME);
             }
             else
             {
-                GetActualFile(fileMetaData, testContext);
+                GetActualFile(testContext, fileMetaData);
             }
         }
-        void GetActualFile(GetFileMetadata fileData, UUnitTestContext testContext)
+        void GetActualFile(UUnitTestContext testContext, GetFileMetadata fileData)
         {
-            PlayFabHttp.SimpleGetCall(fileData.DownloadUrl,
-                result =>
-                {
-                    _testFileData = Encoding.UTF8.GetString(result);
-                    TestFileContent(testContext);
-                },
+            PlayFabHttp.SimpleGetCall(fileData.DownloadUrl, PlayFabUUnitUtils.SimpleApiActionWrapper<byte[]>(testContext, TestFileContent),
                 error =>
                 {
                     testContext.EndTest(UUnitFinishState.FAILED, error);
                 });
         }
-        void UploadFile(string fileName, UUnitTestContext testContext)
+        void UploadFile(UUnitTestContext testContext, string fileName)
         {
             var request = new InitiateFileUploadsRequest
             {
@@ -245,10 +238,9 @@ namespace PlayFab.UUnit
                 },
             };
 
-            _testContext = testContext;
-            PlayFabEntityAPI.InitiateFileUploads(request, OnInitFileUpload, OnInitFailed);
+            PlayFabEntityAPI.InitiateFileUploads(request, PlayFabUUnitUtils.ApiActionWrapper<InitiateFileUploadsResponse>(testContext, OnInitFileUpload), PlayFabUUnitUtils.ApiActionWrapper<PlayFabError>(testContext, OnInitFailed), testContext);
         }
-        void DeleteFiles(List<string> fileName, UUnitTestContext testContext)
+        void DeleteFiles(UUnitTestContext testContext, List<string> fileName)
         {
             var request = new DeleteFilesRequest
             {
@@ -263,6 +255,8 @@ namespace PlayFab.UUnit
         }
         void OnInitFailed(PlayFabError error)
         {
+            var testContext = (UUnitTestContext)error.CustomData;
+
             if (error.Error == PlayFabErrorCode.EntityFileOperationPending)
             {
                 var request = new AbortFileUploadsRequest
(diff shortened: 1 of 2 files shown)


targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs
```diff
@@ -141,6 +145,194 @@ private void GetObjectCallback2(GetObjectsResponse result)
...
+        void DeleteFiles(List<string> fileName, UUnitTestContext testContext)
+        {
+            var request = new DeleteFilesRequest
+            {
+                Entity = _entityKey,
+                FileNames = fileName,
+            };
+
+            PlayFabEntityAPI.DeleteFiles(request, result =>
+            {
```
pgilmorepf: Deletefiles success callback has no contents so if it gets here, the test will time out and fail.

targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs
```diff
@@ -14,8 +16,10 @@ public class EntityApiTests : UUnitTestCase
         private const string TEST_OBJ_NAME = "testCounter";
         // Test variables
         private EntityKey _entityKey;
-        private string _testFileUrl;
-        private string _testFileChecksum;
+        private string _testFileData;
+        private const string TEST_FILE_NAME = "testfile";
+        private const string _testPayload = "{123456789}";
+        private UUnitTestContext _testContext;
```
pgilmorepf: This is really unsafe.
This is a great way for the testContext of one test to leak into another.
The testContext instance is passed into the constructor of each test specifically to avoid any potential leakage like this.
If you only have one test in your test class, it would be... LESS dangerous, but this isn't the case even here.

Avoiding that leakage is exactly why I created wrappers like ApiActionWrapper.

targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs
```diff
@@ -141,6 +145,194 @@ private void GetObjectCallback2(GetObjectsResponse result)
...
+            }
+            else
+            {
+                GetActualFile(fileMetaData, testContext);
+            }
+        }
+        void GetActualFile(GetFileMetadata fileData, UUnitTestContext testContext)
+        {
+            PlayFabHttp.SimpleGetCall(fileData.DownloadUrl,
+                result =>
```
pgilmorepf: These simple callbacks are almost certainly the right way for customers to write this.
But unfortunately, if they fail in some way, they don't communicate the test failure back to the testContext.
For example, add:
```throw new Exception("do not commit this line");```
inside of one of these little lambda callbacks.
When you do, the test won't report that error, it will just time out mysteriously.

That's why the ApiActionWrapper exists for PfApi callbacks.
I'd suggest making another ApiActionWrapper in the same place, that wraps calls meant for SimpleGet/SimplePut.
I know it seems overly complex for testing the success case, but we also need to make sure this test captures failure cases and reports them in a useful way.

Comment (pgilmorepf) [reply targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs]: (I'm not gonna repeat this comment in all the relevant places, but essentially EVERY lambda callback needs to be wrapped such that if it throws an exception, that exception gets properly displayed as a test failure, and not a test timeout)

File: targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs [hunk 4]
```diff
@@ -141,6 +145,194 @@ private void GetObjectCallback2(GetObjectsResponse result)
...
+                {
+                    _testFileData = Encoding.UTF8.GetString(result);
```
pgilmorepf: TestFileContent should just receive _testFileData as a parameter, and it should not be saved to a static variable like this. Use a local var, and pass it into TestFileContent.
This is a less severe kind of potential test-leakage that I was talking about with _testContext.

targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs
```diff
@@ -141,6 +145,194 @@ private void GetObjectCallback2(GetObjectsResponse result)
...
+                    Id = result.EntityToken.Entity.Id,
+                    Type = entityType,
+                    TypeString = result.EntityToken.Entity.TypeString
+                };
+
+                LoadFiles(testContext);
+            }
+            else
+            {
+                testContext.EndTest(UUnitFinishState.FAILED, "Entity Token is null!");
```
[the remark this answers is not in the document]
alex-xbox: I certainly can, but I was following current code patterns. In EntityTest it's .EndTest approach everywhere. I can change it, but consistency will be lost. Same applies for other comments about it. Please let me know

targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs
alex-xbox: Callback(lambda) here is empty, but code is correct. This will not fail, it will just continue without any additional actions.

targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs
alex-xbox: I'm not sure how exactly leak can happen, I removed it anyway.

targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs
alex-xbox: Created 2 more wrappers

targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs
alex-xbox: Modified

targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs
pgilmorepf: Perhaps I'm confused about the code flow (callbacks can get messy no matter what).
How does this get to DeleteFiles, and what happens next if it's called.

targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs
```diff
@@ -214,27 +212,22 @@ void OnGetFilesInfo(GetFilesResponse result)
...
-        void GetActualFile(GetFileMetadata fileData, UUnitTestContext testContext)
+        void GetActualFile(UUnitTestContext testContext, GetFileMetadata fileData)
         {
-            PlayFabHttp.SimpleGetCall(fileData.DownloadUrl,
-                result =>
-                {
-                    _testFileData = Encoding.UTF8.GetString(result);
-                    TestFileContent(testContext);
-                },
+            PlayFabHttp.SimpleGetCall(fileData.DownloadUrl, PlayFabUUnitUtils.SimpleApiActionWrapper<byte[]>(testContext, TestFileContent),
```
pgilmorepf: This is great.

PR #192, opened 2018-07-27
Getparams in UnitySDK and C# SDK

Waznop:


targets/csharp/templates/PlayFabSettings.cs.ejs
```diff
@@ -34,12 +40,36 @@ namespace PlayFab
...
+            StringBuilder sb = new StringBuilder(1000);
+        
             var baseUrl = UseDevelopmentEnvironment ? DevelopmentEnvironmentUrl : ProductionEnvironmentUrl;
-            if (baseUrl.StartsWith("http"))
-                return baseUrl;
-            return "https://" + TitleId + baseUrl + apiCall;
+            if (!baseUrl.StartsWith("http"))
+                sb.Append("https://").Append(TitleId);
+            sb.Append(baseUrl);
+            if (!baseUrl.StartsWith("http"))
```
pgilmorepf: I think this second if statement is incorrect.
You always need the apiCall added to the string.
Remove line 51, and fix the tabbing for (current) line 52.

targets/unity-v2/templates/PlayFabSettings.cs.ejs
```diff
@@ -148,15 +156,35 @@ namespace PlayFab
...
             var baseUrl = ProductionEnvironmentUrlPrivate;
-            if (baseUrl.StartsWith("http"))
-                output = baseUrl + apiCall;
-            else
-                output = "https://" + TitleId + baseUrl + apiCall;
-            return output;
+            if (!baseUrl.StartsWith("http"))
+                sb.Append("https://").Append(TitleId);
+        
+            sb.Append(baseUrl).Append(apiCall);
```
pgilmorepf: Comparison with above:
You always add apiCall here, and this one is correct.

PR #194, opened 2018-07-30
As3 ApiGroup TestFix

pgilmorepf:


targets/actionscript/make.js
```diff
@@ -13,7 +15,7 @@ exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
     removeExcessFiles(apis, apiOutputDir);
 
     var locals = {
-        airVersion: 28, // Latest version of AirSdk installed overtop of Flex SDK
+        airVersion: airVersion,
```
Vicodex: >airVersion [](start = 8, length = 10)

Do we even need this in locals since we moved it to a global variable?

targets/actionscript/make.js
```diff
@@ -5,6 +5,8 @@ var path = require("path");
 if (typeof (getCompiledTemplate) === "undefined") getCompiledTemplate = function () { };
 if (typeof (templatizeTree) === "undefined") templatizeTree = function () { };
 
+var airVersion = 30; // Latest version of AirSdk installed overtop of Flex SDK
```
Vicodex: >30 [](start = 17, length = 2)

Just curious why we moved it to a global variable

pgilmorepf: My machine is new and I hadn't yet installed Flex & Air. It's fairly hard to install the older versions of air sdk, so I always just install the latest and update this version number.
I couldn't find it as fast as I wanted, so after searching for like 20 seconds, I found it, but also decided it needed to be closer to the top and more "findable", for the next time.

PR #202, opened 2018-07-31
getparams in python

Waznop:


targets/PythonSdk/source/playfab/PlayFabSettings.py
```diff
@@ -57,14 +57,29 @@ class InternalSettings:
...
+        for idx, (k, v) in enumerate(getParams.items()):
+            if idx == 0:
+                url.append("?")
+            else:
+                url.append("&")
+            url.append(k)
+            url.append("=")
+            url.append(v)
+
+    return "".join(url)
```
pgilmorepf: That's pretty brilliant.

PR #203, opened 2018-08-01
ApiGroup test fixes, and fixing lots of warnings from Plugins.

pgilmorepf:


targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs
```diff
@@ -319,10 +315,10 @@ void TestFileContent(UUnitTestContext testContext, byte[] result)
         {
             var testFileData = Encoding.UTF8.GetString(result);
 
-            testContext.True(testFileData.Equals(_testPayload));
+            testContext.True(testFileData.Equals(_testPayload), testFileData);
```
pgilmorepf: Alex, can you help me figure out why this test is failing here?
testFileData is null.

pgilmorepf: Only happens after my changes, but I don't know why.

targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs
```diff
@@ -116,23 +126,23 @@ private void GetObjectCallback1(GetObjectsResponse result)
...
+                Entity = new DataModels.EntityKey { Id = _entityId, TypeString = _entityTypeString },
+                Objects = new List<DataModels.SetObject> {
+                    new DataModels.SetObject{ ObjectName = TEST_OBJ_NAME, DataObject = _testInteger }
                 }
             };
-            PlayFabEntityAPI.SetObjects(updateRequest, PlayFabUUnitUtils.ApiActionWrapper<SetObjectsResponse>(testContext, UpdateObjectCallback), PlayFabUUnitUtils.ApiActionWrapper<PlayFabError>(testContext, SharedErrorCallback), testContext);
+            PlayFabDataAPI.SetObjects(updateRequest, PlayFabUUnitUtils.ApiActionWrapper<DataModels.SetObjectsResponse>(testContext, UpdateObjectCallback), PlayFabUUnitUtils.ApiActionWrapper<PlayFabError>(testContext, SharedErrorCallback), testContext);
         }
-        private void UpdateObjectCallback(SetObjectsResponse result)
+        private void UpdateObjectCallback(DataModels.SetObjectsResponse result)
```
toddbellMSFT: a bit of a nitpick here: may be worth having using namespace DataModels? Makes the function declaration a bit cleaner.

targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs
```diff
@@ -219,38 +224,44 @@ void OnGetFilesInfo(GetFilesResponse result)
...
             };
 
-            PlayFabEntityAPI.InitiateFileUploads(request, PlayFabUUnitUtils.ApiActionWrapper<InitiateFileUploadsResponse>(testContext, OnInitFileUpload), PlayFabUUnitUtils.ApiActionWrapper<PlayFabError>(testContext, OnInitFailed), testContext);
+            PlayFabDataAPI.InitiateFileUploads(request, PlayFabUUnitUtils.ApiActionWrapper<DataModels.InitiateFileUploadsResponse>(testContext, OnInitFileUpload), PlayFabUUnitUtils.ApiActionWrapper<PlayFabError>(testContext, OnInitFailed), testContext);
         }
-        void DeleteFiles(UUnitTestContext testContext, List<string> fileName)
+        void DeleteFiles(UUnitTestContext testContext, List<string> fileName, bool shouldEndTest, UUnitFinishState finishState, string finishMessage)
         {
-            var request = new DeleteFilesRequest
+            if (!_shouldDeleteFiles) // Only delete the file if it was created
```
alex-xbox: Will it gracefully end test after return in all cases?

targets/unity-v2/make.js
```diff
@@ -1,10 +1,10 @@
 var path = require("path");
 
 // Making resharper less noisy - These are defined in Generate.js
-if (typeof (copyTree) === "undefined") copyTree = function () { };
```
toddbellMSFT: Sweet! Finally got the chance to update it 👍

targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs
pgilmorepf: I'm at a bit of a crossroads on that thinking.
Certainly a customer can do it if/when they want to.
For us:
There are name overlaps in MANY of the model namespaces at this point, and that's only going to get worse.
Those places are mandatory for defining the model namespace, and so, I'm leaning towards consistency, where internally we define all the namespaces.

Also...

There's a task where we remove all using's in our own plugin to avoid namespace collision problems with customers who define classes of the same names in the global space by accident, and then see compiler errors in our code, despite it not being our problem. So, this might just get worse for us all around.

targets/unity-v2/Testing/Tests/Entity/EntityApiTests.cs
pgilmorepf: Nope!
This is certainly something we should address in the TestSuite itself, which is that SetUp and TearDown should be able to be long-running async as well.
For now, the "standard case" is that the test won't have any execution outside of itself.
If it failed, then may (likely) have callbacks executing after the text execution, but it's already a fail case, so later test-corruption isn't... any worse.
Certainly some tech debt for later, but for now, the standard/passing case is that it runs cleanly and correctly without any chance of cross-test-corruption in the passing case.

targets/unity-v2/make.js
pgilmorepf: Yeah, I couldn't find the file in-place, and I realized it was still in templates, so I moved it.

PR #205, opened 2018-08-01
Revert "Change Blueprint module name from PlayFab to PlayFabBp (#187)"

sanchitgarg:
This reverts commit 405e14ce4bf6839f68957b7444da4b8da96e3e01.


targets/newTarget/make.js
```diff
@@ -4,7 +4,7 @@ var path = require("path");
 if (typeof (copyTree) === "undefined") copyTree = function () { };
 if (typeof (getCompiledTemplate) === "undefined") getCompiledTemplate = function () { };
 
-// generate.js looks for some specific exported functions (as defined in TOC.json) in make.js, like:
```
pgilmorepf: Please pull the changes to newTarget into a separate checkin and restore them back to master.
As an addendum, this is a great example of why these should have been in a separate checkin from Unreal.

PR #210, opened 2018-08-03
Renaming ApiGroups in Python

pgilmorepf:
And making a "compile test" that can be run by Jenkins.

### DIFF
--- a/targets/PythonSdk/source/playfab/testCompile.py.ejs
+++ b/targets/PythonSdk/source/playfab/testCompile.py.ejs
@@ -1,0 +1,16 @@
+import py_compile
+
+files = [
+<% for (var a = 0; a < apis.length; a++) { var api = apis[a];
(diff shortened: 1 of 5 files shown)


targets/PythonSdk/source/playfab/testCompile.py.ejs
```diff
@@ -0,0 +1,16 @@
+import py_compile
+
+files = [
+<% for (var a = 0; a < apis.length; a++) { var api = apis[a];
+%>"PlayFab<%- api.name %>API.py",
```
Vicodex: >api [](start = 14, length = 3)

Could just use apis[a] here without creating api each cycle?

pgilmorepf: Yep, probably. I mostly write it this way because the for() { var ...; is written that way in every other template.

PR #211, opened 2018-08-03
A customer commented on how this line is very confusing.

pgilmorepf:


targets/unity-v2/source/Shared/Internal/PlayFabHttp/PlayFabHTTP.cs
```diff
@@ -288,7 +288,7 @@ private void OnEnable()
             }
 
 #if ENABLE_PLAYFABENTITY_API && !DISABLE_PLAYFABCLIENT_API
-            if ((screenTimeTracker != null) && (!PlayFabSettings.DisableFocusTimeCollection != false))
+            if (screenTimeTracker != null && !PlayFabSettings.DisableFocusTimeCollection)
```
alex-xbox: I agree, it is looked confusing. I'd recommend keeping () intact.

PR #213, opened 2018-08-04
Fixed Windows SDK tests

Vicodex:


targets/WindowsSdk/make.js
```diff
@@ -11,7 +11,7 @@ exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
     var locals = {
         apis: apis,
         buildIdentifier: exports.buildIdentifier,
-        extraDefines: "ENABLE_PLAYFABADMIN_API;ENABLE_PLAYFABENTITY_API;ENABLE_PLAYFABMATCHMAKER_API;ENABLE_PLAYFABSERVER_API;",
+        extraDefines: "ENABLE_PLAYFABADMIN_API;ENABLE_PLAYFABENTITY_API;ENABLE_PLAYFABMATCHMAKER_API;ENABLE_PLAYFABSERVER_API;ENABLE_PLAYFABAUTHENTICATION_API;ENABLE_PLAYFABDATA_API;",
```
pgilmorepf: You have found an important issue, but you have fixed it in a way that does not match plans that I didn't tell you about.
:P sorry.
All of the new ApiGroups should be clustered under the Entity api define, despite the fact that the names changed.
See this:
https://github.com/PlayFab/SDKGenerator/blob/master/targets/unity-v2/make.js#L416
Do something similar to that. Don't create lots of new ApiGroups.

PR #218, opened 2018-08-07
Get params in Javascript SDK

Waznop:


targets/javascript/templates/PlayFab_Api.js.ejs
```diff
@@ -57,6 +60,21 @@ if(!PlayFab._internalSettings) {
             var startTime = new Date();
             var requestBody = JSON.stringify(request);
 
+            var completeUrl = url;
+            var getParams = PlayFab._internalSettings.requestGetParams;
+            if (getParams != null) {
+                var firstParam = true;
+                for (var key in getParams) {
+                    if (firstParam) {
+                        completeUrl += "?";
```
toddbellMSFT: Can we add the question mark to completeUrl before the for loop and remove the firstParam bool/if check?

Waznop: I don't think that'd cover the case where getParams is empty though

pgilmorepf: I don't strongly care about either of these solutions, but:
Hong's current code seems to work fine.
Todd's suggestion could be achieved by changing:
 if (getParams != null) {
to
 if (getParams != null && getParams.length > 0) {
     completeUrl += "?";
     ...
(and then removing firstParam)

I don't care which way, they both work, and they're both clear, so it's up to you guys to decide whichever you think is more readable to customers.

targets/javascript/templates/PlayFab_Api.js.ejs
```diff
@@ -57,6 +60,21 @@ if(!PlayFab._internalSettings) {
...
+                        firstParam = false;
+                    } else {
+                        completeUrl += "&";
+                    }
+                    completeUrl += key + "=" + getParams[key];
```
pgilmorepf: Are you sure that key is the string here, and not an integer index?

Waznop: yes

targets/javascript/templates/PlayFab_Api.js.ejs
Waznop: I think Todd's solution would also either involve checking for the last key (completeUrl += key + "=" + getParams[key] + "&"; except for the last ley) or handling the first key outside of the for loop.

PR #231, opened 2018-08-09
Moving a script from hard-coded in Jenkins into SdkGenerator repo

pgilmorepf:


JenkinsConsoleUtility/JenkinsScripts/publishSquashAndMerge.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# Squash and Merge from automated->master
+#   If versioned exists: master->versioned, and then tag and notate the release in GitHub
+
+cd "$WORKSPACE/sdks/$SdkName"
+
+git fetch --progress origin
+
+echo === Sync $AUTOMATED_GIT_BRANCH ===
+git checkout $AUTOMATED_GIT_BRANCH || git checkout -b $AUTOMATED_GIT_BRANCH
```
toddbellMSFT: nit: since this is used at least twice (here at line 11 and below at line 14 and 23), it may be cleaner making this a more general method that takes a branch name.

pgilmorepf: Agree. Let's tackle that sometime.

PR #232, opened 2018-08-09
SDK info in Java SDK get params

Waznop:


targets/java/source_shared/src/main/java/com/playfab/PlayFabSettings.java.ejs
```diff
@@ -24,7 +36,21 @@ public class PlayFabSettings {
     public static final String AD_TYPE_IDFA = "Idfa";
     public static final String AD_TYPE_ANDROID_ID = "Adid";
 <% } %>
-    public static String GetURL() {
-        return "https://" + TitleId + ".playfabapi.com";
+    public static String GetURL(String apiCall) {
+        StringBuilder sb = new StringBuilder();
```
pgilmorepf: Java has StringBuilders?!?  That's awesome!
Good find.

targets/java/source_shared/src/main/java/com/playfab/PlayFabSettings.java.ejs
```diff
@@ -24,7 +36,21 @@ public class PlayFabSettings {
...
+        sb.append("https://").append(TitleId).append(".playfabapi.com").append(apiCall);
```
pgilmorepf: We need to make a change here.
".playfabapi.com" hard coded here means that the Java SDK is not capable of being redirected to a vertical.
I think we should check this in now, since it's not a regression, but we should make a task to make the Java endpoint adjustable in the Unity fashion.

pgilmorepf: I've made a task on the board, so we can accept this review now, and move on.

PR #235, opened 2018-08-10
SDK info in get params in XPlatCppSdk

Waznop:


targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabSettings.cpp.ejs
```diff
@@ -39,10 +43,25 @@ namespace PlayFab
...
     }
 
-    std::string PlayFabSettings::GetUrl(const std::string& urlPath)
+    std::string PlayFabSettings::GetUrl(const std::string& urlPath, const std::map<std::string, std::string>& getParams)
     {
         if (serverURL.length() == 0)
             serverURL = "https://" + titleId + (useDevelopmentEnvironment ? developmentEnvironmentURL : productionEnvironmentURL);
-        return serverURL + urlPath;
+        
+        std::string fullUrl = serverURL + urlPath;
```
pgilmorepf: Discussed in person about using std::string::resize() here, to be more inline with other StringBuilder mechanics in other SDKs.

PR #239, opened 2018-08-16
Update the Cloud Script typescript syntax

pgilmorepf:
... to account for the new ApiGroup restructuring.


targets/SdkTestingCloudScript/source/Scripts/typings/PlayFab/CloudScript.d.ts.ejs
```diff
@@ -67,21 +67,43 @@ interface IApiError {
...
+<% for(var sa in serverApiList) { var api = serverApiList[sa];
+for(var i in api.calls) { var apiCall = api.calls[i];
+%><%- generateApiSummary("    ", apiCall, "summary", "https://api.playfab.com/Documentation/" + api.name + "/method/" + apiCall.name) -%>
+    <%- apiCall.name %>(request: PlayFab<%- api.name %>Models.<%- apiCall.request %>): PlayFab<%- api.name %>Models.<%- apiCall.result %>;
+
+<% } } %>
+}
+
+<% for(var ea in entityApiList) { var api = entityApiList[ea]; %>
+/** <%- api.name %>API.Models as interfaces */
```
toddbellMSFT: Should this commented line still be in?

pgilmorepf: I think so? The new output is intended to match the old output as closely as possible in this case.
So this is just a pure refactor to get this job working and building valid output again.

PR #250, opened 2018-08-28
Toddbell/node get param

toddbellMSFT:
adding getParam to Node sdk.

updated javascript sdk getParam solution to construct a string with an array instead of multiple string copies.

### DIFF
--- a/targets/javascript/templates/PlayFab_Api.js.ejs
+++ b/targets/javascript/templates/PlayFab_Api.js.ejs
@@ -60,24 +60,20 @@ if(!PlayFab._internalSettings) {
             var startTime = new Date();
             var requestBody = JSON.stringify(request);
 
-            var urlArr = ["", url]; //make a new array for the URL
+            var completeUrl = url; 
             var getParams = PlayFab._internalSettings.requestGetParams;
             if (getParams != null) {
                 var firstParam = true;
                 for (var key in getParams) {
                     if (firstParam) {
-                        urlArr.push("?");
+                        completeUrl += "?"; 
                         firstParam = false;
                     } else {
-                        urlArr.push("&");
+                        completeUrl += "&"; 
                     }
-                    urlArr.push(key);
-                    urlArr.push("=");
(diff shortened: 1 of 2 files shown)


targets/javascript/templates/PlayFab_Api.js.ejs
```diff
@@ -60,21 +60,25 @@ if(!PlayFab._internalSettings) {
             var startTime = new Date();
             var requestBody = JSON.stringify(request);
 
-            var completeUrl = url;
+            var urlArr = ["", url]; //make a new array for the URL
```
pgilmorepf: You don't need to init with an empty string here.
It can just be [url] as a list of one.

targets/js-node/source/Scripts/PlayFab/playfab.js.ejs
```diff
@@ -43,6 +46,25 @@ exports.MakeRequest = function (urlStr, request, authType, authValue, callback)
     else
         requestBody = JSON.stringify(request);
 
+    var urlArr = ["", url]; //make a new array for the URL
```
pgilmorepf: You don't need to init with an empty string here.
It can just be [url] as a list of one.

pgilmorepf: url is also undefined here, as you probably meant to use urlStr here.

targets/js-node/source/Scripts/PlayFab/playfab.js.ejs
```diff
@@ -43,6 +46,25 @@ exports.MakeRequest = function (urlStr, request, authType, authValue, callback)
...
+            }
+            urlArr.push(key);
+            urlArr.push("=");
+            urlArr.push(getParams[key]);
+        }
+    }
+
+    completeUrl = urlArr.join("");
+
     var options = url.parse(urlStr);
```
pgilmorepf: This is using the initial urlStr, which should be completeUrl.

PR #261, opened 2018-08-31
Add netstandard1.1 and netstandard2.0 frameworks

kyschouv:
Convert to the new [csproj project Sdk format](https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-use-project-sdk?view=vs-2017).
Add `netstandard1.1` and `netstandard2.0` target frameworks, while maintaining the `.NETPortable` target framework.
Move `.nuspec` info into the csproj file (as described in [Additions to the csproj format for .NET Core](https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#nuget-metadata-properties)).

The new generated projects will require Visual Studio 2017 with the **.NET Core** workload installed to compile, though the generated nuget package should be usable by older projects.


kyschouv:
The nuspec properties in the project accomplish the same thing as having a nuspec file, but make it significantly easier to maintain library references. The build tools will generate a nuspec at compile time (that you can view in the `obj` folder), though you can instead use `dotnet pack` to generate the `nupkg`. We use it for all of our internal nuget packages, and I tested it very thoroughly with this project to make sure it produced a nupkg with the same metadata (I believe the only missing property is `summary`, which isn't available as a project-file property).

If you'd like help updating your build I'm happy to assist.

PR #265, opened 2018-09-05
Removing ancient deprecated fields.

pgilmorepf:
Updating internal usage to use the proper non-deprecated fields.


targets/unity-v2/source/Server/PlayFabSettings.cs
```diff
@@ -12,13 +12,13 @@ public static partial class PlayFabSettings
 #if ENABLE_PLAYFABPLAYSTREAM_API && ENABLE_PLAYFABSERVER_API
         public static string ProductionEnvironmentPlayStreamUrl
         {
-            set { PlayFabShared.ProductionEnvironmentPlayStreamUrl = value; }
+            set { PlayFabSharedPrivate.ProductionEnvironmentPlayStreamUrl = value; }
```
Vicodex: >PlayFabSharedPrivate [](start = 18, length = 20)

Are those actually in a different class now?

PR #273, opened 2018-09-10
Added support for verticals in C# SDK

Vicodex:
- Removed DevelopmentURL archaism
- Added VerticalName configuration setting to SDK (null if no vertical)
- Updated dynamic URL to support verticals
- Verified tests

### DIFF
--- a/targets/csharp/templates/PlayFabSettings.cs.ejs
+++ b/targets/csharp/templates/PlayFabSettings.cs.ejs
@@ -15,9 +15,7 @@ namespace PlayFab
         /// <summary> This is only for customers running a private cluster.  Generally you shouldn't touch this </summary>
         public static string ProductionEnvironmentUrl = ".playfabapi.com";
         /// <summary> Tha name of a customer vertical. This is only for customers running a private cluster.  Generally you shouldn't touch this </summary>
-        public static string VerticalName = <% if (verticalName) { %>"<%- verticalName %>"<%
-        } else { %>null<%
-        } %>;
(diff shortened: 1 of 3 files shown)


targets/csharp/templates/PlayFabSettings.cs.ejs
```diff
@@ -12,12 +12,12 @@ namespace PlayFab
...
-        /// <summary> This is for PlayFab internal debugging.  Generally you shouldn't touch this </summary>
-        public static bool UseDevelopmentEnvironment = false;
-        /// <summary> This is for PlayFab internal debugging.  Generally you shouldn't touch this </summary>
-        public static string DevelopmentEnvironmentUrl = ".playfabsandbox.com";
         /// <summary> This is only for customers running a private cluster.  Generally you shouldn't touch this </summary>
         public static string ProductionEnvironmentUrl = ".playfabapi.com";
+        /// <summary> Tha name of a customer vertical. This is only for customers running a private cluster.  Generally you shouldn't touch this </summary>
+        public static string VerticalName = <% if (verticalName) { %>"<%- verticalName %>"<%
+        } else { %>null<%
+        } %>;
```
pgilmorepf: I typically put this kind of logic in the make.js file as a function.
So what I would put here is:
public static string VerticalName = <%- getVerticalDefault() %>;
That makes the template look nicer, and the configuration is more controllable from make.js

Vicodex: Sounds like an awesome idea, i'll do that. Thanks!

PR #274, opened 2018-09-11
Fixing a name conflict with FPlayFabBaseModel

pgilmorepf:
It exists in both the Cpp Module and the Bp Module, and it was getting cross-connected inappropriately.
Also adding a random header that MAY fix something else that we couldn't repro.
Clearing out builds for the old UE projects.


targets/UnrealMarketplacePlugin/examplesource/ExampleProject/Source/ExampleProject.Target.cs
```diff
@@ -9,6 +9,7 @@ public class ExampleProjectTarget : TargetRules
 {
     public ExampleProjectTarget(TargetInfo Target) : base(Target)
     {
+        bUseUnityBuild = false;
```
debhaldarMS: This should be for testing only ? We should remove that from the example project before submission. Also, did you test it with the Unity build turned on ?

pgilmorepf: It's possible in a future checkin to set that flag for the Example Project, and not for the plugin.
No, I only tested it as written here.

PR #278, opened 2018-09-12
Added support for verticals in Java SDK

Vicodex:


targets/java/source_shared/src/main/java/com/playfab/PlayFabSettings.java.ejs
```diff
@@ -20,6 +20,8 @@ public class PlayFabSettings {
         RequestGetParams = Collections.unmodifiableMap(getParams);
     }
 
+    public static String ProductionEnvironmentUrl = ".playfabapi.com"; // This is only for customers running a private cluster.  Generally you shouldn't touch this
+    public static String VerticalName = <%- getVerticalNameDefault() %>; // Tha name of a customer vertical. This is only for customers running a private cluster.  Generally you shouldn't touch this
```
toddbellMSFT: nitpick: can the comments be above the variable (to prevent word wrap)? Also, I think there's a spelling error (Tha should be The)

PR #281, opened 2018-09-12
Added support for verticals in PHP SDK

Vicodex:


targets/PhpSdk/make.js
```diff
@@ -27,13 +28,22 @@ function makeApi(api, sourceDir, apiOutputDir) {
...
     var apiTemplate = getCompiledTemplate(path.resolve(sourceDir, "templates/PlayFabApi.php.ejs"));
     writeFile(path.resolve(apiOutputDir, "PlayFab" + api.name + "Api.php"), apiTemplate(locals));
 }
 
+function getVerticalNameDefault() {
+    if (exports.verticalName) {
+        return "\"" + exports.verticalName + "\"";
+    }
+
+    return "null";
```
pgilmorepf: null isn't correct in PHP.
I think it's all caps:
http://php.net/manual/en/language.types.null.php

targets/PhpSdk/source/PlayFabHttp.php.ejs
```diff
@@ -5,8 +5,9 @@ if (!class_exists("PlayFabSettings")) {
     {
         public static $versionString = "PhpSdk-<%- sdkVersion %>";
         public static $requestGetParams = array("sdk" => "<%- sdkVersion %>");
-        public static $PlayFabApiUrl = "https://{titleId}.playfabapi.com";
+        public static $productionEnvironmentUrl = ".playfabapi.com"; // This is only for customers running a private cluster.  Generally you shouldn't touch this
         public static $enableCompression = false; // Can't get this to work
+        public static $verticalName = <%- getVerticalNameDefault() %>; // The name of a customer vertical. This is only for customers running a private cluster. Generally you shouldn't touch this
         public static $titleId = null;
```
pgilmorepf: Which... apparently means I'm a dumb-something something, because I made the same mistake.

pgilmorepf:
Apparently "null" doesn't seem to break because I ran this at least once. Yet, the official identifier is supposed to be "NULL".
I'll leave it to you if it's tested and working as lowercase, then I don't care.

PR #284, opened 2018-09-13
http plugin interface

toddbellMSFT:
The base plugin interface is now implemented and works. We don't have a test for the plugin manager just yet, but I wanted to keep these reviews smaller.

THERE IS STILL AN ISSUE WITH SOME C# FILES SHOWING UP. I'll need some guidance to un-wind this problem, but I wanted the cpp files to be visible immediately for your consideration.


targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp
```diff
@@ -106,15 +117,49 @@ namespace PlayFab
...
         reqContainer->errorWrapper.UrlPath = urlPath;
         reqContainer->authKey = authKey;
         reqContainer->authValue = authValue;
         reqContainer->errorWrapper.Request = requestBody;
         reqContainer->internalCallback = internalCallback;
-        reqContainer->successCallback = successCallback;
+
+        if (sCurrentCallback != nullptr)
+        {
+            reqContainer->successCallback = sCurrentCallback;
```
pgilmorepf: If we need to track additional callbacks as part of the InternalAddRequest signature, that's fine. We should NOT be providing this parameter via a static parameter. That's just going to cause concurrency or threading issues.
The nice thing about InternalAddRequest is that we can change the signature however we need, so just add it as a parameter. Don't do this roundabout static field as a property thing.

targets/XPlatCppSdk/templates/PlayFab_Api.cpp.ejs
```diff
@@ -32,7 +41,16 @@ namespace PlayFab
 <%- getRequestActions("        ", apiCall) %>
         IPlayFabHttp& http = IPlayFabHttp::Get();
         const auto requestJson = request.ToJson();
-        http.AddRequest("<%- apiCall.url %>", <%- getAuthParams(apiCall) %>, requestJson, On<%- apiCall.name %>Result, SharedVoidPointer((callback == nullptr) ? nullptr : new ProcessApiCallback<<%- apiCall.result %>>(callback)), errorCallback, customData);
+
+        Json::FastWriter writer;
+        std::string jsonAsString = writer.write(requestJson);
+
+        std::unordered_map<std::string, std::string> headers;
+        headers.emplace(<%- getAuthParams(apiCall) %>);
```
pgilmorepf: This is really great. We set the auth at the very beginning, and never have to think about them again.
Every bit of auth processing AFTER this point is redundant.

pgilmorepf: (What happens if auth is null/null, such as login calls?)

targets/XPlatCppSdk/templates/PlayFab_Api.cpp.ejs
```diff
@@ -32,7 +41,16 @@ namespace PlayFab
...
+
+        http.NextCallback(SharedVoidPointer((callback == nullptr) ? nullptr : new ProcessApiCallback<<%- apiCall.result %>>(callback)));
```
pgilmorepf: Ok, now I see where this is used, and this is definitely not the right way to do this additional callback.
Let's look at why these callbacks are needed and structured this way, and figure out the right solution.

targets/XPlatCppSdk/source/cppsdk/include/playfab/PlayFabHttp.h
```diff
@@ -19,17 +20,24 @@ namespace PlayFab
...
     {
     public:
         static IPlayFabHttp& Get();
 
         virtual ~IPlayFabHttp();
 
-        virtual void AddRequest(const std::string& urlPath, const std::string& authKey, const std::string& authValue, const Json::Value& requestBody, RequestCompleteCallback internalCallback, SharedVoidPointer successCallback, ErrorCallback errorCallback, void* customData) = 0;
         virtual size_t Update() = 0;
+
+        virtual void AddPostRequest(const std::string&, std::unordered_map<std::string, std::string>, const std::string&, RequestCompleteCallback, ErrorCallback = nullptr, void* = nullptr) override { };
```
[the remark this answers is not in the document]
toddbellMSFT: That will end up with an "unused parameter" issue unless you want void(url); 

I'll go for adding them back in as commented out names to at least help readability.

targets/XPlatCppSdk/templates/PlayFab_Api.cpp.ejs
toddbellMSFT: Normal thing will happen: auth is passed as two empty strings and acts in the same manner as it did before.

toddbellMSFT:
Toddsa mentioned we may want to change the name from AddPostRequest to something like MakeRequest or MakePostRequest. 

The presence of Add implies we will be using a queue of some kind which may not always be the case for all other clients.

targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp
```diff
@@ -106,15 +115,29 @@ namespace PlayFab
         return (blockSize * blockCount);
     }
 
-    void PlayFabHttp::AddRequest(const std::string& urlPath, const std::string& authKey, const std::string& authValue, const Json::Value& requestBody, RequestCompleteCallback internalCallback, SharedVoidPointer successCallback, ErrorCallback errorCallback, void* customData)
+    void PlayFabHttp::AddPostRequest(const std::string& url, std::unordered_map<std::string, std::string> headers, const std::string& requestBody, std::function<void(CallRequestContainerBase&)> internalCallback, ErrorCallback userErrorCallback, void* extraBuffer)
+    {
+        Json::Value root;
+        Json::Reader reader;
```
Vicodex: Will these work with both _WIN32 and !_WIN32 defines?

targets/XPlatCppSdk/source/cppsdk/include/playfab/PlayFabHttp.h
```diff
@@ -41,12 +47,16 @@ namespace PlayFab
...
-        void AddRequest(const std::string& urlPath, const std::string& authKey, const std::string& authValue, const Json::Value& requestBody, RequestCompleteCallback internalCallback, SharedVoidPointer successCallback, ErrorCallback errorCallback, void* customData) override;
+        virtual void AddPostRequest(const std::string& url, std::unordered_map<std::string, std::string> headers, const std::string& requestBody, std::function<void(CallRequestContainerBase&)> callback, ErrorCallback errorCallback = nullptr, void* extrabuffer = nullptr) override;
+
         size_t Update() override;
+
     private:
         PlayFabHttp(); // Private constructor, to enforce singleton instance
         PlayFabHttp(const PlayFabHttp& other); // Private copy-constructor, to enforce singleton instance
 
+        virtual void InternalAddRequest(const std::string& urlPath, std::unordered_map<std::string, std::string> headers, /*const std::string& authKey, const std::string& authValue,*/ const Json::Value& requestBody, std::function<void(CallRequestContainerBase&)> internalCallback, ErrorCallback errorCallback, void* customData) override;
```
Vicodex: >unordered_map [](start = 73, length = 13)

Are we passing a copy of maps with every call or a reference or a pointer would be enough?

targets/XPlatCppSdk/templates/PlayFab_Api.cpp.ejs
```diff
@@ -32,22 +41,29 @@ namespace PlayFab
...
-        }
+        Json::FastWriter writer;
+        std::string jsonAsString = writer.write(requestJson);
+
+        std::unordered_map<std::string, std::string> headers;
+        headers.emplace(<%- getAuthParams(apiCall) %>);
+
+        http.AddPostRequest(
+            "<%- apiCall.url %>", 
+            headers,
```
Vicodex: >headers [](start = 12, length = 7)

Be careful here passing in headers here that you created on the stack while in the API call. If you access those headers after that stack variable leaves the scope (and destroyed) there can be a problem

targets/XPlatCppSdk/templates/PlayFab_Api.cpp.ejs
```diff
@@ -32,22 +41,29 @@ namespace PlayFab
 <%- getRequestActions("        ", apiCall) %>
```
debhaldarMS: Why is this generated on the same line and without braces ?

targets/XPlatCppSdk/source/cppsdk/include/playfab/PlayFabPluginManager.h
```diff
@@ -30,7 +30,8 @@ namespace PlayFab
     /// </summary>
     class IPlayFabHttpPlugin : public IPlayFabPlugin
     {
-        virtual void AddRequest(std::string url, std::unordered_map<std::string, std::string> headers, SharedVoidPointer callback, ErrorCallback errorCallback = nullptr, void* extrabuffer = nullptr);
+    public:
+        virtual void MakePostRequest(const std::string& url, const std::unordered_map<std::string, std::string>& headers, const std::string& requestBody, std::function<void(CallRequestContainerBase&)> callback, ErrorCallback errorCallback = nullptr, void* extrabuffer = nullptr) = 0;
```
debhaldarMS: Please break it down into multiple lines

debhaldarMS: >std::string& url [](start = 43, length = 16)

Can you please consult this and check if we still need to pass strings by const reference ?
[https://stackoverflow.com/questions/10231349/are-the-days-of-passing-const-stdstring-as-a-parameter-over](https://stackoverflow.com/questions/10231349/are-the-days-of-passing-const-stdstring-as-a-parameter-over)

debhaldarMS: >std::string, std::string [](start = 86, length = 24)

Do you expect anyone to change this stuff ?

targets/XPlatCppSdk/source/cppsdk/include/playfab/PlayFabHttp.h
```diff
@@ -41,12 +47,16 @@ namespace PlayFab
...
         ~PlayFabHttp() override;
 
-        void AddRequest(const std::string& urlPath, const std::string& authKey, const std::string& authValue, const Json::Value& requestBody, RequestCompleteCallback internalCallback, SharedVoidPointer successCallback, ErrorCallback errorCallback, void* customData) override;
+        virtual void MakePostRequest(const std::string& url, const std::unordered_map<std::string, std::string>& headers, const std::string& requestBody, std::function<void(CallRequestContainerBase&)> callback, ErrorCallback errorCallback = nullptr, void* extrabuffer = nullptr) override;
+
         size_t Update() override;
+
     private:
         PlayFabHttp(); // Private constructor, to enforce singleton instance
         PlayFabHttp(const PlayFabHttp& other); // Private copy-constructor, to enforce singleton instance
```
debhaldarMS: Explicitly delete the move constructor and move assignment operators

targets/XPlatCppSdk/source/cppsdk/include/playfab/PlayFabHttp.h
```diff
@@ -41,12 +47,16 @@ namespace PlayFab
...
 
+        virtual void InternalAddRequest(const std::string& urlPath, const std::unordered_map<std::string, std::string>& headers, const Json::Value& requestBody, std::function<void(CallRequestContainerBase&)> internalCallback, ErrorCallback errorCallback, void* customData) override;
```
debhaldarMS: >InternalAddRequest [](start = 21, length = 18)

Consider "AddRequestInternal" ? your call

debhaldarMS:
>         bool threadRunning;

This is not threadsafe and you might run into an inconsistent state while pumping your worker thread - use std::atomic<bool> threadRunning

---
Refers to: targets/XPlatCppSdk/source/cppsdk/include/playfab/PlayFabHttp.h:68 in 8d55fd7. [](commit_id = 8d55fd7066b4258b55f81342a69bf65069e482f4, deletion_comment = False)

debhaldarMS:
>         int activeRequestCount;

Change this to a std::atomic and you can get rid of a bunch of mutex lock/unlocks

---
Refers to: targets/XPlatCppSdk/source/cppsdk/include/playfab/PlayFabHttp.h:69 in 8d55fd7. [](commit_id = 8d55fd7066b4258b55f81342a69bf65069e482f4, deletion_comment = False)

debhaldarMS:
>             PlayFabHttp::MakeInstance();

Braces

---
Refers to: targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp:26 in 4cc061e. [](commit_id = 4cc061e7214cf2c5cd6e680d3b01004ec4637b3c, deletion_comment = False)

debhaldarMS:
>             delete pendingResults[i];

Braces, line breaks between individual unit of work, indentations

---
Refers to: targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp:45 in 4cc061e. [](commit_id = 4cc061e7214cf2c5cd6e680d3b01004ec4637b3c, deletion_comment = False)

targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp
```diff
@@ -106,15 +115,41 @@ namespace PlayFab
...
+        }
+    }
+
+    void PlayFabHttp::InternalAddRequest(
+        std::string urlPath,
+        const std::unordered_map<std::string, std::string>& headers,
+        const Json::Value& requestBody,
+        std::function<void(CallRequestContainerBase&)> internalCallback,
+        ErrorCallback errorCallback,
+        void* customData)
```
debhaldarMS: Can we get rid of these two ?

debhaldarMS:
>             std::unique_lock<std::mutex> lock(httpRequestMutex);

TODO(victor): The threads will block here - we need signalling by means of CV to not spin CPU cores unnecessarily . Implement something like:

cv.Wait()
Scoped Acquire Mutex
...
Cv.Signal()


---
Refers to: targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp:157 in 4cc061e. [](commit_id = 4cc061e7214cf2c5cd6e680d3b01004ec4637b3c, deletion_comment = False)

debhaldarMS:
>                 std::this_thread::sleep_for(std::chrono::milliseconds(10));

TODO(Victor): So we'll be blocking the core whenever we're context switched in and hog the core for 10ms at a time doing no work whatsoever. PLease consider yielding the thread and allow the other little guys a fair chance :) [https://en.cppreference.com/w/cpp/thread/yield](https://en.cppreference.com/w/cpp/thread/yield)

---
Refers to: targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp:85 in 4cc061e. [](commit_id = 4cc061e7214cf2c5cd6e680d3b01004ec4637b3c, deletion_comment = False)

debhaldarMS:
>         reqContainer.curlHandle = curl_easy_init();

TODO(Victor): Why are we not caching this ? Also, we might be leaking handles every request - please double check.

---
Refers to: targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp:166 in 4cc061e. [](commit_id = 4cc061e7214cf2c5cd6e680d3b01004ec4637b3c, deletion_comment = False)

debhaldarMS:
>             reqContainer.errorWrapper.HttpCode = 408;

This is a really confusing error - a 400 level response means we got a response from the server. We should change it to something that communicates that the correct response code.

Ironically, playfab returns 200 level response even for failed requests. We should be congnizant of the issues this causes.

---
Refers to: targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp:207 in 4cc061e. [](commit_id = 4cc061e7214cf2c5cd6e680d3b01004ec4637b3c, deletion_comment = False)

debhaldarMS:
>             HandleResults(reqContainer);

braces

---
Refers to: targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp:98 in 4cc061e. [](commit_id = 4cc061e7214cf2c5cd6e680d3b01004ec4637b3c, deletion_comment = False)

debhaldarMS:
>         PlayFabHttp& instance = static_cast<PlayFabHttp&>(Get());

TODO(Victor): This check is redundant in most cases - should be moved inside the if statement

---
Refers to: targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp:100 in 4cc061e. [](commit_id = 4cc061e7214cf2c5cd6e680d3b01004ec4637b3c, deletion_comment = False)

Vicodex:
Another PR was created and merged into master that took all work from here.

PR #285, opened 2018-09-13
Added support for verticals in JavaScript and NodeJS SDKs

Vicodex:


targets/js-node/source/Scripts/PlayFab/playfab.js.ejs
```diff
@@ -33,8 +32,16 @@ var _internalSettings = exports._internalSettings = {
 };
 
 exports.GetServerUrl = function () {
-    var baseUrl = exports.settings.useDevEnv ? exports._internalSettings.devEnvUrl : exports._internalSettings.liveUrl;
-    return "https://" + exports.settings.titleId + baseUrl;
+    var baseUrl = exports._internalSettings.liveUrl;
```
pgilmorepf: I see that liveUrl already existed in _internalSettings, but that's inconsistent with the other SDKs in both name and intent.
Let's move this to public settings, and rename it to productionUrl.

Vicodex: Ok, will do

PR #287, opened 2018-09-14
Added support for verticals to Python SDK

Vicodex:


targets/PythonSdk/source/playfab/PlayFabSettings.py.ejs
```diff
@@ -2,7 +2,13 @@ import playfab.PlayFabErrors as PlayFabErrors
 import sys
 import traceback
 
-ProductionEnvironmentURL = "https://{titleId}.playfabapi.com{methodUrl}"
+ProductionEnvironmentURL = ".playfabapi.com"
+
+"""
+The name of a customer vertical. This is only for customers running a private cluster. Generally you shouldn't touch this
+"""
```
toddbellMSFT: docstrings should be below the item in question (weird python convention). https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

Vicodex: Following our offline discussion I am going to leave it consistent with the way how other comments are written already in that file. If it turns out to be wrong, we will need to change all of them across the board, anyway.

PR #288, opened 2018-09-15
Added support for verticals and URL params in ActionScript SDK

Vicodex:
Finally managed to set up a working ActionScript dev environment (what a pain in the b..!)
Changes are verified, all tests pass.

### DIFF
--- a/targets/actionscript/source/com/playfab/PlayFabSettings.as.ejs
+++ b/targets/actionscript/source/com/playfab/PlayFabSettings.as.ejs
@@ -1,8 +1,15 @@
 package com.playfab
 {
+    import flash.utils.Dictionary;
+
     public class PlayFabSettings
     {
         public static var ProductionEnvironmentURL:String = ".playfabapi.com";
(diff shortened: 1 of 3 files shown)


targets/actionscript/source/com/playfab/PlayFabSettings.as.ejs
```diff
@@ -18,12 +25,48 @@ package com.playfab
...
+                }
+                else
+                {
+                    fullUrl += "&";
+                }
+
+                fullUrl = fullUrl + key + "=" + value;
+            }
+
+            return fullUrl;
```
pgilmorepf: I think we should make a tech-debt task for a future sprint to optimize this string construction.
Some web results seem to indicate the best would be an array of strings, with a join at the end, much like Node/JS.
But this should be committed now, and published today.

Vicodex: Makes sense, I created a task for that

PR #290, opened 2018-09-18
Unreal Mkpl submission improvements

pgilmorepf:
FIX: Windows specific OS call put behind a PLATFORM_WINDOWS define.
Improvement: For UE Mkpl submission, we generate the plugin uniquely targeting each supported version of Unreal

Templatize the source dir in one pass, not a bunch of random intermediate chunks
 - Removes some redundant code and unneeded things

Add the Jenkins build scripts to the templates, so each version can be quickly tested from the console


targets/UnrealMarketplacePlugin/make.js
```diff
@@ -14,35 +13,54 @@ var copyright =
...
+        cppMakeJsPath.makeCppCombinedAPI(apis, copyright, sourceDir, apiOutputDir, ueTargetVersion, exports.sdkVersion, exports.buildIdentifier);
+
+        var authMechanisms = getAuthMechanisms(apis);
+        var locals = {
+            apis: apis,
+            buildIdentifier: exports.buildIdentifier,
+            copyright: copyright,
+            enumTypes: collectEnumsFromApis(apis),
+            errorList: apis[0].errorList,
+            errors: apis[0].errors,
```
toddbellMSFT: Any particular reason why you want to use index 0 of apis?

pgilmorepf: Not specifically.
Every api document is expected to have a full copy of that particular information. It's copy paste in every API.
For now, you could pick list.any(), but that isn't a thing in JS.

PR #297, opened 2018-09-27
Adding a synchronous Send() when we lose focus.

toddbellMSFT:
Small fix ensures if we receive an OnApplicationFocus(false) event, we will force the HTTP call immediately.

### DIFF
--- a/targets/unity-v2/source/Entity/ScreenTimeTracker.cs
+++ b/targets/unity-v2/source/Entity/ScreenTimeTracker.cs
@@ -141,6 +141,14 @@ namespace PlayFab.Public
             eventsRequests.Enqueue(eventInfo);
 
             initialFocus = false;
+
+            if(!isFocused)
+            {
+                // Force the eventsRequests queue to empty.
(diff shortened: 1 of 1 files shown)


targets/unity-v2/source/Entity/ScreenTimeTracker.cs
```diff
@@ -141,6 +141,14 @@ public void OnApplicationFocus(bool isFocused)
             eventsRequests.Enqueue(eventInfo);
 
             initialFocus = false;
+
+            if(!isFocused)
+            {
+                // Force the eventsRequests queue to empty.
+                // If we are losing focus we should make an attempt to push out a focus lost event ASAP
+                Send();
```
pgilmorepf: You might not be sending the message you just Enqueued. You might be sending some earlier queued message.
Though, as I think about it, this might just be a general problem with this batching implementation.
I think this checkin is fine, but we're going to have to do a proper design pass on a real "batching" plugin for pluginmanager, and think carefully through our goals.

PR #299, opened 2018-09-27
adding verticalization to xplat

toddbellMSFT:


targets/XPlatCppSdk/make.js
```diff
@@ -260,3 +261,9 @@ function ifHasProps(datatype, displayText) {
         return "";
     return displayText;
 }
+
+function getDefaultVerticalization() {
+    if (exports.verticalName) {
+        return exports.verticalName;
+    }
```
pgilmorepf: No explicit default return.
you should add:
return "";

targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabSettings.cpp.ejs
```diff
@@ -48,7 +49,16 @@ namespace PlayFab
         fullUrl.reserve(1000);
 
         fullUrl += "https://";
-        fullUrl += titleId;
+
+        if(PlayFabSettings::verticalName != "")
```
pgilmorepf: Doing a string comparison is way more expensive than doing a length check.
if(PlayFabSettings::verticalName.length > 0)

PR #303, opened 2018-09-28
Making the targetMaker hack slightly less hacky, by moving it to a global

pgilmorepf:
Neither is ideal, but at least it defines more clearly where those variables come from, and how they get there.

### DIFF
--- a/SDKGenerator.njsproj
+++ b/SDKGenerator.njsproj
@@ -93,7 +93,7 @@
   <PropertyGroup Condition=" '$(Configuration)' == 'PlayFabGameServer' ">
     <DebugSymbols>true</DebugSymbols>
     <OutputPath>bin\</OutputPath>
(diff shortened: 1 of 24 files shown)


SDKGenerator.njsproj
```diff
@@ -93,7 +93,7 @@
   <PropertyGroup Condition=" '$(Configuration)' == 'PlayFabGameServer' ">
     <DebugSymbols>true</DebugSymbols>
     <OutputPath>bin\</OutputPath>
-    <ScriptArguments>csharp-unity-gameserver=..\sdks\PlayFabGameServer -apiSpecPath -buildIdentifier UnitySDK_manual</ScriptArguments>
+    <ScriptArguments>csharp-unity-gameserver=..\sdks\PlayFabGameServer -apiSpecPath -buildIdentifier UnitySDK_manual -verticalName testVerticalName</ScriptArguments>
```
Vicodex: >-verticalName testVerticalName [](start = 117, length = 30)

Do you need to keep this here by default?

pgilmorepf: DOH, nope.
Thanks!

PR #305, opened 2018-09-29
Update the git-branch destination to the new field

pgilmorepf:
Setting up GitDestBranch as the variable branch output instead of the more or less constant AUTOMATED_GIT_BRANCH and gitTarget, which really could only be "automated", or generally converted to a single other constant value, but was never able to be variable job-to-job.


pgilmorepf:
Apologies for self-review, but it's the weekend, and this blocks other tasks I hope to get done today.

PR #319, opened 2018-10-15
XplatCppSDK - fixed a memory leak when allocating a new json reader every time

Vicodex:
There was a JSON reader allocation memory leak bug in our SDK. It was pointed out by a customer. Fixing this.


targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp
```diff
@@ -191,6 +191,7 @@ namespace PlayFab
             Json::CharReader* jsonReader(jsonReaderFactory.newCharReader());
```
pgilmorepf: It seems to me that maybe jsonReader shouldn't be a pointer?
Maybe we should just create jsonReader here as a stack variable instead.
I think the change would be equally as small, and we won't have to figure out how to optimize or wrap a memory allocation later.

Vicodex: We discussed this offline, we receive the pointer from library's API newCharReader() and can't control if it's to be a pointer or an object. Resolving this for now.

PR #323, opened 2018-10-24
added unity changes for feature development acceleration.

toddbellMSFT:
Added unitySdk subfolder so the sdkGenerator can output to the correct folder.


targets/unity-v2/make.js
```diff
@@ -17,7 +19,7 @@ exports.MakeUnityV2Sdk = function (apis, sourceDir, apiOutputDir) {
     var locals = {
         errorList: apis[0].errorList,
         errors: apis[0].errors,
-        sdkVersion: sdkGlobals.sdkVersion,
```
pgilmorepf: It is quite important that you don't commit this line hard coding our version number to a specific number.
LOL

PR #325, opened 2018-10-30
adding tests for pubsub. 

toddbellMSFT:
Not expected to compile until the pubsub package is installed, so this shouldn't be exposed before that package is available.

### DIFF
--- a/targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
+++ b/targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
@@ -40,9 +40,9 @@ namespace PlayFab.UUnit
                 ec.Entity = new EventsModels.EntityKey();
                 ec.Entity.Id = _MyEntityKey.Id;
                 ec.Entity.Type = _MyEntityKey.Type;
-                ec.Name = "Ping_PubSub_Constructor";
+                ec.Name = "testevent";
 
-                ec.EventNamespace = "com.playfab.events.PubSubTestNamespace";
+                ec.EventNamespace = "com.playfab.events.test";
 
                 req.Events = new List<EventsModels.EventContents>();
                 req.Events.Add(ec);
--- a/targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
+++ b/targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
@@ -52,21 +52,8 @@ namespace PlayFab.UUnit
                 req.Events = new List<EventsModels.EventContents>();
                 req.Events.Add(ec);
 
-                PlayFabEventsAPI.WriteEvents(req, WriteEventSuccessful, WriteEventFail);
+                PlayFabEventsAPI.WriteEvents(req, null, null);
             }
-        }
-
-        public void WriteEventSuccessful(EventsModels.WriteEventsResponse response)
-        {
-            foreach (var id in response.AssignedEventIds)
-            {
-                Debug.Log("WriteEvent Succeeded and Id assigned: " + id);
-            }
-        }
-
-        public void WriteEventFail(PlayFabError error)
-        {
-            Debug.Log("Write Event Failed: " + error.ErrorMessage);
         }
 
         public override void TearDown(UUnitTestContext testContext)
--- a/targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
+++ b/targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
@@ -29,12 +29,6 @@ namespace PlayFab.UUnit
         {
             PlayFabSettings.TitleId = _previousTitleId;
             PlayFabSettings.VerticalName = null;
-
-            // if test didn't complete, fail
-            if (testContext.ActiveState != UUnitActiveState.COMPLETE)
-            {
-                testContext.Fail("PubSub UnitTest didn't recieve a message in time");
-            }
         }
 
         public override void ClassTearDown()
@@ -56,7 +50,6 @@ namespace PlayFab.UUnit
 
             PubSub pubSub = new PubSub(message =>
             {
-                Debug.Log("Received a Message: " + message);
                 ((UUnitTestContext) result.CustomData).EndTest(UUnitFinishState.PASSED, "PubSub construction successful");
             }, 
             new Topic { EventNamespace = "com.playfab.events.test", Name = "testevent", Entity = new Entity { Type = _MyEntityKey.Type, Id = _MyEntityKey.Id } }); // will this Entity conflict?
--- a/targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
+++ b/targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
@@ -94,14 +94,14 @@ namespace PlayFab.UUnit
 
             pubSub = new PubSub(message =>
             {
-                ((UUnitTestContext)result.CustomData).EndTest(UUnitFinishState.PASSED, "PubSub construction successful");
+                ((UUnitTestContext)result.CustomData).EndTest(UUnitFinishState.PASSED, "");
             },
             new Topic { EventNamespace = "com.playfab.events.test", Name = "testevent", Entity = new Entity { Type = _MyEntityKey.Type, Id = _MyEntityKey.Id } });
         }
 
         void LoginFailure(PlayFab.PlayFabError error)
         {
-            ((UUnitTestContext)error.CustomData).Fail("PubSub UnitTest Login Failed with the message: " + error.ErrorMessage);
+            ((UUnitTestContext)error.CustomData).Fail("PubSub UnitTest Login Failed with the message: " + error.GenerateErrorReport());
         }
     }
 }
--- a/targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
+++ b/targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
@@ -11,14 +11,30 @@ namespace PlayFab.UUnit
     {
         static PlayFab.EventsModels.EntityKey _MyEntityKey = null;
         static string _previousTitleId = null;
+        public override void SetUp(UUnitTestContext testContext)
+        {
+            // specific title id for relay test
+            _previousTitleId = PlayFabSettings.TitleId;
+            PlayFabSettings.TitleId = "70B02F89";
+            PlayFabSettings.VerticalName = "spi-relay";
+        }
 
         public override void Tick(UUnitTestContext testContext)
         {
             // No async work needed?
             // should this be a theoretical HubConnection SendAsync call to print a message out in the test?
         }
+
         public override void TearDown(UUnitTestContext testContext)
         {
+            PlayFabSettings.TitleId = _previousTitleId;
+            PlayFabSettings.VerticalName = null;
+
+            // if test didn't complete, fail
+            if (testContext.ActiveState != UUnitActiveState.COMPLETE)
+            {
+                testContext.Fail("PubSub UnitTest didn't recieve a message in time");
+            }
         }
 
         public override void ClassTearDown()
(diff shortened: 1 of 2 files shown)


targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
```diff
@@ -0,0 +1,73 @@
...
+            login.CreateAccount = true;
+            PlayFabClientAPI.LoginWithCustomID(login, LoginSuccess, LoginFailure, testContext);
+        }
+
+        void LoginSuccess(ClientModels.LoginResult result)
+        {
+            // This needs to be set in PlayFabSettings?... worst case scenario we pass this through the constructor
+            _MyEntityKey = new PlayFab.EventsModels.EntityKey { Id = result.EntityToken.Entity.Id, Type = result.EntityToken.Entity.Type };
+
+            // TODO: construct a PubSub object here
```
pgilmorepf: This todo seems done?

targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
```diff
@@ -0,0 +1,73 @@
...
+            PubSub pubSub = new PubSub(message =>
+            {
+                Console.WriteLine(message.ToString());
+                Debug.Log("Received a Message: " + message);
+            }, new Topic { EventNamespace = "com.playfab.events.test", Name = "testevent", Entity = new Entity { Type = _MyEntityKey.Type, Id = _MyEntityKey.Id } }); // will this Entity conflict?
+
+            // see if you can print some messages out from here
+            pubSub.PumpMessagesTest();
```
pgilmorepf: As best as I can tell, the only thing this test actually _does_, is confirm that this line doesn't throw an exception.
Is this a useful test if this is the only thing that's actually being tested?
Does this even throw an exception in normal operation?
I don't see anything that's actually testing an event receipt sequence.

targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        public override void ClassTearDown()
+        {
+        }
+
+        [UUnitTest]
+        public void TestPubSubConstruction(UUnitTestContext testContext)
+        {
+            //_testContext = testContext;
+            Debug.Log("Starting PubSub UnitTest");
```
pgilmorepf: Let's avoid spammy logs in tests.

targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            // specific title id for relay test
+            _previousTitleId = PlayFabSettings.TitleId;
```
pgilmorepf: This should be in the setup, not the test.

targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
```diff
@@ -0,0 +1,73 @@
...
+                Console.WriteLine(message.ToString());
+                Debug.Log("Received a Message: " + message);
+            }, new Topic { EventNamespace = "com.playfab.events.test", Name = "testevent", Entity = new Entity { Type = _MyEntityKey.Type, Id = _MyEntityKey.Id } }); // will this Entity conflict?
+
+            // see if you can print some messages out from here
+            pubSub.PumpMessagesTest();
+
+            Debug.Log("Ending PubSub UnitTest");
+            ((UUnitTestContext) result.CustomData).EndTest(UUnitFinishState.PASSED, "PubSub construction successful****************************************************************************************************");
+            PlayFabSettings.TitleId = _previousTitleId;
```
pgilmorepf: This should be in the teardown, not the test.

targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
```diff
@@ -0,0 +1,73 @@
...
+            //_testContext = testContext;
+            Debug.Log("Starting PubSub UnitTest");
+
+            // specific title id for relay test
+            _previousTitleId = PlayFabSettings.TitleId;
+            PlayFabSettings.TitleId = "70B02F89";
+            PlayFabSettings.VerticalName = "spi-relay";
+
+            ClientModels.LoginWithCustomIDRequest login = new ClientModels.LoginWithCustomIDRequest();
+            login.CustomId = "ToddIsTryingToTestPersistentSocketsInUnity" + Guid.NewGuid().ToString();
```
pgilmorepf: I'm not a fan of generating a new user every test execution. It's not going to break anything, but it's an overhead on the PlayFab system as a whole, which is unnecessary.

pgilmorepf: Also, you probably shouldn't commit your name into source code.

targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
```diff
@@ -0,0 +1,73 @@
...
+        void LoginSuccess(ClientModels.LoginResult result)
+        {
+            // This needs to be set in PlayFabSettings?... worst case scenario we pass this through the constructor
+            _MyEntityKey = new PlayFab.EventsModels.EntityKey { Id = result.EntityToken.Entity.Id, Type = result.EntityToken.Entity.Type };
+
+            // TODO: construct a PubSub object here
+            PubSub pubSub = new PubSub(message =>
+            {
+                Console.WriteLine(message.ToString());
+                Debug.Log("Received a Message: " + message);
```
pgilmorepf: What I'd really like to see is the testContext.EndTest() condition inside of this pubSub lambda.
That would ensure that a message is actually received from the server, in order for this test to pass.

targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
```diff
@@ -11,14 +11,30 @@ public class PubSubTest : UUnitTestCase
...
             // No async work needed?
             // should this be a theoretical HubConnection SendAsync call to print a message out in the test?
         }
+
         public override void TearDown(UUnitTestContext testContext)
         {
+            PlayFabSettings.TitleId = _previousTitleId;
+            PlayFabSettings.VerticalName = null;
+
+            // if test didn't complete, fail
```
pgilmorepf: This is unnecessary.
It's literally saying "if the test failed, then fail the test"
You generally shouldn't change the state of the test in teardown.

targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
```diff
@@ -28,44 +44,31 @@ public override void ClassTearDown()
...
         void LoginSuccess(ClientModels.LoginResult result)
         {
-            // This needs to be set in PlayFabSettings?... worst case scenario we pass this through the constructor
             _MyEntityKey = new PlayFab.EventsModels.EntityKey { Id = result.EntityToken.Entity.Id, Type = result.EntityToken.Entity.Type };
 
-            // TODO: construct a PubSub object here
             PubSub pubSub = new PubSub(message =>
             {
-                Console.WriteLine(message.ToString());
                 Debug.Log("Received a Message: " + message);
```
pgilmorepf: Now that the test flags itself as passed, the Debug.Log() here is redundant. You know this happened if the test passes.

targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
```diff
@@ -0,0 +1,103 @@
...
+    public class PubSubTest : UUnitTestCase
+    {
+        static PlayFab.EventsModels.EntityKey _MyEntityKey = null;
+        static string _previousTitleId = null;
+
+        private PubSub pubSub;
+        private Entity entity;
+
+        private int testTicks = 0;
+        private int playstreamTickDelay = 150;
```
pgilmorepf: Ticks is a fairly unreliable time measurement.
If I were writing this, I would typically define readonly Time elements, specifically TimeSpan: writeDelay, and DateTime nextWrite.
That's effectively what you have with your tick vars, but if the framerate were VERY LOW for some reason (like we're using machines on aws with really low hardware resources - which is very relevant), then you might fail because your tickrate is too slow.

targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
```diff
@@ -0,0 +1,103 @@
...
+                ec.EventNamespace = "com.playfab.events.test";
+
+                req.Events = new List<EventsModels.EventContents>();
+                req.Events.Add(ec);
+
+                PlayFabEventsAPI.WriteEvents(req, WriteEventSuccessful, WriteEventFail);
+            }
+        }
+
+        public void WriteEventSuccessful(EventsModels.WriteEventsResponse response)
```
pgilmorepf: If your callback is empty and does nothing, then I'd rather you just pass "null" in for that callback.

targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
pgilmorepf: Ticks is a fairly unreliable time measurement.
If I were writing this, I would typically define readonly Time elements, specifically TimeSpan: writeDelay, and DateTime nextWrite.
That's effectively what you have with your tick vars, but if the framerate were VERY LOW for some reason (like we're using machines on aws with really low hardware resources - which is very relevant), then you might fail because your tickrate is too slow.

targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            // check that we have assigned id's?
+            // this will probably get called a few times before the hub connection opens
+            // but once the hub connection opens it should immediatley start seeing something each Tick
+        }
+
+        public void WriteEventFail(PlayFabError error)
+        {
+            Debug.Log("Write Event Failed: " + error.ErrorMessage);
```
pgilmorepf: Is a WriteEvent failure a reason to fail the test?
I'd typically prefer a test failure over a Debug Log.
Generally I don't want tests to debug log at all.

targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
```diff
@@ -0,0 +1,103 @@
...
+            PlayFabClientAPI.LoginWithCustomID(login, LoginSuccess, LoginFailure, testContext);
+        }
+
+        void LoginSuccess(ClientModels.LoginResult result)
+        {
+            _MyEntityKey = new PlayFab.EventsModels.EntityKey { Id = result.EntityToken.Entity.Id, Type = result.EntityToken.Entity.Type };
+
+            pubSub = new PubSub(message =>
+            {
+                ((UUnitTestContext)result.CustomData).EndTest(UUnitFinishState.PASSED, "PubSub construction successful");
```
pgilmorepf: I don't really like fixed comments in the "passed" condition. Passed tells you that you got to this line and everything was great.
I usually only put dynamic information, which would be needed to debug this/other tests, in the pass condition. Like PlayFagId, or entityToken, or whatever. Static strings don't serve that purpose.
Change the string to null.

targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
```diff
@@ -0,0 +1,103 @@
...
+            },
+            new Topic { EventNamespace = "com.playfab.events.test", Name = "testevent", Entity = new Entity { Type = _MyEntityKey.Type, Id = _MyEntityKey.Id } });
+        }
+
+        void LoginFailure(PlayFab.PlayFabError error)
+        {
+            ((UUnitTestContext)error.CustomData).Fail("PubSub UnitTest Login Failed with the message: " + error.ErrorMessage);
```
pgilmorepf: There is a more thorough string that can be accessed: error.GenerateErrorReport().
ErrorMessage is only one of potentially many relevant strings contained within an error obj.

targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
toddbellMSFT: I've been using it for debugging purposes locally. I can check in null for now.

targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
```diff
@@ -52,23 +52,10 @@ public override void Tick(UUnitTestContext testContext)
                 req.Events = new List<EventsModels.EventContents>();
                 req.Events.Add(ec);
 
-                PlayFabEventsAPI.WriteEvents(req, WriteEventSuccessful, WriteEventFail);
+                PlayFabEventsAPI.WriteEvents(req, null, null);
```
pgilmorepf: Are you sure that a WriteEvents failure is correctly handled as "no-op"?
It might be relevant that a WriteEvents failure is just "test failure".

toddbellMSFT: WriteEvents failing may not be a direct indicator that PubSub is not working correctly. 

I don't think we should fail the test early for one WriteEvent failure. The test will fail at a timeout, but could potentially still get one successful WriteEvent.

PR #326, opened 2018-11-02
Fixed PluginManager, Unit Tests

Vicodex:
Important for the libHttpClient plugin story next week:

- Fixed PluginManager (moved from dangerous references to safe shared pointers)
- Added instanceability support (both singleton and instances of PluginManager are supported (request from a major customer, also a move towards much better memory management model)
- Restored unit tests and their infrastructure
- Generated SDK has been built and ran on Windows and Linux, verified along with unit tests

### DIFF
--- a/targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp
+++ b/targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp
@@ -10,16 +10,22 @@
 
 namespace PlayFab
 {
-    std::unique_ptr<IPlayFabHttp> IPlayFabHttp::httpInstance = nullptr;
+    std::shared_ptr<IPlayFabHttp> IPlayFabHttp::httpInstance = nullptr;
     IPlayFabHttp::~IPlayFabHttp() = default;
     IPlayFabHttp& IPlayFabHttp::Get()
+    {
+        return *GetPtr().get();
+    }
+
+    std::shared_ptr<IPlayFabHttp> IPlayFabHttp::GetPtr()
     {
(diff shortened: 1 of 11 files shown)


targets/XPlatCppSdk/source/cppsdk/source/playfab/PlayFabHttp.cpp
```diff
@@ -188,7 +194,7 @@ namespace PlayFab
         else
         {
             Json::CharReaderBuilder jsonReaderFactory;
-            std::unique_ptr<Json::CharReader> jsonReader(jsonReaderFactory.newCharReader());
```
Vicodex: >std::unique_ptr<Json::CharReader> jsonReader(jsonReaderFactory.newCharReader()); [](start = 12, length = 80)

Oops, this is a regression by accident (fixing it now)

targets/XPlatCppSdk/source/testapps/cppLinuxTestApp/main.cpp
```diff
@@ -15,7 +14,7 @@ void OnPlayFabFail(const PlayFab::PlayFabError& error, void*)
 
 void OnProfile(const PlayFab::ClientModels::GetPlayerProfileResult& result, void*)
 {
-    printf(("========== PlayFab Profile Success: " + result.PlayerProfile->PublisherId + "\n").c_str());
+    printf(("========== PlayFab Profile Success: " + result.PlayerProfile->PlayerId + "\n").c_str());
```
Vicodex: >PlayerId [](start = 75, length = 8)

I changed it for better visibility that we received something, because PublisherId was returning an empty string in this test

PR #327, opened 2018-11-03
QoS API

sanchitgarg:


targets/XPlatCppSdk/source/cppsdk/XPlatCppLinux.vcxproj.filters.ejs
```diff
@@ -48,10 +54,32 @@
...
+    </ClInclude>
+    <ClInclude Include="include\playfab\QoS\XPlatSocket.h">
+      <Filter>Header Files\playfab\QoS</Filter>
+    </ClInclude>
+    <ClInclude Include="source\playfab\PlayFabAdminApi.cpp">
   </ItemGroup>
   <ItemGroup>
     <ClCompile Include="stdafx.cpp">
-      <Filter>Source Files</Filter>
+      <Filter>Source Files\playfab</Filter>
```
Vicodex: >playfab [](start = 27, length = 7)

stdafx.cpp is not in "playfab" subfolder, this is not necessary. Consider removing this particular modification

targets/XPlatCppSdk/source/cppsdk/source/playfab/QoS/QoSSocket.cpp
```diff
@@ -0,0 +1,112 @@
...
+                else
+                {
+                    break;
+                }
+            }
+
+            // Snap the end time
+            chrono::steady_clock::time_point end = chrono::steady_clock::now();
+
+            // TODO sangarg : Optionala : Can add error checking for the reply and return based on that
```
Vicodex: >Optionala [](start = 30, length = 9)

Please correct minor spelling error

targets/XPlatCppSdk/source/cppsdk/include/playfab/QoS/PlayFabQoSApi.h
```diff
@@ -0,0 +1,53 @@
...
+#include <playfab/QoS/QoSSocket.h>
+#include <playfab/PlayFabMultiplayerDataModels.h>
+#include <playfab/PlayFabEventsDataModels.h>
+#include <playfab/PlayFabError.h>
+
+namespace PlayFab
+{
+    namespace QoS
+    {
+        PingResult GetQoSResultForDatacenter(std::shared_ptr<QoSSocket> socket);
```
Vicodex: >GetQoSResultForDatacenter [](start = 19, length = 25)

Since this function is global, please add a comment above it explaining what this function is for and whether it is meant to be called by user

targets/XPlatCppSdk/source/cppsdk/include/playfab/QoS/DataCenterResult.h
```diff
@@ -0,0 +1,33 @@
...
+        {
+        public:
+            DataCenterResult(const std::string& dataCenterName, int latencyMs, int errorCode);
+            DataCenterResult() = delete;
+            DataCenterResult(DataCenterResult&&) = default;
+            DataCenterResult(const DataCenterResult&) = delete;
+            DataCenterResult& operator=(DataCenterResult&&) = delete;
+            DataCenterResult& operator=(const DataCenterResult&) = delete;
+
+
```
Vicodex: Please remove an extra empty line

targets/XPlatCppSdk/source/cppsdk/include/playfab/QoS/QoS.h
```diff
@@ -0,0 +1,27 @@
...
+
+// define body for logging or debug output
+#define LOG_QOS(S) /*std::cout << S;*/
+
+namespace PlayFab
+{
+    namespace QoS
+    {
+        // the error code that is returned if a QoS exception was thrown
+        const int QOS_EXCEPTION_ERROR_CODE = -777;
```
toddbellMSFT: nit: Should these be static? Or classed? Right now they aren't marked public, but since they are not within a class I think they count as public to the QoS namespace?

With all caps for these names, makes me think they could be #DEFINE's. But that makes the code slightly uglier. They both seem to be very C like in nature (without the need for a class). 

Right now the call syntax would be QoS.QOS_*. My opinion would be to drop the initial QOS_ for each of these consts so we wouldn't need to read it twice (so just QoS.*). 

But the above is only true if you don't have using namespace QoS; (which is nested in PlayFab, would you want to be using namespace PlayFab.QoS? Or separate calls?). Are you expecting this?

toddbellMSFT: Could constexpr work here? similar to what you have in QoSResult?

targets/XPlatCppSdk/source/cppsdk/XPlatCppLinux.vcxproj.filters.ejs
pgilmorepf: agree.

PR #329, opened 2018-11-07
removing older signalR version

toddbellMSFT:


pgilmorepf:
I got confirmation from D&M that this aligns with our goals and won't cause unexpected issues.
Land it.

PR #330, opened 2018-11-07
Date deserialization fix

avl93:
ISerializer.cs uses CultureInfo.CurrentCulture to deserialize dates from JSON. It causes errors on some non-standard locales, like Thai one:
`ArgumentOutOfRangeException: Not a valid calendar for the given culture.`


pgilmorepf:
Thanks for your submission!
Approved and landed.

PR #331, opened 2018-11-09
creating some more tests regarding new PubSub behavior

toddbellMSFT:


targets/unity-v2/Testing/Tests/PersistentSockets/PubSubTest.cs
```diff
@@ -35,6 +35,48 @@ private void UpdateNextWriteTime()
         }
 
         public override void Tick(UUnitTestContext testContext)
+        {
```
toddbellMSFT: Note, This file shouldn't be here. There's still switch statements that shouldn't exist. recalling this file.

PR #335, opened 2018-11-14
ES6 SDK

DrewRomanyk:
This is creating a new SDK for ES6 which allows for new applications that use the modern web platform. This uses the latest exporting/importing syntax and also promises. This handles backwards compatibility by adding a polyfill for the library. This requires a new npm package to be installed for the repo in order to run the polyfill commands. The documentation for the SDK does not exist here, as i saw that other targets (such as the javascript and Node sdks ) didn't have the documentation in their target files, but I'm unsure where it should exist.


DrewRomanyk:
@pgilmorepf This is not an urgent delivery, as I can just run/bundle this code when I need to, so close this until there is more time?

PR #350, opened 2018-11-27
Revert the unreal engine build fix as the TOC has been updated

sanchitgarg:


Vicodex:
What was the fix that Paul did? Pulled down the problematic spec json?

PR #351, opened 2018-11-30
making Multiplayer api more flexible.

toddbellMSFT:


targets/unity-v2/make.js
```diff
@@ -22,7 +22,8 @@ exports.MakeUnityV2Sdk = function (apis, sourceDir, apiOutputDir) {
         sdkVersion: sdkGlobals.sdkVersion,
         buildIdentifier: sdkGlobals.buildIdentifier,
         hasClientOptions: getAuthMechanisms(apis).includes("SessionTicket"),
-        getVerticalNameDefault: getVerticalNameDefault
+        getVerticalNameDefault: getVerticalNameDefault,
+        isPartial: isPartial(apis.name)
```
pgilmorepf: apis is a list, and thus doesn't have a name.

pgilmorepf: I think this means that this change isn't necessary.

PR #354, opened 2018-12-05
added one-ds protocol handler, cs3.0 serializer, one-ds events api

Vicodex:


Vicodex:
It was reviewed in an SDK repo

PR #358, opened 2018-12-12
Implemented OneDS pipeline, minor fixes

Vicodex:
- also updated event router
- also added tests for lightweight events that use OneDS pipeline

### DIFF
--- a/targets/XPlatCppSdk/source/code/source/playfab/OneDSEventPipeline.cpp
+++ b/targets/XPlatCppSdk/source/code/source/playfab/OneDSEventPipeline.cpp
@@ -1,0 +1,167 @@
+#include <stdafx.h>
+
+#ifndef DISABLE_ONEDS_API
+
+#include <playfab/PlayFabTelemetryEventsApi.h>
+#include <playfab/OneDSEventPipeline.h>
+#include <playfab/OneDSEventsApi.h>
+
+#pragma warning (disable: 4100) // formal parameters are part of a public interface
+
+namespace PlayFab
+{
+    OneDSEventPipeline::OneDSEventPipeline(std::shared_ptr<OneDSEventPipelineSettings> settings) :
+        PlayFabEventPipeline(settings)
+    {
+    }
+
+    std::shared_ptr<OneDSEventPipelineSettings> OneDSEventPipeline::GetSettings() const
+    {
+        return std::dynamic_pointer_cast<OneDSEventPipelineSettings>(PlayFabEventPipeline::GetSettings());
+    }
+
+    void OneDSEventPipeline::SendBatch(size_t& batchCounter)
+    {
+        // create a WriteTelemetryEvents API request to send the batch
+        EventsModels::WriteEventsRequest batchReq;
+        for (const auto& eventEmitRequest : this->batch)
+        {
+            const auto& playFabEmitRequest = std::dynamic_pointer_cast<const PlayFabEmitEventRequest>(eventEmitRequest);
+            batchReq.Events.push_back(playFabEmitRequest->event->eventContents);
+        }
+
+        // add batch to flight tracking map
+        void* customData = reinterpret_cast<void*>(batchCounter); // used to track batches across asynchronous OneDS Events API
+        this->batchesInFlight[customData] = std::move(this->batch);
+        batchCounter++;
+
+        this->batch.clear(); // batch vector will be reused
+        this->batch.reserve(this->GetSettings()->maximalNumberOfItemsInBatch);
+
+        // authenticate OneDS API client as needed
+        if (!oneDSEventsApi.GetIsOneDSAuthenticated())
+        {
+            if (!AuthenticateOneDSApi(customData))
+                return;
+        }
+
+        // call OneDS Events API to send the batch
+        oneDSEventsApi.WriteTelemetryEvents(
+            batchReq,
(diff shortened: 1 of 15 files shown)


targets/XPlatCppSdk/source/code/source/playfab/PlayFabEventRouter.cpp
```diff
@@ -13,15 +13,45 @@ namespace PlayFab
...
-            pipelineEntry.second->IntakeEvent(request);
+            for (auto& pipelineEntry : this->pipelines)
+            {
+                switch (pfRequestPtr->event->eventType)
+                {
+                    case PlayFabEventType::Default:
+                    case PlayFabEventType::Lightweight:
+                    {
+                        // route lightweight (and default) events to OneDS pipeline only
+                        if (pipelineEntry.first == 1)
```
sanchitgarg: What are these numbers? the 1 and 0 later. Can we create an enum for it.

targets/XPlatCppSdk/source/code/source/playfab/OneDSEventPipeline.cpp
```diff
@@ -0,0 +1,167 @@
...
+                isOneDSAuthenticated = true;
+                operationComplete = true;
+            },
+            [&](const PlayFab::PlayFabError& error, void* relayedCustomData)
+            {
+                // failed to get OneDS context info (including credentials), relay error to user
+                WriteTelemetryEventsApiErrorCallback(error, customData);
+                isOneDSAuthenticated = false;
+                operationComplete = true;
+            });
```
pgilmorepf: While I understand most of the nuances about why this particular set of code will work, I am worried about it from a maintenance and code-cleanliness standpoint.

Specifically: Having async lambdas that refer to stack variables is dangerous in almost any other example. It works here, but it's a dangerous thing to mimic.

targets/XPlatCppSdk/source/code/source/playfab/OneDSHttpPlugin.cpp
```diff
@@ -66,7 +66,7 @@ namespace PlayFab
         {
             reqContainer.errorWrapper.HttpCode = 408;
```
debhaldarMS: >408 [](start = 49, length = 3)

Please log a bug.

targets/XPlatCppSdk/source/code/source/playfab/OneDSEventPipeline.cpp
pgilmorepf: I fail to see the point of relayedCustomData in this context.

targets/XPlatCppSdk/source/code/source/playfab/OneDSEventPipeline.cpp
```diff
@@ -0,0 +1,167 @@
...
+                    auto playFabError = std::shared_ptr<PlayFabError>(new PlayFabError());
+                    playFabError->HttpCode = 200;
+                    playFabError->ErrorCode = PlayFabErrorCode::PlayFabErrorSuccess;
+                    playFabEmitEventResponse->playFabError = playFabError;
+                    playFabEmitEventResponse->writeEventsResponse = std::shared_ptr<EventsModels::WriteEventsResponse>(new EventsModels::WriteEventsResponse(result));
+                    playFabEmitEventResponse->batch = requestBatchPtr;
+                    playFabEmitEventResponse->batchNumber = reinterpret_cast<size_t>(customData);
+
+                    // call an emit event callback
+                    playFabEmitRequest->callback(playFabEmitRequest->event, std::move(playFabEmitEventResponse));
```
pgilmorepf: You should always have a separate try/catch around customer callbacks.

targets/XPlatCppSdk/source/code/source/playfab/OneDSEventPipeline.cpp
```diff
@@ -0,0 +1,167 @@
...
+                {
+                    const auto& playFabEmitRequest = std::dynamic_pointer_cast<const PlayFabEmitEventRequest>(eventEmitRequest);
+                    auto playFabEmitEventResponse = std::shared_ptr<PlayFabEmitEventResponse>(new PlayFabEmitEventResponse());
+                    playFabEmitEventResponse->emitEventResult = EmitEventResult::Success;
+                    playFabEmitEventResponse->playFabError = std::shared_ptr<PlayFabError>(new PlayFabError(error));
+                    playFabEmitEventResponse->batch = requestBatchPtr;
+                    playFabEmitEventResponse->batchNumber = reinterpret_cast<size_t>(customData);
+
+                    // call an emit event callback
+                    playFabEmitRequest->callback(playFabEmitRequest->event, std::move(playFabEmitEventResponse));
```
pgilmorepf: You should always have separate callback around customer callbacks.

targets/XPlatCppSdk/source/code/source/playfab/OneDSEventPipeline.cpp
Vicodex: Changed pattern to use shared pointers instead of automatic variables as an example of a better (more reliable) memory model 

---
In reply to: [241204682](https://github.com/PlayFab/SDKGenerator/pull/358#discussion_r241204682) [](ancestors = 241204682)

Vicodex: It's a part of method signature that the callback must follow

---
In reply to: [241205630](https://github.com/PlayFab/SDKGenerator/pull/358#discussion_r241205630) [](ancestors = 241205630)

PR #366, opened 2018-12-21
Update the xplat autopkg files to create only one nuget

sanchitgarg:


targets/XPlatCppSdk/source/com.playfab.xplatcppsdk.v141.autopkg.ejs
```diff
@@ -40,7 +40,7 @@ nuget {
         };
 
         nestedInclude: {
-            #destination = ${d_include}\playfab;
+            #destination = ${d_include}playfab;
```
sanchitgarg: ${d_include} ends with a \
${d_include}\playfab creates an unnamed folder. Removed the \

PR #368, opened 2018-12-28
Implemented WinHTTP plugins for PF and OneDS

Vicodex:
- WinHTTP plugins for PF and OneDS (default on Windows platform). This HTTP stack is built in Windows and doesn't require any dependencies, hence it's ideal for this platform.
- Curl-based plugins for Windows are not really needed anymore (we can remove them at any time now along with their dependencies and git submodules)
- All tests and test app pass
- Everything is verified and tested

### DIFF
--- a/targets/XPlatCppSdk/source/code/include/playfab/PlayFabTransportHeaders.h
+++ b/targets/XPlatCppSdk/source/code/include/playfab/PlayFabTransportHeaders.h
@@ -1,12 +1,24 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// 
+// This header file is used to include headers of transport plugins supported on each platform.
+
 #pragma once
 
 #include <playfab/PlayFabPlatformMacros.h>
 
 #ifdef PLAYFAB_PLATFORM_XBOX
 #include <playfab/PlayFabIXHR2HttpPlugin.h>
+#include <playfab/OneDSIXHR2HttpPlugin.h>
 #endif // PLAYFAB_PLATFORM_XBOX
 
-#if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_LINUX) 
-#include <curl/curl.h>
+#ifdef PLAYFAB_PLATFORM_WINDOWS
+#include <playfab/PlayFabWinHttpPlugin.h>
 #include <playfab/PlayFabHttp.h>
-#endif // PLAYFAB_PLATFORM_WINDOWS || PLAYFAB_PLATFORM_LINUX
+#include <playfab/OneDSWinHttpPlugin.h>
+#include <playfab/OneDSHttpPlugin.h>
+#endif // PLAYFAB_PLATFORM_WINDOWS
+
+#ifdef PLAYFAB_PLATFORM_LINUX
+#include <playfab/PlayFabHttp.h>
+#include <playfab/OneDSHttpPlugin.h>
+#endif // PLAYFAB_PLATFORM_LINUX
--- a/targets/XPlatCppSdk/source/code/include/playfab/PlayFabWinHttpPlugin.h
+++ b/targets/XPlatCppSdk/source/code/include/playfab/PlayFabWinHttpPlugin.h
@@ -4,6 +4,7 @@
 #include <playfab/PlayFabPluginManager.h>
 #include <thread>
 #include <mutex>
+#include <atomic>
 #include <windows.h>
 #include <winhttp.h>
 
@@ -37,7 +38,7 @@ namespace PlayFab
 
         std::thread workerThread;
         std::mutex httpRequestMutex;
-        bool threadRunning;
+        std::atomic<bool> threadRunning;
         int activeRequestCount;
         std::deque<std::unique_ptr<CallRequestContainerBase>> pendingRequests;
         std::deque<std::unique_ptr<CallRequestContainerBase>> pendingResults;
--- a/targets/XPlatCppSdk/source/code/include/playfab/PlayFabWinHttpPlugin.h
+++ b/targets/XPlatCppSdk/source/code/include/playfab/PlayFabWinHttpPlugin.h
@@ -1,0 +1,45 @@
+#pragma once
+
+#include <playfab/PlayFabCallRequestContainer.h>
+#include <playfab/PlayFabPluginManager.h>
+#include <thread>
+#include <mutex>
+#include <windows.h>
+#include <winhttp.h>
+
+namespace PlayFab
+{
+    /// <summary>
+    /// PlayFabWinHttpPlugin is an https implementation to interact with PlayFab services using WinHTTP.
+    /// </summary>
+    class PlayFabWinHttpPlugin : public IPlayFabHttpPlugin
+    {
+    public:
+        PlayFabWinHttpPlugin();
+        PlayFabWinHttpPlugin(const PlayFabWinHttpPlugin& other) = delete;
+        PlayFabWinHttpPlugin(PlayFabWinHttpPlugin&& other) = delete;
+        PlayFabWinHttpPlugin& operator=(PlayFabWinHttpPlugin&& other) = delete;
+        virtual ~PlayFabWinHttpPlugin();
+
+        virtual void MakePostRequest(std::unique_ptr<CallRequestContainerBase> requestContainer) override;
+        virtual size_t Update() override;
+
+    protected:
+        virtual void ExecuteRequest(std::unique_ptr<CallRequestContainer> requestContainer);
+        virtual std::string GetUrl(CallRequestContainer& requestContainer) const;
+        virtual void SetPredefinedHeaders(CallRequestContainer& requestContainer, HINTERNET hRequest);
+        virtual bool GetBinaryPayload(CallRequestContainer& requestContainer, LPVOID& payload, DWORD& payloadSize) const;
+        virtual void ProcessResponse(CallRequestContainer& requestContainer, const int httpCode);
+        void WorkerThread();
+        void HandleCallback(std::unique_ptr<CallRequestContainer> requestContainer);
+        void HandleResults(std::unique_ptr<CallRequestContainer> requestContainer);
+        void SetErrorInfo(CallRequestContainer& requestContainer, const std::string& errorMessage, const int httpCode = 0) const;
+
+        std::thread workerThread;
+        std::mutex httpRequestMutex;
+        bool threadRunning;
+        int activeRequestCount;
+        std::deque<std::unique_ptr<CallRequestContainerBase>> pendingRequests;
+        std::deque<std::unique_ptr<CallRequestContainerBase>> pendingResults;
+    };
+}
--- a/targets/XPlatCppSdk/source/code/source/playfab/OneDSHttpPlugin.cpp
+++ b/targets/XPlatCppSdk/source/code/source/playfab/OneDSHttpPlugin.cpp
@@ -3,8 +3,7 @@
 #ifndef DISABLE_ONEDS_API
 
 #include <playfab/OneDSHttpPlugin.h>
-
-#include <playfab/PlayFabTransportHeaders.h>
+#include <curl/curl.h>
 
 namespace PlayFab
 {
--- a/targets/XPlatCppSdk/source/code/source/playfab/OneDSWinHttpPlugin.cpp
+++ b/targets/XPlatCppSdk/source/code/source/playfab/OneDSWinHttpPlugin.cpp
@@ -11,9 +11,11 @@
 
 namespace PlayFab
 {
+    constexpr auto ONEDS_SERVICE_URL = "https://self.events.data.microsoft.com/OneCollector/1.0/";
+
     std::string OneDSWinHttpPlugin::GetUrl(CallRequestContainer& requestContainer) const
     {
-        return std::string("https://self.events.data.microsoft.com/OneCollector/1.0/");
+        return std::string(ONEDS_SERVICE_URL);
     }
 
     void OneDSWinHttpPlugin::SetPredefinedHeaders(CallRequestContainer& requestContainer, HINTERNET hRequest)
@@ -35,7 +37,7 @@ namespace PlayFab
     bool OneDSWinHttpPlugin::GetBinaryPayload(CallRequestContainer& requestContainer, LPVOID& payload, DWORD& payloadSize) const
     {
         OneDSCallRequestContainer& reqContainer = (OneDSCallRequestContainer&)requestContainer;
-        payloadSize = (DWORD)reqContainer.requestBinaryBody.size();
+        payloadSize = static_cast<DWORD>(reqContainer.requestBinaryBody.size());
         payload = reqContainer.requestBinaryBody.data();
         return true;
     }
--- a/targets/XPlatCppSdk/source/code/source/playfab/OneDSWinHttpPlugin.cpp
+++ b/targets/XPlatCppSdk/source/code/source/playfab/OneDSWinHttpPlugin.cpp
@@ -1,0 +1,111 @@
+#include <stdafx.h>
+
+#ifndef DISABLE_ONEDS_API
+
+#include <playfab/OneDSWinHttpPlugin.h>
+#include <windows.h>
+#include <winhttp.h>
+
+#pragma warning (disable: 4100) // formal parameters are part of a public interface
+#pragma warning (disable: 4245) // Some DWORD arguments of WinHTTP API can accept -1, according to documentation
+
+namespace PlayFab
+{
+    std::string OneDSWinHttpPlugin::GetUrl(CallRequestContainer& requestContainer) const
+    {
+        return std::string("https://self.events.data.microsoft.com/OneCollector/1.0/");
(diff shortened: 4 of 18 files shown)


debhaldarMS:
Can we get rid of the necessity to build jsoncpp for clients as well?

Vicodex:
Not yet. Unfortunately jsoncpp is the only dependency that we still need across the board, but it is (thankfully) also easy to build

---
In reply to: [450943902](https://github.com/PlayFab/SDKGenerator/pull/368#issuecomment-450943902) [](ancestors = 450943902)

Vicodex:
[https://playfab.visualstudio.com/PlayFab/_workitems/edit/22593](https://playfab.visualstudio.com/PlayFab/_workitems/edit/22593)

---
In reply to: [450943078](https://github.com/PlayFab/SDKGenerator/pull/368#issuecomment-450943078) [](ancestors = 450943078)

debhaldarMS:
Why don't we distribute jsoncpp as part of the project ?

---
In reply to: [450986014](https://github.com/PlayFab/SDKGenerator/pull/368#issuecomment-450986014) [](ancestors = 450986014,450943902)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabTransportHeaders.h
```diff
@@ -1,12 +1,24 @@
...
 #ifdef PLAYFAB_PLATFORM_XBOX
 #include <playfab/PlayFabIXHR2HttpPlugin.h>
+#include <playfab/OneDSIXHR2HttpPlugin.h>
 #endif // PLAYFAB_PLATFORM_XBOX
 
-#if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_LINUX) 
-#include <curl/curl.h>
+#ifdef PLAYFAB_PLATFORM_WINDOWS
+#include <playfab/PlayFabWinHttpPlugin.h>
+#include <playfab/PlayFabHttp.h>
```
debhaldarMS: remove ?

targets/XPlatCppSdk/source/code/include/playfab/PlayFabTransportHeaders.h
```diff
@@ -1,12 +1,24 @@
...
+#include <playfab/OneDSWinHttpPlugin.h>
+#include <playfab/OneDSHttpPlugin.h>
```
debhaldarMS: >OneDSHttpPlugin [](start = 18, length = 15)

remove ?

targets/XPlatCppSdk/source/code/source/playfab/OneDSWinHttpPlugin.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+        WinHttpAddRequestHeaders(hRequest, L"Accept: */*", -1, 0);
+        WinHttpAddRequestHeaders(hRequest, L"Client-Id: NO_AUTH", -1, 0);
+        WinHttpAddRequestHeaders(hRequest, L"Content-Type: application/bond-compact-binary", -1, 0);
+        WinHttpAddRequestHeaders(hRequest, L"Expect: 100-continue", -1, 0);
+        WinHttpAddRequestHeaders(hRequest, L"SDK-Version: EVT-PlayFab-XPlat-C++-No-3.0.275.1", -1, 0);
+        int64_t now = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+        WinHttpAddRequestHeaders(hRequest, (L"Upload-Time: " + std::to_wstring(now)).c_str(), -1, 0);
+        WinHttpAddRequestHeaders(hRequest, (L"Content-Length: " + std::to_wstring(reqContainer.requestBinaryBody.size())).c_str(), -1, 0);
+        WinHttpAddRequestHeaders(hRequest, L"Connection: Keep-Alive", -1, 0);
```
debhaldarMS: >    WinHttpAddRequestHeaders(hRequest, L"Connection: Keep-Alive", -1, 0); [](start = 4, length = 73)

Can you please point me to the documentation ? We should strive to use the keepalive feature to reuse the http connections

targets/XPlatCppSdk/source/code/source/playfab/OneDSWinHttpPlugin.cpp
```diff
@@ -0,0 +1,111 @@
...
+        WinHttpAddRequestHeaders(hRequest, L"Cache-Control: no-cache", -1, 0);
+    }
+
+    bool OneDSWinHttpPlugin::GetBinaryPayload(CallRequestContainer& requestContainer, LPVOID& payload, DWORD& payloadSize) const
+    {
+        OneDSCallRequestContainer& reqContainer = (OneDSCallRequestContainer&)requestContainer;
+        payloadSize = (DWORD)reqContainer.requestBinaryBody.size();
```
debhaldarMS: >(DWORD) [](start = 22, length = 7)

avoid C style casts

targets/XPlatCppSdk/source/code/source/playfab/OneDSHttpPlugin.cpp
```diff
@@ -3,8 +3,7 @@
 #ifndef DISABLE_ONEDS_API
 
 #include <playfab/OneDSHttpPlugin.h>
-
-#include <playfab/PlayFabTransportHeaders.h>
+#include <curl/curl.h>
```
sanchitgarg: Do we want to keep all transport related includes in the PlayFabTransportHeaders.h? That was the original design of adding the header.

targets/XPlatCppSdk/source/code/source/playfab/OneDSWinHttpPlugin.cpp
```diff
@@ -0,0 +1,111 @@
...
+            JSONCPP_STRING jsonParseErrors;
+            const bool parsedSuccessfully = jsonReader->parse(reqContainer.responseString.c_str(), reqContainer.responseString.c_str() + reqContainer.responseString.length(), &reqContainer.responseJson, &jsonParseErrors);
+
+            if (parsedSuccessfully)
+            {
+                auto result = reqContainer.responseJson.get("acc", Json::Value::null).asInt();
+                if (result > 0)
+                {
+                    // fully successful response
+                    reqContainer.errorWrapper.HttpCode = httpCode != 0 ? httpCode : 200;
```
sanchitgarg: [Curious] : In what scenario can the returned HttpCode be 0?

Vicodex:
Its binary (and headers) content is distributed as part of our SDK (the lib is literally linked into ours). Its source code is (sort of) part of the project too (in the meaning that its source code is pulled together with the source code of our SDK). But its vcxproj file is not part of our solution and currently needs to be built separately (simply followed the pattern with other external dependencies here). But we probably could make it a part of our build, since we have all the source code. Not as part of this particular CR though, i would prefer to have a separate dedicated task and PR for that.

---
In reply to: [450996322](https://github.com/PlayFab/SDKGenerator/pull/368#issuecomment-450996322) [](ancestors = 450996322,450986014,450943902)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabTransportHeaders.h
Vicodex: We are not removing anything for Windows as part of this (it may also break many things at once). There will be a separate task/story for it. 

---
In reply to: [244867250](https://github.com/PlayFab/SDKGenerator/pull/368#discussion_r244867250) [](ancestors = 244867250)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabTransportHeaders.h
Vicodex: We are not removing anything for Windows as part of this (it may also break many things at once). There will be a separate task/story for it. 

---
In reply to: [244867295](https://github.com/PlayFab/SDKGenerator/pull/368#discussion_r244867295) [](ancestors = 244867295)

targets/XPlatCppSdk/source/code/source/playfab/OneDSWinHttpPlugin.cpp
Vicodex: That header was literally taken from Aria's SDK and our plugin is a substitute for it.

---
In reply to: [244868108](https://github.com/PlayFab/SDKGenerator/pull/368#discussion_r244868108) [](ancestors = 244868108)

targets/XPlatCppSdk/source/code/source/playfab/OneDSWinHttpPlugin.cpp
```diff
@@ -0,0 +1,111 @@
...
+#include <winhttp.h>
+
+#pragma warning (disable: 4100) // formal parameters are part of a public interface
+#pragma warning (disable: 4245) // Some DWORD arguments of WinHTTP API can accept -1, according to documentation
+
+namespace PlayFab
+{
+    std::string OneDSWinHttpPlugin::GetUrl(CallRequestContainer& requestContainer) const
+    {
+        return std::string("https://self.events.data.microsoft.com/OneCollector/1.0/");
```
debhaldarMS: >"https://self.events.data.microsoft.com/OneCollector/1.0/" [](start = 27, length = 58)

constexpr

targets/XPlatCppSdk/source/code/source/playfab/OneDSHttpPlugin.cpp
Vicodex: No, we do not want to have a common shared header that includes the whole world (even separated by platform macros). Specific dependencies must stay only within the CPP code of particular plugins if they are not exposed publicly through their headers.

---
In reply to: [244868921](https://github.com/PlayFab/SDKGenerator/pull/368#discussion_r244868921) [](ancestors = 244868921)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabPluginManager.cpp
```diff
@@ -79,6 +79,8 @@ namespace PlayFab
     {
 #ifdef PLAYFAB_PLATFORM_XBOX
         return std::shared_ptr<PlayFabIXHR2HttpPlugin>(new PlayFabIXHR2HttpPlugin());
+#elif defined(PLAYFAB_PLATFORM_WINDOWS)
+        return std::shared_ptr<PlayFabWinHttpPlugin>(new PlayFabWinHttpPlugin());
```
debhaldarMS: >std::shared_ptr<PlayFabWinHttpPlugin>(new PlayFabWinHttpPlugin()); [](start = 15, length = 66)

curious why you're using new instead of make_Shared ?

targets/XPlatCppSdk/source/code/include/playfab/PlayFabWinHttpPlugin.h
```diff
@@ -0,0 +1,45 @@
...
+        virtual bool GetBinaryPayload(CallRequestContainer& requestContainer, LPVOID& payload, DWORD& payloadSize) const;
+        virtual void ProcessResponse(CallRequestContainer& requestContainer, const int httpCode);
+        void WorkerThread();
+        void HandleCallback(std::unique_ptr<CallRequestContainer> requestContainer);
+        void HandleResults(std::unique_ptr<CallRequestContainer> requestContainer);
+        void SetErrorInfo(CallRequestContainer& requestContainer, const std::string& errorMessage, const int httpCode = 0) const;
+
+        std::thread workerThread;
+        std::mutex httpRequestMutex;
+        bool threadRunning;
```
debhaldarMS: >  bool threadRunning; [](start = 6, length = 21)

should this be an atomic bool ?

targets/XPlatCppSdk/source/code/source/playfab/PlayFabPluginManager.cpp
Vicodex: To remain compatible with C++ 11 and Paul pointed out that we have customers that required C++ 11 (not later)

---
In reply to: [244876289](https://github.com/PlayFab/SDKGenerator/pull/368#discussion_r244876289) [](ancestors = 244876289)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -0,0 +1,378 @@
...
+    {
+        activeRequestCount = 0;
+        threadRunning = true;
+        workerThread = std::thread(&PlayFabWinHttpPlugin::WorkerThread, this);
+    };
+
+    PlayFabWinHttpPlugin::~PlayFabWinHttpPlugin()
+    {
+        threadRunning = false;
+        workerThread.join();
```
debhaldarMS: >  workerThread.join(); [](start = 6, length = 22)

this might throw if the thread is not joinable

targets/XPlatCppSdk/source/code/include/playfab/PlayFabWinHttpPlugin.h
Vicodex: in most cases making bool atomic is an overkill. With this particular bool i am not introducing any new pattern here. If you insist on making this atomic we should change that in all 6 other plugins too.

---
In reply to: [244876563](https://github.com/PlayFab/SDKGenerator/pull/368#discussion_r244876563) [](ancestors = 244876563)

targets/XPlatCppSdk/source/code/source/playfab/OneDSWinHttpPlugin.cpp
Vicodex: I wanted to be super-safe here and return a real HTTP code when possible, but also fallback to what application considers a successful call in situations when a real HTTP code was not returned (for example because WinHttpQueryHeaders API call failed in some implicit way). The HTTP call itself is not necessarily a failure, in this situation, it just we didn't get its response code, since it's a separate API. If we get to the point that we got "a response" in the form of "some result content" from the server (and not just an HTTP error code), then chances are close to 99% that it was 200.

---
In reply to: [244869709](https://github.com/PlayFab/SDKGenerator/pull/368#discussion_r244869709) [](ancestors = 244869709)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabWinHttpPlugin.cpp
Vicodex: I agree, but it won't be not joinable because we control all 3 conditions of it:
[http://www.cplusplus.com/reference/thread/thread/joinable/](http://www.cplusplus.com/reference/thread/thread/joinable/)

---
In reply to: [244876721](https://github.com/PlayFab/SDKGenerator/pull/368#discussion_r244876721) [](ancestors = 244876721)

targets/XPlatCppSdk/source/code/source/playfab/OneDSWinHttpPlugin.cpp
Vicodex: I actually found a nice and simple explanation here:
[https://en.wikipedia.org/wiki/HTTP_persistent_connection](https://en.wikipedia.org/wiki/HTTP_persistent_connection)


---
In reply to: [244875698](https://github.com/PlayFab/SDKGenerator/pull/368#discussion_r244875698) [](ancestors = 244875698,244868108)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabPluginManager.cpp
debhaldarMS: make_shared is C++11 -> make_Unique is not

---
In reply to: [244876570](https://github.com/PlayFab/SDKGenerator/pull/368#discussion_r244876570) [](ancestors = 244876570,244876289)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabWinHttpPlugin.cpp
debhaldarMS: Yes - now we do -- we're not sure how the code will morph over time - better be safe than sorry :)

---
In reply to: [244881236](https://github.com/PlayFab/SDKGenerator/pull/368#discussion_r244881236) [](ancestors = 244881236,244876721)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabWinHttpPlugin.h
debhaldarMS: Yes - let's change them all through

---
In reply to: [244877114](https://github.com/PlayFab/SDKGenerator/pull/368#discussion_r244877114) [](ancestors = 244877114,244876563)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabPluginManager.cpp
Vicodex: Ah, good catch! I''l fix that


---
In reply to: [244893864](https://github.com/PlayFab/SDKGenerator/pull/368#discussion_r244893864) [](ancestors = 244893864,244876570,244876289)

PR #369, opened 2019-01-04
Removed Curl-based plugins Curl dependencies for Windows platform in XplatCppSDK

Vicodex:
- Curl-based plugins and dependencies are not used on Windows platform anymore and are removed from Windows solution/projects
- Curl-based plugins are still used as the only HTTP plugin implementation for Linux (but no dependencies need to be built since on Linux they are considered part of the platform)
- Curl-based plugins for Linux were renamed to be consistent with the naming convention used in all other HTTP plugins (PlayFabHttp -> PlayFabCurlHttpPlugin, OneDSHTTPPlugin -> OneDSCurlHttpPlugin)
- PluginManager is updated to automatically instantiate the default OneDS HTTP plugin for each platform, similar to how the default PlayFab HTTP plugin is instantiated. It required to be set explicitly before, which was a bug
- Tests and Test app are verified on Windows and Linux
- JSON cpp dependency is still needed and i will look into ways to incorporate it directly into Windows and Xbox solutions in a separate PR
- Documentation (MDs) will be updated with JSON PR since it will require changes from both PRs.


targets/XPlatCppSdk/source/build/Windows/XPlatCppWindows.vcxproj.filters.ejs
```diff
@@ -26,7 +26,7 @@
     <ClInclude Include="$(SdkSourceDir)\include\playfab\PlayFabPlatformMacros.h">
       <Filter>Header Files\playfab</Filter>
     </ClInclude>
-    <ClInclude Include="$(SdkSourceDir)\include\playfab\OneDSHttpPlugin.h">
+    <ClInclude Include="$(SdkSourceDir)\include\playfab\OneDSCurlHttpPlugin.h">
```
sanchitgarg: We can probably remove them from the filter as well since we arent building it.

Vicodex: Good catch!

---
In reply to: [245173781](https://github.com/PlayFab/SDKGenerator/pull/369#discussion_r245173781) [](ancestors = 245173781)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabPluginManager.cpp
```diff
@@ -82,7 +85,18 @@ namespace PlayFab
 #elif defined(PLAYFAB_PLATFORM_WINDOWS)
         return std::make_shared<PlayFabWinHttpPlugin>();
 #else
-        return std::make_shared<PlayFabHttp>();
+        return std::make_shared<PlayFabCurlHttpPlugin>();
+#endif // PLAYFAB_PLATFORM_XBOX
```
sanchitgarg: nit : PLAYFAB_PLATFORM_WINDOWS

PR #371, opened 2019-01-08
Sets a custom GetPlugin Comparer

toddbellMSFT:
Fix for bug 6418. Reduces our allocations per frame from 488 to 32 bytes.


targets/unity-v2/source/Shared/Public/PluginContractKey.cs
```diff
@@ -0,0 +1,27 @@
...
+using PlayFab.Json;
+
+namespace PlayFab
+{
+    public class PluginContractKey
+    {
+        public PluginContract _pluginContract;
+        public string _pluginName;
+    }
+
```
pgilmorepf: Please use a line break between classes.

PR #385, opened 2019-01-31
Fix UE4 ExampleProject crash in standalone build by lazy-loading data

M-Pixel:
Move UE4 C++ SDK test setup (LoadTitleData) to lazy-load from first test.  Previously, it was called from the constructor, however the test class (as per UE4 convention) must have a static instance.  In standalone builds, it was constructing before the log system was ready, causing a crash when the file-loading method attempted to log.  `FAutomationTestBase` does not have a "setup" override, and `GetTests` which is the closest thing to a setup function is declared `const`, so the remaining solution was to set up in `RunTest`.

### DIFF
--- a/targets/UnrealMarketplacePlugin/examplesource/ExampleProject/Source/ExampleProject/PlayFabApiTests.h
+++ b/targets/UnrealMarketplacePlugin/examplesource/ExampleProject/Source/ExampleProject/PlayFabApiTests.h
@@ -319,16 +319,7 @@ public:
     FString CLOUD_FUNCTION_THROW_ERROR = TEXT("throwError");
 
     // Input from TestTitleData.json
-    TestTitleData testTitleDataStorage;
-    TestTitleData& GetTestTitleData()
-    {
-        if (!testTitleDataStorage.loaded)
-        {
-            testTitleDataStorage.loaded = LoadTitleData();
-            // IPlayFabModuleInterface::Get().SetTitleInformationFromJson(//todo);
-        }
-        return testTitleDataStorage;
-    }
+    TestTitleData testTitleData;
 
     PlayFabApiTestSuite(const FString& InName)
         : FAutomationTestBase(InName, false)
@@ -384,9 +375,11 @@ protected:
             success &= FJsonSerializer::Deserialize(jsonReader, jsonParsed);
         }
 
-        if (success) success &= jsonParsed->TryGetStringField("titleId", testTitleDataStorage.titleId);
-        if (success) success &= jsonParsed->TryGetStringField("developerSecretKey", testTitleDataStorage.developerSecretKey);
-        if (success) success &= jsonParsed->TryGetStringField("userEmail", testTitleDataStorage.userEmail);
+        if (success) success &= jsonParsed->TryGetStringField("titleId", testTitleData.titleId);
+        if (success) success &= jsonParsed->TryGetStringField("developerSecretKey", testTitleData.developerSecretKey);
+        if (success) success &= jsonParsed->TryGetStringField("userEmail", testTitleData.userEmail);
+
+        testTitleData.loaded = success;
 
         return success;
     }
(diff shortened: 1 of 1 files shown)


targets/UnrealMarketplacePlugin/examplesource/ExampleProject/Source/ExampleProject/PlayFabApiTests.h
```diff
@@ -421,7 +431,9 @@ class PlayFabApiTestSuite : public FAutomationTestBase
 
     bool InvalidLogin() const
     {
-        ADD_LATENT_AUTOMATION_COMMAND(PlayFabApiTest_0LoginWithEmail(testTitleData.userEmail, INVALID_PASSWORD));
+        if (!testTitleDataStorage.loaded)
+            return false;
```
Vicodex: If RunTest is called for each test function, then this check might not be necessary? (RunTest already has it)

targets/UnrealMarketplacePlugin/examplesource/ExampleProject/Source/ExampleProject/PlayFabApiTests.h
```diff
@@ -318,15 +319,20 @@ class PlayFabApiTestSuite : public FAutomationTestBase
     FString CLOUD_FUNCTION_THROW_ERROR = TEXT("throwError");
 
     // Input from TestTitleData.json
-    TestTitleData testTitleData;
+    TestTitleData testTitleDataStorage;
```
Vicodex: >testTitleDataStorage [](start = 18, length = 20)

I would keep the old name (just to change less code) but i don't insist, up to you :)

M-Pixel: I had a choice between this and changing the name of the local variable assigned from `GetTestTitleData()` in `LoadTestTitleData()`, in order to avoid a name-is-hidden warning.  I thought that changing the member to xStorage would be more clear than anything I could come up with for alternate names for the local.

targets/UnrealMarketplacePlugin/examplesource/ExampleProject/Source/ExampleProject/PlayFabApiTests.h
M-Pixel: You're right - totally redundant.  Will remove.

targets/UnrealMarketplacePlugin/examplesource/ExampleProject/Source/ExampleProject/PlayFabApiTests.h
pgilmorepf: Change the local variable in the other location, and just use tempTestTitleData or localTestTitleData or myTestTitleData.
I consider it a "priority" thing, where this variable is more important than a local variable.

pgilmorepf: Honestly though... they should never be different, never change for the life of this test, and it should only be loaded once...
So the local variable just shouldn't exist it all. Everything should be referencing this one.

targets/UnrealMarketplacePlugin/examplesource/ExampleProject/Source/ExampleProject/PlayFabApiTests.h
```diff
@@ -318,15 +319,20 @@ class PlayFabApiTestSuite : public FAutomationTestBase
...
+    TestTitleData& GetTestTitleData()
+    {
+        if (!testTitleDataStorage.loaded)
+        {
+            testTitleDataStorage.loaded = LoadTitleData();
+            // IPlayFabModuleInterface::Get().SetTitleInformationFromJson(//todo);
```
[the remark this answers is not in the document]
M-Pixel: Even if it was there beforehand?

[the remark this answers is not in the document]
pgilmorepf: This diff indicates that you added it. But yes. It's clearly not needed, delete it here, and if there is other commented code elsewhere, please remove it if it's clearly not needed.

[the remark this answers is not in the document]
pgilmorepf: OIC, the lines don't match up.
Yes, delete it.

PR #390, opened 2019-02-04
Add Example Project Build batch script

Vallantyn:
Lots of flags on the command line, this will probably need to be refined later on.

I intentionnaly didn't put any pause command at the end, since this won't be executed by a human.

### DIFF
--- a/SDKBuildScripts/unity_build_beta.bat
+++ b/SDKBuildScripts/unity_build_beta.bat
@@ -1,5 +1,0 @@
-setlocal
-set sdkGenArgs=-flags beta
-
-call unity_build.bat
-endlocal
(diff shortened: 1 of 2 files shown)


Vallantyn:
Also without #385 the build actually fails.

SDKBuildScripts/unity_build_beta.bat
```diff
@@ -0,0 +1,5 @@
+setlocal
```
pgilmorepf: Don't commit this file.
Rename this file to "beta_unity.bat", so that it is filtered by the .gitignore, and that will be sufficient.

Vallantyn: Done !
Still getting used to how pull request works, the first two commits weren't really intended to be there. I'll try to be more carefull about what's inside a PR.

PR #392, opened 2019-02-07
Go initial support

dgkanatsios:
This is some initial work to support Go. 
An export of it here: https://github.com/dgkanatsios/playfabsdk-go 
More tests and examples are definitely needed, of course!

edit: I can't GZIP compression to work, I'd appreciate some hints on what I should checkout in order to debug properly.

A small nit is that I was not able to get the value of sdkVersion variable, not sure why.


targets/go/source/PlayFabHttp.go.ejs
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if result.Code != 200 {
+		errorString := fmt.Sprintf("error code %d", result.Code)
+		if result.Status != "" {
+			errorString = fmt.Sprintf("%s and status %s", errorString, result.Status)
+		}
+		return nil, errors.New(errorString)
```
MarcoWilliamsPF: This should create structured data for the error when it fails.  Go looks similar to the NodeSDK.. you might want to check that out here: https://github.com/PlayFab/NodeSDK/blob/master/PlayFabSdk/Scripts/PlayFab/PlayFab.js  around line 96

targets/go/source/PlayFabHttp.go.ejs
```diff
@@ -0,0 +1,165 @@
...
+
+	bodyBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	result := playFabResponse{}
+	err = json.Unmarshal(bodyBytes, &result)
+	if err != nil {
+		return nil, err
```
MarcoWilliamsPF: This should create structured data for the error when it fails. Go looks similar to the NodeSDK.. you might want to check that out here: https://github.com/PlayFab/NodeSDK/blob/master/PlayFabSdk/Scripts/PlayFab/PlayFab.js around line 96

targets/go/source/PlayFabHttp.go.ejs
```diff
@@ -0,0 +1,165 @@
...
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	bodyBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
```
MarcoWilliamsPF: This should create structured data for the error when it fails. Go looks similar to the NodeSDK.. you might want to check that out here: https://github.com/PlayFab/NodeSDK/blob/master/PlayFabSdk/Scripts/PlayFab/PlayFab.js around line 96

targets/go/source/PlayFabHttp.go.ejs
```diff
@@ -0,0 +1,165 @@
...
+	req.Header.Add("X-ReportErrorAsSuccess", "true")
+	req.Header.Add("X-PlayFabSDK", p.VersionString)
+
+	if authKey != "" && authValue != "" {
+		req.Header.Add(authKey, authValue)
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, err
```
MarcoWilliamsPF: This should create structured data for the error when it fails. Go looks similar to the NodeSDK.. you might want to check that out here: https://github.com/PlayFab/NodeSDK/blob/master/PlayFabSdk/Scripts/PlayFab/PlayFab.js around line 96

targets/go/source/PlayFabHttp.go.ejs
```diff
@@ -0,0 +1,165 @@
...
+		req, err = http.NewRequest("POST", p.GetFullURL(apiPath), &buf)
+		if err != nil {
+			return nil, err
+		}
+		req.Header.Add("Content-Encoding", "GZIP")
+		req.Header.Add("Accept-Encoding", "GZIP")
+	} else {
+		req, err = http.NewRequest("POST", p.GetFullURL(apiPath), bytes.NewBuffer(b))
+		if err != nil {
+			return nil, err
```
MarcoWilliamsPF: This should create structured data for the error when it fails. Go looks similar to the NodeSDK.. you might want to check that out here: https://github.com/PlayFab/NodeSDK/blob/master/PlayFabSdk/Scripts/PlayFab/PlayFab.js around line 96

targets/go/source/PlayFabHttp.go.ejs
```diff
@@ -0,0 +1,165 @@
...
+		client.Transport = &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		}
+	}
+
+	var req *http.Request
+	var err error
+
+	// compression enabled, so gzip POST data
+	if p.EnableCompression {
```
MarcoWilliamsPF: This should create structured data for the error when it fails. Go looks similar to the NodeSDK.. you might want to check that out here: https://github.com/PlayFab/NodeSDK/blob/master/PlayFabSdk/Scripts/PlayFab/PlayFab.js around line 96  ... you have multiple points of failure,  each should return a structured data.. suggest making a new model, and return it with 503 (or some other code)

targets/go/source/PlayFabHttp.go.ejs
```diff
@@ -0,0 +1,165 @@
...
+// Request executes the actual HTTP call
+func Request(p *Settings, b []byte, apiPath string, authKey string, authValue string) (map[string]interface{}, error) {
+
+	client := &http.Client{
+		Timeout: time.Second * 30, // 30 seconds timeout should be enough
+	}
+
+	if p.DisableSsl {
+		client.Transport = &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
MarcoWilliamsPF: This is a potential security risk,  I pinged you on slack about it.. Is there a better solution in GO where we can validate the certificate?

dgkanatsios: Go does TLS validation by default on HTTP requests.
I copied this from the PHP SDK, I’d be more than happy to remove it if you think it’s unnecessary.

dgkanatsios:
About e2e testing, is there anywhere a testTitleData.json I should use?

dgkanatsios:
So, just did another push, to summarize and ask for help:
1. Pushed e2e tests here: https://github.com/dgkanatsios/playfabsdk-go/blob/master/e2e/main.go
2. when I'm GZIPing, this is the error I see: `invalid character '\x1f' looking for beginning of value`. It might be that my GZIP code is wrong, though :(
3. I still can't get a value for sdkGlobals.sdkVersion, probably [this](https://github.com/PlayFab/SDKGenerator/blob/master/generate.ts#L429) is the piece of code that generates that, probably `SdkManualNotes.json` needs to be modified in the Jenkins server?
4. In order for the e2e test to run, Go files should be copied to a proper directory in the [GOPATH](https://golang.org/doc/code.html#GOPATH) of the Jenkins server (or wherever the tests run). It should be like "C:\users\<username>\Go\src\github.com\PlayFab\playfabsdk-go". Then, `go e2e\main.go` would need to run. Also an external dependency would need to be installed via `go get github.com/mitchellh/mapstructure`.

Happy to help in all of the above!

dgkanatsios:
@MarcoWilliamsPF just fixed the GZIP issue, it turns out that only Content-Encoding header is needed, if you add Accept-Encoding it will not work (which makes thing that there might exist a bug [here](https://github.com/PlayFab/PhpSdk/blob/bc3639e775007375363284d0128888abbaaeb053/PlayFabSDK/PlayFabHttp.php#L76)). 
Let me know what else is needed, thanks!

dgkanatsios:
Just switched all methods to return plain `error` interface, as recommended [here](https://golang.org/doc/faq#nil_error)

MarcoWilliamsPF:
I don't think we will get to this until beginning of April due to our workload.  We do have this in our tasking to evaluate.  Thanks!

bketelsen:
Volunteering to help get the Go client landed.  How can I help?

dgkanatsios:
Closing temporarily since I'm testing GitHub Actions on my repo. Sorry for the Jenkins noise!

PR #395, opened 2019-02-08
(for Milestones 33-34) DONT DELETE (will be needed in 20Q1) PROTOTYPE: Implementation of lightweight events in C# SDK

Vicodex:
PROTOTYPE WORK:

Lightweight events are custom high-volume, high-throughput events sent to Microsoft's OneDS (One Collector) service account associated with PlayFab user's entity ID. It is expected that these events can be quickly emitted at a high frequency from different threads of a customer app (i.e. non-blocking thread-safe emit event calls), batched up according to specific settings, and sent out (whether when a batch is full or an allowed timeframe to assemble a batch ran out) to the OneDS service from a background thread.
Sending events to OneDS service requires specific authentication info (tokens, IDs) that must be obtained first in runtime by calling PlayFab's "/Event/GetTelemetryIngestionConfig" API.

The names of classes are specifically chosen to be consistent with C++ implementation.

The code in this PR includes the following major components:
1) Classes in <SDK>/source/external/aria is a bare minimum custom dependency carefully taken from much larger Aria SDK to support Bond serialization and Common Schema 3.0 composition required by OneDS service
2) OneDSSysHttpPlugin is an HTTP plugin specialized to communicate with OneDS server
3) OneDSEventsAPI is a OneDS-equivalent of PlayFabEventsAPI: it provides an API method to send an already assembled batch of events to OneDS server. In addition, this class includes an API method to get OneDS authentication info from PlayFab server (GetTelemetryIngestionConfigAsync)
4) OneDSEventsModels contains data models required for OneDSEventsAPI
5) PlayFabEvent is a class that represents a custom event (such as a lightweight event, for example. There can be other types of custom events in the future)
6) PlayFabEventAPI is an API that allows to emit custom events (such as lightweight events)
7) PlayFabEventRouter is a class that routes custom events to different event processing pipelines, depending on the type of an event and other conditions
8) Classes in <SDK>/source/Pipeline contain the implementation of an event processing pipeline that is largely based on the pipeline implementation from the Internal C# SDK (used in PlayFab Thunderhead services). We would like to keep these implementations as close as possible (and have only one, eventually). A pipeline consists of several buffers (event buffer, batch buffer) and stages (batching stage, sending stage). Each stage reads an item from one buffer, processes it, and puts it to another buffer. The goal is to eliminate performance bottlenecks associated with sending batches over HTTP and, to some degree, batching. The pipeline code was reasonably tested in Thunderhead environment.
9) OneDSEventPipelineSettings contains parameters that regulate main technical characteristics of an event pipeline (size of buffers, size of event batches, max wait interval before a batch needs to be sent out, etc).
10) <SDK>/source/logger contains classes that provide a logger interface (and the default debug logger) used in event pipelines
11) Test for OneDSEventsAPI that sends an already prepared batch of events to OneDS server
12) Test for PlayFabEventAPI that emits many lightweight events and uses a OneDS event pipeline to batch them in background, and send those batches to OneDS server


targets/csharp/source/source/Logger/DebugLogger.cs
```diff
@@ -0,0 +1,60 @@
+using System;
+
+namespace PlayFab.Logger
```
pgilmorepf: Can you describe your goals in creating this please?

targets/csharp/source/source/OneDSEventsAPI.cs
```diff
@@ -0,0 +1,182 @@
...
+                    Error = new PlayFabError
+                    {
+                        Error = PlayFabErrorCode.ContentNotFound,
+                        ErrorMessage = "No events found in request. Please make sure to provide at least one event in the batch.",
+                        HttpStatus = "OneDSError"
+                    },
+                    CustomData = customData
+                };
+
+                PlayFabSettings.GlobalErrorHandler?.Invoke(apiMethodResult.Error);
```
pgilmorepf: Mixed feelings about expanding the usage of this handler.
It is named "global" I guess...

targets/csharp/source/source/OneDSEventsAPI.cs
```diff
@@ -0,0 +1,182 @@
...
+
+            var httpResult = await transport.DoPost(serializedBatch, headers);
+            if (httpResult is PlayFabError)
+            {
+                var error = (PlayFabError)httpResult;
+                PlayFabSettings.GlobalErrorHandler?.Invoke(error);
+                return new PlayFabResult<WriteEventsResponse> { Error = error, CustomData = customData };
+            }
+
+            return new PlayFabResult<WriteEventsResponse> { Result = new WriteEventsResponse(), CustomData = customData };
```
pgilmorepf: It seems odd to me to construct a new WriteEventsResponse here, and potentially ignore any good information that may have been returned by the successful httpResult.

targets/csharp/source/source/Pipeline/EventSendingStage.cs
```diff
@@ -0,0 +1,120 @@
...
+                    logger.Error($"Exception in OnNextInputItem {e.Source} with message: {e}.");
+                }
+            }
+
+            Task<PlayFabResult<WriteEventsResponse>> apiTask = oneDSEventsApi.WriteTelemetryEventsAsync(writeEventsRequest);
+            try
+            {
+                apiTask.Wait(cts.Token);
+                this.FulfillPromises(batch.Events, apiTask.Result);
+                //this.OnRaiseStageNotificationEvent(new BatchSendOutcomeEventArgs(batch, writeEventsTask.Result));
```
pgilmorepf: Fix or delete, don't leave commented code.

targets/csharp/source/source/external/aria/Utils.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    class Utils
+    {
+        /// <summary>
+        /// Calculates milliseconds since 1970
+        /// </summary>
+        /// <returns>Milliseconds since 1970</returns>
+        public static long TimestampNow()
+        {
+            return (DateTime.UtcNow.Ticks - (new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc).Ticks)) / 10000;
```
pgilmorepf: Even as their code, this should be fixed as I describe above.

targets/csharp/source/source/external/aria/Utils.cs
```diff
@@ -0,0 +1,44 @@
...
+            try
+            {
+                var endIndex = tenant.IndexOf("-", StringComparison.Ordinal);
+                var tenantId = string.Empty;
+                if (endIndex > 0)
+                    return tenantId = tenant.Substring(0, endIndex);
+            }
+            catch
+            {
+                // TODO: log this ("Failed to convert tenantId")
```
pgilmorepf: Is this Aria todo problematic for us?

Vicodex: Technically this is an external dependency code. I would like to keep it with minimal changes so that it's easier to update in the future, unless something is really broken.

pgilmorepf:
Closing this for later when Vicodex returns.

pgilmorepf:
retest

pgilmorepf:
We're redirecting "lightweight events" to an alternate endpoint, and much of this PR is no longer necessary.

PR #396, opened 2019-02-08
Melihceyhan/multi instant support for servers

melihceyhan:
Milestone #31

### DIFF
--- a/targets/csharp/make.js
+++ b/targets/csharp/make.js
@@ -289,7 +289,7 @@ function getRequestActions(tabbing, apiCall) {
     if (apiCall.auth === "SecretKey")
         return tabbing + "if (PlayFabSettings.DeveloperSecretKey == null) throw new PlayFabException(PlayFabExceptionCode.DeveloperKeyNotSet, \"Must have PlayFabSettings.DeveloperSecretKey set to call this method\");\n";
     if (apiCall.auth === "SecretKeyInstance")
-        return tabbing + "var developerSecretKey = request.AuthenticationContext?.DeveloperSecretKey ?? (authenticationContext?.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey);\n            if (developerSecretKey == null) throw new PlayFabException(PlayFabExceptionCode.DeveloperKeyNotSet, \"Must have PlayFabSettings.DeveloperSecretKey set to call this method\");\n";
+        return tabbing + "var developerSecretKey = request.AuthenticationContext?.DeveloperSecretKey ?? (authenticationContext?.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey);\n            if (developerSecretKey == null) throw new PlayFabException(PlayFabExceptionCode.DeveloperKeyNotSet, \"DeveloperSecretKey is not found in Request, Server Instance or PlayFabSettings\");\n";
     if (apiCall.url === "/Authentication/GetEntityToken")
         return tabbing + "string authKey = null, authValue = null;\n"
             + tabbing + "if (PlayFabSettings.ClientSessionTicket != null) { authKey = \"X-Authorization\"; authValue = PlayFabSettings.ClientSessionTicket; }\n"
--- a/targets/csharp/make.js
+++ b/targets/csharp/make.js
@@ -15,7 +15,7 @@ exports.makeClientAPI2 = function (apis, sourceDir, apiOutputDir) {
     for (var i = 0; i < apis.length; i++)
         makeApi(apis[i], sourceDir, apiOutputDir);
     generateSimpleFiles(apis, sourceDir, apiOutputDir);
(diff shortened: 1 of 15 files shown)


targets/csharp/make.js
```diff
@@ -41,7 +49,16 @@ exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
     for (var i = 0; i < apis.length; i++)
         makeApi(apis[i], sourceDir, apiOutputDir);
     generateSimpleFiles(apis, sourceDir, apiOutputDir);
-    generateProject(apis, sourceDir, apiOutputDir, "All", ";ENABLE_PLAYFABADMIN_API;ENABLE_PLAYFABSERVER_API");
+    generateProject(apis, sourceDir, apiOutputDir, "All", ";ENABLE_PLAYFABADMIN_API;ENABLE_PLAYFABSERVER_API;ENABLE_PLAYFABCLIENT_API");
+}
+
+exports.makeCombinedInstanceAPI = function (apis, sourceDir, apiOutputDir) {
```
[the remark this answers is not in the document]
melihceyhan: Sorry for confusion, I got a wrong workflow logic in my mind. Now It generates for Instancable ones for all API classes.

Vicodex:
Please address the comments from Paul, and i think it should be good to go

targets/csharp/make.js
```diff
@@ -263,32 +253,32 @@ function getPropertyCsType(property, datatype, needOptional) {
...
     if (apiCall.url === "/Authentication/GetEntityToken")
         return "authKey, authValue";
     if (apiCall.auth === "EntityToken")
         return "\"X-EntityToken\", PlayFabSettings.EntityToken";
-    if (apiCall.auth === "SecretKey")
+    if (apiCall.auth === "SecretKey" && isApiInstance === false)
         return "\"X-SecretKey\", PlayFabSettings.DeveloperSecretKey";
-    if (apiCall.auth === "SecretKeyInstance")
+    if (apiCall.auth === "SecretKey" && isApiInstance === true)
         return "\"X-SecretKey\", developerSecretKey";
```
pgilmorepf: Let's check if developerSecretKey is string.IsNullOrEmpty(), and if it is, then fall back on PlayFabSettings.DeveloperSecretKey.
@Vicodex do you agree?

melihceyhan: It already generates something like you said. Here is a is a bunch of code which is generated from this part:  
image

pgilmorepf:
Since this is a big merge, please send me a final comment saying that this is ready to land.

melihceyhan:
I run both developer and unit tests with it. In my opinion we are ready

melihceyhan:
should I do anything more?

targets/csharp/make.js
```diff
@@ -223,28 +253,33 @@ function getPropertyCsType(property, datatype, needOptional) {
...
     if (apiCall.auth === "SessionTicket")
         return tabbing + "if (PlayFabSettings.ClientSessionTicket == null) throw new PlayFabException(PlayFabExceptionCode.NotLoggedIn, \"Must be logged in to call this method\");\n";
-    if (apiCall.auth === "SecretKey")
+    if (apiCall.auth === "SecretKey" && isApiInstance === false)
         return tabbing + "if (PlayFabSettings.DeveloperSecretKey == null) throw new PlayFabException(PlayFabExceptionCode.DeveloperKeyNotSet, \"Must have PlayFabSettings.DeveloperSecretKey set to call this method\");\n";
+    if (apiCall.auth === "SecretKey" && isApiInstance === true)
+        return tabbing + "var developerSecretKey = request.AuthenticationContext?.DeveloperSecretKey ?? (authenticationContext?.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey);\n            if (developerSecretKey == null) throw new PlayFabException(PlayFabExceptionCode.DeveloperKeyNotSet, \"DeveloperSecretKey is not found in Request, Server Instance or PlayFabSettings\");\n";
```
M-Pixel: I get a build failure on release config because this is not protected with `#if ENABLE_PLAYFABSERVER_API || ENABLE_PLAYFABADMIN_API`

pgilmorepf: Good catch. Please submit a fix.

melihceyhan: image
Here is a bunch of code which is taken from multi-user-instance-support branch. We do this macro sets.  When we merge it will be ok. Should I make an other submit for fix

melihceyhan: this part of code should only work with Instanceable apis and  developer secret keys should be there. But To prevent an error which can be occur later I will add this flags.

 @pgilmorepf is there a problem if I add them in the multi-user-instance-support branch which is reviewing by  you 

@M-Pixel  on the other hand both release and debug configuration builds without any error on my side so which branch and solution(client,server,admin) you are working on ?  Can you send me screenshot of error to make understand where error occured

PR #401, opened 2019-02-12
Milestone 26b: Package testTitleData.json in UE4 for easy inclusion in mobile builds

M-Pixel:
If environment variable PF_TEST_TITLE_DATA_JSON is defined, copy it to a Content/JSON within every UE4 ExampleProject.  Configure UE4 to include Content/JSON in asset packaging.  This allows access to the
file from a mobile build, without requiring the developer to manually copy testTitleData.json onto the mobile device, which is particularly difficult on iOS.  This is also potentially useful on desktop, but moreso on target platforms where apps are tightly sandboxed.

This utilizes UE4's file library that unifies packaged-asset-retrieval and raw-filesystem-retrieval to the same API.  If the file is not available at generation time, and thus does not become available in the correct packaged location, the environment variable still kicks in at runtime and can retrieve directly from the filesystem.

### DIFF
--- a/targets/UnrealMarketplacePlugin/examplesource/ExampleProject/Plugins/PlayFab/Source/PlayFab/Classes/PfTestActor.h.ejs
+++ b/targets/UnrealMarketplacePlugin/examplesource/ExampleProject/Plugins/PlayFab/Source/PlayFab/Classes/PfTestActor.h.ejs
@@ -116,7 +116,7 @@ public:
         FString _outputSummary;
     // A bunch of constants: load these from testTitleData.json
     UPROPERTY()
-        FString TEST_TITLE_DATA_LOC = "testTitleData.json";
+        FString TEST_TITLE_DATA_LOC = FPaths::ProjectContentDir() + TEXT("JSON/testTitleData.json");
     UPROPERTY()
         FString userEmail;
     UPROPERTY()
--- a/targets/UnrealMarketplacePlugin/make.js
+++ b/targets/UnrealMarketplacePlugin/make.js
@@ -1,6 +1,7 @@
 var path = require("path");
 var cppMakeJsPath = require("./makeCpp.js");
 var bpMakeJsPath = require("./makebp.js");
+var fs = require("fs");
 
 // Making resharper less noisy - These are defined in Generate.js
 if (typeof (generateApiSummaryLines) === "undefined") generateApiSummaryLines = function () { };
@@ -16,6 +17,7 @@ var copyright =
 exports.makeCombinedAPI = function (apis, sourceDir, baseApiOutputDir) {
     // The list of current supported UE versions - Intended to be the latest 3
     var ueTargetVersions = ["4.18", "4.19", "4.20", "4.21"];
+    var testTitleJsonPath = process.env.PF_TEST_TITLE_DATA_JSON;
 
     for (var v = 0; v < ueTargetVersions.length; v++) {
         var ueTargetVersion = ueTargetVersions[v];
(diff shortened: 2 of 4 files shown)


targets/UnrealMarketplacePlugin/make.js
```diff
@@ -53,9 +55,23 @@ exports.makeCombinedAPI = function (apis, sourceDir, baseApiOutputDir) {
...
         templatizeTree(locals, path.resolve(apiOutputDir, "PlayFabPlugin"), path.resolve(apiOutputDir, "ExampleProject/Plugins"));
+
+        // Create a Content folder that can hold testTitleJSON, so that UE4 can handle putting it onto mobile devices in particular
+        // This works in conjunction with a line in DefaultGame.ini
+        var jsonFolderPath = path.resolve(apiOutputDir, "ExampleProject", "Content", "JSON");
+        fs.mkdir(jsonFolderPath, { recursive: true }, handleError);
+
+        // Don't require PF_TEST_TITLE_DATA_JSON during SDK generation - some devs don't need it, or might just want to put it into the folder manually
+        if (testTitleJsonPath !== undefined) {
+            fs.copyFile(testTitleJsonPath, path.join(jsonFolderPath, "testTitleData.json"), handleError);
```
pgilmorepf: Definitely don't do this.
We don't want to accidentally leak any secure testTitleData files.

targets/UnrealMarketplacePlugin/examplesource/ExampleProject/Plugins/PlayFab/Source/PlayFab/Classes/PfTestActor.h.ejs
```diff
@@ -116,7 +116,7 @@ public:
         FString _outputSummary;
     // A bunch of constants: load these from testTitleData.json
     UPROPERTY()
-        FString TEST_TITLE_DATA_LOC = "testTitleData.json";
+        FString TEST_TITLE_DATA_LOC = FPaths::ProjectContentDir() + TEXT("JSON/testTitleData.json");
```
pgilmorepf: Before we do this, we need to make sure that the .gitIgnore for all Unreal repos contains exclusions for testTitleData files, so they never get submitted.

Add this to .gitIgnore, in every UE repo:

```
**/*TestTitleData.json*
**/*testTitleData.json*
```

We need to land and approve those first, before we can submit this change.

targets/UnrealMarketplacePlugin/make.js
pgilmorepf: Even after the .gitIgnore changes, we can't do this as part of SdkGenerator.
I'm not sure if this is ever safe to do in an automated way.
I will think about it, but we're definitely not doing this here.

targets/UnrealMarketplacePlugin/examplesource/ExampleProject/Plugins/PlayFab/Source/PlayFab/Classes/PfTestActor.h.ejs
Vicodex: Those two lines in .gitignore can possibly be done in one:
**/*[Tt]estTitleData.json*

---
In reply to: [256133097](https://github.com/PlayFab/SDKGenerator/pull/401#discussion_r256133097) [](ancestors = 256133097)

targets/UnrealMarketplacePlugin/make.js
M-Pixel: I think the one danger here is if a developer (who is set up with testTitleData.json) doesn't know that this functionality exists, and then decides to copy-paste their ExampleProject to somewhere public.  That could be solved by moving this functionality into a dedicated command script - if you have to click "copy testTitleData to test runner projects", then at that point I think it's equal risk to not having this code at all (i.e. requiring developers to manually copy/paste the file into every test project), but still less tedious.

Vicodex:
Decided to postpone this and not to require for this deliverable to be a blocker.

PR #403, opened 2019-02-14
Milestone 28: Added multi-user authenticationContext support in JavaScript

malcolmhoward:
There's a refactor-in-progress of PlayFab._internalSettings.GetAuthValue called PlayFab._internalSettings.GetAuthInfo.  It currently breaks on GetPlayerProfile in the unit tests, but all other previous tests before that one pass.  I'm submitting this pull request using GetAuthValue for now, but I wanted to make you aware of the unused function before you come across it in the code.  The idea behind the refactor is to generalize the logic in the various API methods, so that a method such as GetPlayerProfile, for example, can go from

```
GetPlayerProfile: function (request, callback, customData, extraHeaders) {
        var authKey = "X-Authorization"; var authValue = PlayFab._internalSettings.GetAuthValue(request, authKey);
        if (!authValue) throw PlayFab._internalSettings.errorLoggedIn;
        PlayFab._internalSettings.ExecuteRequest(PlayFab._internalSettings.GetServerUrl() + "/Client/GetPlayerProfile", request, authKey, authValue, callback, customData, extraHeaders);
    },
```

to 

```
GetPlayerProfile: function (request, callback, customData, extraHeaders) {
        var authKey = "X-Authorization"; var authInfo = PlayFab._internalSettings.GetAuthInfo(request, authKey);
        var authValue = authInfo.authValue  ; var authError = authInfo.authError;
        if (!authValue) throw authError ;
        PlayFab._internalSettings.ExecuteRequest(PlayFab._internalSettings.GetServerUrl() + "/Client/GetPlayerProfile", request, authKey, authValue, callback, customData, extraHeaders);
    },
```

I think this change would make it easier to abstract the logic for the bulk of the API methods into one or two helper functions, which should significantly reduce the number of lines of code in the various API files.  Let me know what you think!

### DIFF
--- a/targets/javascript/make.js
+++ b/targets/javascript/make.js
@@ -76,11 +76,20 @@ function getRequestActions(tabbing, apiCall) {
     if (apiCall.result === "LoginResult" || apiCall.request === "RegisterPlayFabUserRequest")
         return tabbing + "request.TitleId = PlayFab.settings.titleId ? PlayFab.settings.titleId : request.TitleId; if (!request.TitleId) throw PlayFab._internalSettings.errorTitleId;\n";
     if (apiCall.auth === "EntityToken")
-        return tabbing + "if (!PlayFab._internalSettings.entityToken) throw PlayFab._internalSettings.errorEntityToken;\n";
+        return tabbing + "var authKey = \"X-EntityToken\"; var authValue = PlayFab._internalSettings.GetAuthValue(request, authKey);\n"
+            + tabbing + "if (!authValue) throw PlayFab._internalSettings.errorEntityToken;\n";
     if (apiCall.auth === "SessionTicket")
-        return tabbing + "if (!PlayFab._internalSettings.sessionTicket) throw PlayFab._internalSettings.errorLoggedIn;\n";
+        return tabbing + "var authKey = \"X-Authorization\"; var authValue = PlayFab._internalSettings.GetAuthValue(request, authKey);\n"
+            + tabbing + "if (!authValue) throw PlayFab._internalSettings.errorLoggedIn;\n";
     if (apiCall.auth === "SecretKey")
-        return tabbing + "if (!PlayFab.settings.developerSecretKey) throw PlayFab._internalSettings.errorSecretKey;\n";
+        return tabbing + "var authKey = \"X-SecretKey\"; var authValue = PlayFab._internalSettings.GetAuthValue(request, authKey);\n"
+            + tabbing + "if (!authValue) throw PlayFab._internalSettings.errorSecretKey;\n";
+//    TODO: Replace the 3 apiCall.auth if-blocks above with the generic solution below once PlayFab._internalSettings.GetAuthInfo() is ready
+//    if (["EntityToken", "SessionTicket", "SecretKey"].includes(apiCall.auth))
+//        return tabbing + "var authKey = \"X-EntityToken\"; var authInfo = PlayFab._internalSettings.GetAuthInfo(request, authKey);\n"
+//        + tabbing + "var authValue = authInfo.authValue; var authError = authInfo.authError;\n"
+//        + tabbing + "if (!authValue) throw authError;\n";
+
     return "";
 }
 
@@ -101,7 +110,7 @@ function getResultActions(tabbing, apiCall) {
             + tabbing + "   var playFabId = result.data.PlayFabId;\n"
             + tabbing + "\n"
             + tabbing + "   if(PlayFab._internalSettings.authenticationContext.hasOwnProperty(playFabId) == false) {\n"
-            + tabbing + "       var playFabId = result.data.PlayFabId;\n"
+            + tabbing + "       PlayFab._internalSettings.authenticationContext[playFabId] = {};"
             + tabbing + "   }\n"
             + tabbing + "\n"
             + tabbing + "   if(result.data.SessionTicket != null) {\n"
@@ -133,14 +142,21 @@ function getUrl(apiCall) {
 }
 
 function getAuthParams(apiCall) {
-    if (apiCall.url === "/Authentication/GetEntityToken")
+    var authConditionArray = [
+        apiCall.url === "/Authentication/GetEntityToken",
+        apiCall.auth === "EntityToken",
+        apiCall.auth === "SecretKey",
+        apiCall.auth === "SessionTicket"
+    ];
+
+    function checkAuthCondition(evaluatedExpression) {
+      return evaluatedExpression === true;
+    }
+
+    // EntityToken, SecretKey, & SessionTicket auth conditions now support authKey and authValue in getRequestActions()
+    // The some() method tests that at least one element in the array passes the test implemented by the provided function
+    if (authConditionArray.some(checkAuthCondition))
         return "authKey, authValue";
-    if (apiCall.auth === "EntityToken")
-        return "\"X-EntityToken\", PlayFab._internalSettings.GetAuthValue(request, \"X-EntityToken\")";
-    if (apiCall.auth === "SecretKey")
-        return "\"X-SecretKey\", PlayFab._internalSettings.GetAuthValue(request, \"X-SecretKey\")";
-    if (apiCall.auth === "SessionTicket")
-        return "\"X-Authorization\", PlayFab._internalSettings.GetAuthValue(request, \"X-Authorization\")";
     return "null, null";
 }
 
--- a/targets/javascript/make.js
+++ b/targets/javascript/make.js
@@ -23,6 +23,7 @@ exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
         getRequestActions: getRequestActions,
         getResultActions: getResultActions,
         getUrl: getUrl,
+        getPartialUrl: getPartialUrl,
         hasResultActions: hasResultActions,
         sdkVersion: sdkGlobals.sdkVersion,
         sourceDir: sourceDir,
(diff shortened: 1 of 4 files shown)


targets/javascript/make.js
```diff
@@ -76,11 +76,20 @@ function getRequestActions(tabbing, apiCall) {
...
-        return tabbing + "if (!PlayFab._internalSettings.sessionTicket) throw PlayFab._internalSettings.errorLoggedIn;\n";
+        return tabbing + "var authKey = \"X-Authorization\"; var authValue = PlayFab._internalSettings.GetAuthValue(request, authKey);\n"
+            + tabbing + "if (!authValue) throw PlayFab._internalSettings.errorLoggedIn;\n";
     if (apiCall.auth === "SecretKey")
-        return tabbing + "if (!PlayFab.settings.developerSecretKey) throw PlayFab._internalSettings.errorSecretKey;\n";
+        return tabbing + "var authKey = \"X-SecretKey\"; var authValue = PlayFab._internalSettings.GetAuthValue(request, authKey);\n"
+            + tabbing + "if (!authValue) throw PlayFab._internalSettings.errorSecretKey;\n";
+//    TODO: Replace the 3 apiCall.auth if-blocks above with the generic solution below once PlayFab._internalSettings.GetAuthInfo() is ready
+//    if (["EntityToken", "SessionTicket", "SecretKey"].includes(apiCall.auth))
+//        return tabbing + "var authKey = \"X-EntityToken\"; var authInfo = PlayFab._internalSettings.GetAuthInfo(request, authKey);\n"
```
malcolmhoward: Ah, this hard-coded authKey could be the reason why PlayFab._internalSettingsGetAuthInfo() fails for the GetPlayerProfile unit test.  That definitely shouldn't be hard-coded.  Hmm...

targets/javascript/make.js
```diff
@@ -97,12 +106,21 @@ function hasResultActions(apiCall) {
...
-            + tabbing + "       if (result.data.EntityToken != null) {\n"
-            + tabbing + "           PlayFab._internalSettings.entityToken = result.data.EntityToken.EntityToken;\n"
-            + tabbing + "       }\n"
+            + tabbing + "   // Capture the playFabId, and make sure the authenticationContext exists with the ID in question\n"
+            + tabbing + "   var playFabId = result.data.PlayFabId;\n"
+            + tabbing + "\n"
+            + tabbing + "   if(PlayFab._internalSettings.authenticationContext.hasOwnProperty(playFabId) == false) {\n"
+            + tabbing + "       PlayFab._internalSettings.authenticationContext[playFabId] = {};"
+            + tabbing + "   }\n"
+            + tabbing + "\n"
```
pgilmorepf: please don't add spaces to empty lines.
I would actually prefer if the previous line just ended in:
```
...}\n\n"
```
if you want to have a double-space there.

targets/javascript/make.js
```diff
@@ -97,12 +106,21 @@ function hasResultActions(apiCall) {
...
+            + tabbing + "   if(result.data.SessionTicket != null) {\n"
+            + tabbing + "       PlayFab._internalSettings.sessionTicket = result.data.SessionTicket;\n"
+            + tabbing + "       PlayFab._internalSettings.authenticationContext[playFabId]['sessionTicket'] = result.data.SessionTicket;\n"
```
pgilmorepf: Use .variableName when appropriate.
For example, when accessing variables on an authenticationContext, the syntax should be written as:
authenticationContext.sessionTicket
not
authenticationContext['sessionTicket']

targets/javascript/make.js
```diff
@@ -97,12 +106,21 @@ function hasResultActions(apiCall) {
...
-            + tabbing + "           PlayFab._internalSettings.sessionTicket = result.data.SessionTicket;\n"
-            + tabbing + "       }\n"
-            + tabbing + "       if (result.data.EntityToken != null) {\n"
-            + tabbing + "           PlayFab._internalSettings.entityToken = result.data.EntityToken.EntityToken;\n"
-            + tabbing + "       }\n"
+            + tabbing + "   // Capture the playFabId, and make sure the authenticationContext exists with the ID in question\n"
+            + tabbing + "   var playFabId = result.data.PlayFabId;\n"
+            + tabbing + "\n"
+            + tabbing + "   if(PlayFab._internalSettings.authenticationContext.hasOwnProperty(playFabId) == false) {\n"
+            + tabbing + "       PlayFab._internalSettings.authenticationContext[playFabId] = {};"
```
pgilmorepf: Ok, I figured out what I don't like about this. It took me a while.
It is not the responsibility of the PlayFab SDK to archive or index the authenticationContexts for every logged in player.

PlayFab._internalSettings.authenticationContext should not be a dictionary of all authenticationContext's ever.

PlayFabClientApi.LoginWithCustomId will return an authenticationContext as part of a new parameter.
If the customer keeps it, then great. If they don't, then it's just gone. We're not tracking them.

PlayFab._internalSettings.authenticationContext is just one instance of a single authenticationContext, which is the last person to use the static interface to log in. That's it. It's over-written when somebody else logs in.

The developer is responsible for calling login 3 times, and keeping track of three separate authenticationContext objects, and feeding them back into request objects.

targets/javascript/make.js
```diff
@@ -124,14 +142,21 @@ function getUrl(apiCall) {
...
+        apiCall.auth === "SessionTicket"
+    ];
+
+    function checkAuthCondition(evaluatedExpression) {
+      return evaluatedExpression === true;
+    }
+
+    // EntityToken, SecretKey, & SessionTicket auth conditions now support authKey and authValue in getRequestActions()
+    // The some() method tests that at least one element in the array passes the test implemented by the provided function
+    if (authConditionArray.some(checkAuthCondition))
```
pgilmorepf: I don't really understand this in the first pass, so I'm probably going to bring it up again in the next review.

targets/javascript/templates/PlayFab_Api.js.ejs
```diff
@@ -151,11 +151,77 @@ if(!PlayFab._internalSettings) {
...
+            if (PlayFab._internalSettings.authenticationContext[playFabId].hasOwnProperty(authAttr)) {
+                authValue = PlayFab._internalSettings.authenticationContext[playFabId][authAttr];
+            }
+        }
+    }
+    authAttr = authInfoMap[authKey]['authError'];
+    var authError = PlayFab._internalSettings[authAttr];
+
+    return {'authValue': authValue, 'authError': authError};
+};
```
pgilmorepf: I feel like you're replacing 3 lines of admittedly awkward template code with 50 lines of "clever" code. I don't see it as an improvement, and it doesn't structurally match the way we're doing it in the other SDKs.

targets/javascript/templates/PlayFab_Api.js.ejs
```diff
@@ -176,6 +242,7 @@ PlayFab.<%- api.name %>Api = {
 <% } %>    ForgetAllCredentials: function () {
         PlayFab._internalSettings.sessionTicket = null;
         PlayFab._internalSettings.entityToken = null;
+        PlayFab._internalSettings.authenticationContext = {};
```
pgilmorepf: As discussed above, we should not be archiving all authentication contexts ever.

targets/javascript/testingFiles/PlayFabApiTest.ts
```diff
@@ -234,6 +235,96 @@ var PlayFabApiTests = {
         PlayFabClientSDK.LoginWithCustomID(loginRequest, PlayFabApiTests.CallbackWrapper("advertLoginCallback", advertLoginCallback, assert));
     },
 
+    /* CLIENT API
+     * Test that the login call sequence uses the AuthenticationContext when set
+     */
+    LoginWithAuthenticationContext: function (assert) {
```
pgilmorepf: I'm not going to review this test until after the above comments are resolved.

targets/javascript/testingFiles/index.html.ejs
```diff
@@ -9,7 +9,7 @@
     <div id="qunit"></div>
     <div id="qunit-fixture"></div>
     <script src="http://code.jquery.com/jquery-1.7.1.min.js"></script><!-- HTTP vs HTTPS here should match where this file is hosted (This file is not strictly meant to be hosted, but rather run locally) -->
-    <script src="http://code.jquery.com/qunit/qunit-1.19.0.js"></script>
+    <script src="http://code.jquery.com/qunit/qunit-2.0.1.js"></script>
```
pgilmorepf: Yes, I'm happy with this.

targets/javascript/make.js
Vicodex: Thanks Paul! Yes, that is a good description of the goal. We don't want to cache authentication contexts ourselves. It's up to customers to keep/maintain/store them if they want, but we do not keep these authentication objects in _internalSettings or anywhere else ourselves. 

---
In reply to: [257078270](https://github.com/PlayFab/SDKGenerator/pull/403#discussion_r257078270) [](ancestors = 257078270)

targets/javascript/make.js
```diff
@@ -97,18 +95,20 @@ function hasResultActions(apiCall) {
...
+            + tabbing + "    // Apply the updates for the AuthenticationContext returned to the client\n"
+            + tabbing + "    authenticationContext = PlayFab._internalSettings.UpdateAuthenticationContext(authenticationContext, result);\n"
             + tabbing + "    PlayFab.ClientApi._MultiStepClientLogin(result.data.SettingsForUser.NeedsAttribution);\n"
             + tabbing + "}";
     if (apiCall.result === "RegisterPlayFabUserResult")
         return tabbing + "if (result != null && result.data.SessionTicket != null) {\n"
-            + tabbing + "    PlayFab._internalSettings.sessionTicket = result.data.SessionTicket;\n"
-            + tabbing + "    PlayFab.ClientApi._MultiStepClientLogin(result.data.SettingsForUser.NeedsAttribution);\n"
+            + tabbing + "     PlayFab._internalSettings.sessionTicket = result.data.SessionTicket;\n"
+            + tabbing + "     PlayFab.ClientApi._MultiStepClientLogin(result.data.SettingsForUser.NeedsAttribution);\n"
```
pgilmorepf: you have mis-aligned these lines with that extra space.

pgilmorepf: you have mis-aligned these lines with that extra space.

targets/javascript/make.js
```diff
@@ -119,20 +119,26 @@ function getResultActions(tabbing, apiCall) {
     return "";
 }
 
+function getPartialUrl(apiCall) {
```
pgilmorepf: So, the proper term for this is "urlPath".
So let's call this ```getUrlPath(...)```

targets/javascript/templates/PlayFab_Api.js.ejs
```diff
@@ -152,6 +152,79 @@ if(!PlayFab._internalSettings) {
...
+            "X-SecretKey": {
+                authAttr: "developerSecretKey",
+                authError: "errorSecretKey"
+            }
+        },
+
+        GetAuthInfo: function (request, authKey) {
+            // Use the most-recently saved sessionTicket, unless one was provided in the request via the PlayFabId property
+            var authAttr = PlayFab._internalSettings.AuthInfoMap[authKey].authError;
+            var authError = PlayFab._internalSettings[authAttr];
```
pgilmorepf: These two look backwards...

targets/javascript/templates/PlayFab_Api.js.ejs
```diff
@@ -152,6 +152,79 @@ if(!PlayFab._internalSettings) {
...
+            authAttr = PlayFab._internalSettings.AuthInfoMap[authKey].authAttr;
```
pgilmorepf: And now you're re-assigning authAttr, so I think you should figure out the correct two lines out of these three lines. It seems like authError is probably being set wrong, and the first authAttr assignment should be deleted

targets/javascript/templates/PlayFab_Api.js.ejs
```diff
@@ -186,9 +259,11 @@ PlayFab.<%- api.name %>Api = {
             if (callback != null && typeof (callback) == "function")
```
pgilmorepf: I know this isn't your code, but can you change this to ===?

pgilmorepf:
(and the rename)

Vicodex:
Sorry I merged it prematurely..  BUT I reverted the merge, and restored this branch.

Please re-issue a fresh PR from this branch to master.

PR #404, opened 2019-02-14
Milestone 27: Configuration for generating multi user instance support

melihceyhan:
milestone #27

### DIFF
--- a/targets/csharp/make.js
+++ b/targets/csharp/make.js
@@ -63,7 +63,7 @@ function getBaseTypeSyntax(datatype) {
 
     if (datatype.sortKey)
         parents.push("IComparable<" + datatype.name + ">");
-    
+
     var output = "";
     if (parents.length > 0) {
         output = " : ";
@@ -276,38 +276,36 @@ function getRequestActions(tabbing, apiCall, isApiInstance = false) {
             + tabbing + "if (request.TitleId == null) throw new PlayFabException(PlayFabExceptionCode.TitleNotSet, \"Must be have PlayFabSettings.TitleId set to call this method\");\n";
     if (apiCall.auth === "EntityToken")
         return tabbing + "#if !DISABLE_PLAYFABENTITY_API\n"
-	        + tabbing + "    if ((request.AuthenticationContext?.EntityToken ?? PlayFabSettings.EntityToken) == null) throw new PlayFabException(PlayFabExceptionCode.EntityTokenNotSet, \"Must call GetEntityToken before calling this method\");\n"
-			+ tabbing + "#else\n"    
-			+ tabbing + "    if ((PlayFabSettings.EntityToken) == null) throw new PlayFabException(PlayFabExceptionCode.EntityTokenNotSet, \"Must call GetEntityToken before calling this method\");\n"
-			+ tabbing + "#endif\n";
-	if (apiCall.auth === "SessionTicket")
+            + tabbing + "    if ((request.AuthenticationContext?.EntityToken ?? PlayFabSettings.EntityToken) == null) throw new PlayFabException(PlayFabExceptionCode.EntityTokenNotSet, \"Must call GetEntityToken before calling this method\");\n"
+            + tabbing + "#else\n"
+            + tabbing + "    if ((PlayFabSettings.EntityToken) == null) throw new PlayFabException(PlayFabExceptionCode.EntityTokenNotSet, \"Must call GetEntityToken before calling this method\");\n"
+            + tabbing + "#endif\n";
+    if (apiCall.auth === "SessionTicket")
         return tabbing + "#if !DISABLE_PLAYFABCLIENT_API\n"
-	        + tabbing + "    var clientSessionTicket = request.AuthenticationContext?.ClientSessionTicket ?? PlayFabSettings.ClientSessionTicket;\n"
+            + tabbing + "    var clientSessionTicket = request.AuthenticationContext?.ClientSessionTicket ?? PlayFabSettings.ClientSessionTicket;\n"
             + tabbing + "    if (clientSessionTicket == null) throw new PlayFabException(PlayFabExceptionCode.NotLoggedIn, \"Must be logged in to call this method\");\n"
-			+ tabbing + "#else\n"    
-			+ tabbing + "    var clientSessionTicket = PlayFabSettings.ClientSessionTicket;\n"
-			+ tabbing + "#endif\n";
+            + tabbing + "#else\n"
+            + tabbing + "    var clientSessionTicket = PlayFabSettings.ClientSessionTicket;\n"
+            + tabbing + "#endif\n";
     if (apiCall.auth === "SecretKey" && isApiInstance === false)
         return tabbing + "if (PlayFabSettings.DeveloperSecretKey == null) throw new PlayFabException(PlayFabExceptionCode.DeveloperKeyNotSet, \"Must have PlayFabSettings.DeveloperSecretKey set to call this method\");\n";
     if (apiCall.auth === "SecretKey" && isApiInstance === true)
         return tabbing + "#if ENABLE_PLAYFABSERVER_API || ENABLE_PLAYFABADMIN_API\n"
-			+ tabbing + "    var developerSecretKey = request.AuthenticationContext?.DeveloperSecretKey ?? (authenticationContext?.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey);\n"
-			+ tabbing + "    if (developerSecretKey == null) throw new PlayFabException(PlayFabExceptionCode.DeveloperKeyNotSet, \"DeveloperSecretKey is not found in Request, Server Instance or PlayFabSettings\");\n"
-			+ tabbing + "#else\n"    
-			+ tabbing + "    var developerSecretKey = PlayFabSettings.DeveloperSecretKey;\n"
-			+ tabbing + "#endif\n";
+            + tabbing + "    var developerSecretKey = request.AuthenticationContext?.DeveloperSecretKey ?? (authenticationContext?.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey);\n"
+            + tabbing + "    if (developerSecretKey == null) throw new PlayFabException(PlayFabExceptionCode.DeveloperKeyNotSet, \"DeveloperSecretKey is not found in Request, Server Instance or PlayFabSettings\");\n"
+            + tabbing + "#else\n"
+            + tabbing + "    var developerSecretKey = PlayFabSettings.DeveloperSecretKey;\n"
+            + tabbing + "#endif\n";
     if (apiCall.url === "/Authentication/GetEntityToken")
-        return tabbing + "string authKey = null, authValue = null;\n" 
+        return tabbing + "string authKey = null, authValue = null;\n"
             + tabbing + "#if !DISABLE_PLAYFABCLIENT_API\n"
             + tabbing + "    var clientSessionTicket = request.AuthenticationContext?.ClientSessionTicket ?? PlayFabSettings.ClientSessionTicket;\n"
             + tabbing + "    if (clientSessionTicket != null) { authKey = \"X-Authorization\"; authValue = clientSessionTicket; }\n"
-            + tabbing + "#endif\n"
-            + tabbing + "\n"
+            + tabbing + "#endif\n\n"
             + tabbing + "#if ENABLE_PLAYFABSERVER_API || ENABLE_PLAYFABADMIN_API\n"
             + tabbing + "    var developerSecretKey = request.AuthenticationContext?.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey;\n"
             + tabbing + "    if (developerSecretKey != null) { authKey = \"X-SecretKey\"; authValue = developerSecretKey; }\n"
-            + tabbing + "#endif\n"
-            + tabbing + "\n"
+            + tabbing + "#endif\n\n"
             + tabbing + "#if !DISABLE_PLAYFABENTITY_API\n"
             + tabbing + "    var entityToken = request.AuthenticationContext?.EntityToken ?? PlayFabSettings.EntityToken;\n"
             + tabbing + "    if (entityToken != null) { authKey = \"X-EntityToken\"; authValue = entityToken; }\n"
--- a/targets/csharp/make.js
+++ b/targets/csharp/make.js
@@ -52,15 +52,18 @@ exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
 function getBaseTypeSyntax(datatype) {
     var parents = [];
 
+    //begin classes - only 1 possible
     if (datatype.className.toLowerCase().endsWith("request"))
         parents.push("PlayFabRequestCommon");
-    if (datatype.className.toLowerCase() === "loginresult")
+    else if (datatype.className.toLowerCase() === "loginresult")
         parents.push("PlayFabLoginResultCommon");
     else if (datatype.className.toLowerCase().endsWith("response") || datatype.className.toLowerCase().endsWith("result"))
         parents.push("PlayFabResultCommon");
+    //end classes - only 1 
+
     if (datatype.sortKey)
         parents.push("IComparable<" + datatype.name + ">");
-
+    
     var output = "";
     if (parents.length > 0) {
         output = " : ";
@@ -274,7 +277,8 @@ function getRequestActions(tabbing, apiCall, isApiInstance = false) {
     if (apiCall.auth === "EntityToken")
         return tabbing + "if ((request.AuthenticationContext?.EntityToken ?? PlayFabSettings.EntityToken) == null) throw new PlayFabException(PlayFabExceptionCode.EntityTokenNotSet, \"Must call GetEntityToken before calling this method\");\n";
     if (apiCall.auth === "SessionTicket")
-        return tabbing + "var clientSessionTicket = request.AuthenticationContext?.ClientSessionTicket ?? PlayFabSettings.ClientSessionTicket;\n            if (clientSessionTicket == null) throw new PlayFabException(PlayFabExceptionCode.NotLoggedIn, \"Must be logged in to call this method\");\n";
+        return tabbing + "var clientSessionTicket = request.AuthenticationContext?.ClientSessionTicket ?? PlayFabSettings.ClientSessionTicket;\n"
+            + tabbing + "if (clientSessionTicket == null) throw new PlayFabException(PlayFabExceptionCode.NotLoggedIn, \"Must be logged in to call this method\");\n";
     if (apiCall.auth === "SecretKey" && isApiInstance === false)
         return tabbing + "if (PlayFabSettings.DeveloperSecretKey == null) throw new PlayFabException(PlayFabExceptionCode.DeveloperKeyNotSet, \"Must have PlayFabSettings.DeveloperSecretKey set to call this method\");\n";
     if (apiCall.auth === "SecretKey" && isApiInstance === true)
@@ -282,18 +286,18 @@ function getRequestActions(tabbing, apiCall, isApiInstance = false) {
     if (apiCall.url === "/Authentication/GetEntityToken")
         return tabbing + "string authKey = null, authValue = null;\n" 
             + tabbing + "#if !DISABLE_PLAYFABCLIENT_API\n"
-            + tabbing + "        var clientSessionTicket = request.AuthenticationContext ?.ClientSessionTicket ?? PlayFabSettings.ClientSessionTicket;\n"
-            + tabbing + "        if (clientSessionTicket != null) { authKey = \"X-Authorization\"; authValue = clientSessionTicket; }\n"
+            + tabbing + "    var clientSessionTicket = request.AuthenticationContext?.ClientSessionTicket ?? PlayFabSettings.ClientSessionTicket;\n"
+            + tabbing + "    if (clientSessionTicket != null) { authKey = \"X-Authorization\"; authValue = clientSessionTicket; }\n"
             + tabbing + "#endif\n"
             + tabbing + "\n"
             + tabbing + "#if ENABLE_PLAYFABSERVER_API || ENABLE_PLAYFABADMIN_API\n"
-            + tabbing + "        var developerSecretKey = request.AuthenticationContext ?.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey;\n"
-            + tabbing + "        if (developerSecretKey != null) { authKey = \"X-SecretKey\"; authValue = developerSecretKey; }\n"
+            + tabbing + "    var developerSecretKey = request.AuthenticationContext?.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey;\n"
+            + tabbing + "    if (developerSecretKey != null) { authKey = \"X-SecretKey\"; authValue = developerSecretKey; }\n"
             + tabbing + "#endif\n"
             + tabbing + "\n"
             + tabbing + "#if !DISABLE_PLAYFABENTITY_API\n"
-            + tabbing + "        var entityToken = request.AuthenticationContext ?.EntityToken ?? PlayFabSettings.EntityToken;\n"
-            + tabbing + "        if (entityToken != null) { authKey = \"X-EntityToken\"; authValue = entityToken; }\n"
+            + tabbing + "    var entityToken = request.AuthenticationContext?.EntityToken ?? PlayFabSettings.EntityToken;\n"
+            + tabbing + "    if (entityToken != null) { authKey = \"X-EntityToken\"; authValue = entityToken; }\n"
             + tabbing + "#endif\n";
     return "";
 }
--- a/targets/csharp/make.js
+++ b/targets/csharp/make.js
@@ -275,14 +275,27 @@ function getRequestActions(tabbing, apiCall, isApiInstance = false) {
         return tabbing + "request.TitleId = request.TitleId ?? PlayFabSettings.TitleId;\n"
             + tabbing + "if (request.TitleId == null) throw new PlayFabException(PlayFabExceptionCode.TitleNotSet, \"Must be have PlayFabSettings.TitleId set to call this method\");\n";
     if (apiCall.auth === "EntityToken")
-        return tabbing + "if ((request.AuthenticationContext?.EntityToken ?? PlayFabSettings.EntityToken) == null) throw new PlayFabException(PlayFabExceptionCode.EntityTokenNotSet, \"Must call GetEntityToken before calling this method\");\n";
-    if (apiCall.auth === "SessionTicket")
-        return tabbing + "var clientSessionTicket = request.AuthenticationContext?.ClientSessionTicket ?? PlayFabSettings.ClientSessionTicket;\n"
-            + tabbing + "if (clientSessionTicket == null) throw new PlayFabException(PlayFabExceptionCode.NotLoggedIn, \"Must be logged in to call this method\");\n";
+        return tabbing + "#if !DISABLE_PLAYFABENTITY_API\n"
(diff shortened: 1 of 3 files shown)


targets/csharp/make.js
```diff
@@ -56,7 +54,9 @@ function getBaseTypeSyntax(datatype) {
 
     if (datatype.className.toLowerCase().endsWith("request"))
         parents.push("PlayFabRequestCommon");
-    if (datatype.className.toLowerCase().endsWith("response") || datatype.className.toLowerCase().endsWith("result"))
+    if (datatype.className.toLowerCase() === "loginresult")
```
pgilmorepf: I realize that it wasn't you that created this problem, but let's fix it while you're here.
PlayFabRequestCommon, PlayFabLoginResultCommon, and PlayFabResultCommon are all classes.
Multi-Inheritance of classes is invalid in C#.
You can only inherit one class.

I think the best fix for now, is to wrap the "parent-classes" in some comments that say "begin classes - only 1 possible" and "end classes - only 1 possible", and make sure the if-switch inside only contains a single: if, elseif, elseif. 
We want to minimize the chance of a logic mistake producing two class-parents.

targets/csharp/make.js
```diff
@@ -56,7 +54,9 @@ function getBaseTypeSyntax(datatype) {
...
+        parents.push("PlayFabLoginResultCommon");
+    else if (datatype.className.toLowerCase().endsWith("response") || datatype.className.toLowerCase().endsWith("result"))
         parents.push("PlayFabResultCommon");
     if (datatype.sortKey)
```
pgilmorepf: Let's introduce a blankline here, and a new comment, separating it as the begining of the interface parents, of which there can be as many as is deemed necessary.
Basically, I want to reinforce the inheritance rules of C# here, so that we don't accidentally generate invalid code.

targets/csharp/make.js
```diff
@@ -262,36 +262,49 @@ function getAuthParams(apiCall, isApiInstance = false) {
...
             + tabbing + "if (request.TitleId == null) throw new PlayFabException(PlayFabExceptionCode.TitleNotSet, \"Must be have PlayFabSettings.TitleId set to call this method\");\n";
     if (apiCall.auth === "EntityToken")
-        return tabbing + "if (PlayFabSettings.EntityToken == null) throw new PlayFabException(PlayFabExceptionCode.EntityTokenNotSet, \"Must call GetEntityToken before calling this method\");\n";
+        return tabbing + "if ((request.AuthenticationContext?.EntityToken ?? PlayFabSettings.EntityToken) == null) throw new PlayFabException(PlayFabExceptionCode.EntityTokenNotSet, \"Must call GetEntityToken before calling this method\");\n";
     if (apiCall.auth === "SessionTicket")
-        return tabbing + "if (PlayFabSettings.ClientSessionTicket == null) throw new PlayFabException(PlayFabExceptionCode.NotLoggedIn, \"Must be logged in to call this method\");\n";
+        return tabbing + "var clientSessionTicket = request.AuthenticationContext?.ClientSessionTicket ?? PlayFabSettings.ClientSessionTicket;\n            if (clientSessionTicket == null) throw new PlayFabException(PlayFabExceptionCode.NotLoggedIn, \"Must be logged in to call this method\");\n";
```
pgilmorepf: Please look ten lines down for how to code-format multi-line returns.

targets/csharp/make.js
```diff
@@ -262,36 +262,49 @@ function getAuthParams(apiCall, isApiInstance = false) {
...
     if (apiCall.url === "/Authentication/GetEntityToken")
-        return tabbing + "string authKey = null, authValue = null;\n"
-            + tabbing + "if (PlayFabSettings.ClientSessionTicket != null) { authKey = \"X-Authorization\"; authValue = PlayFabSettings.ClientSessionTicket; }\n"
-            + tabbing + "if (PlayFabSettings.DeveloperSecretKey != null) { authKey = \"X-SecretKey\"; authValue = PlayFabSettings.DeveloperSecretKey; }\n"
-            + tabbing + "if (PlayFabSettings.EntityToken != null) { authKey = \"X-EntityToken\"; authValue = PlayFabSettings.EntityToken; }\n";
+        return tabbing + "string authKey = null, authValue = null;\n" 
+            + tabbing + "#if !DISABLE_PLAYFABCLIENT_API\n"
+            + tabbing + "        var clientSessionTicket = request.AuthenticationContext ?.ClientSessionTicket ?? PlayFabSettings.ClientSessionTicket;\n"
```
pgilmorepf: I think this section will result in a "double tabbed" effect on these lines.
basically you have:
```
+ tabbing + "no spaces"
+ tabbing + "        eight spaces" 
```
in all of this section, and I expect that will produce improperly tabbed code in the destination.

Also please don't put a space when using "?.":
AuthenticationContext ?.ClientSessionTicket

In your C# repo, please open in Visual Studio, and press "Ctrl + K + D", to reformat the file in the VS Standard file spacing, and then make the template produce exactly that spacing.

targets/csharp/source/source/Uunit/tests/PlayFabApiTest.cs
```diff
@@ -429,25 +382,15 @@ public void WriteEvent(UUnitTestContext testContext)
                 EventName = "ForumPostEvent",
                 Timestamp = DateTime.UtcNow,
                 Body = new Dictionary<string, object> {
-                    { "Subject", "My First Post" },
-                    { "Body", "My awesome Post." },
-                }
+                            { "Subject", "My First Post" },
+                            { "Body", "My awesome Post." },
+                        }
```
pgilmorepf: Do not change this spacing.

targets/csharp/source/source/Uunit/tests/PlayFabApiTest.cs
```diff
@@ -504,15 +447,15 @@ private void GetObjects1Continued(PlayFabResult<GetObjectsResponse> result, UUni
...
-                        ObjectName = TEST_DATA_KEY
-                    }
-                }
+                        {
+                            new SetObject
+                            {
+                                DataObject = _testInteger,
+                                ObjectName = TEST_DATA_KEY
+                            }
+                        }
```
pgilmorepf: Do not change this spacing.

targets/csharp/source/source/Uunit/tests/PlayFabApiTest.cs
```diff
@@ -537,10 +480,169 @@ private void GetObjects2Continued(PlayFabResult<GetObjectsResponse> result, UUni
...
-
             testContext.EndTest(UUnitFinishState.PASSED, null);
         }
 #endif
+        /// <summary>
+        /// CLIENT API
+        /// Test that Multiple login can be done
+        /// </summary>
+        [UUnitTest]
+        public void MultipleLogin(UUnitTestContext testContext)
```
pgilmorepf: Please change the name to indicate that it's a multiple login using the static interface.
The next team is going to come behind you and test multiple login using the instance interface, and they keep having to rename these tests.

targets/csharp/source/source/Uunit/tests/PlayFabApiTest.cs
```diff
@@ -537,10 +480,169 @@ private void GetObjects2Continued(PlayFabResult<GetObjectsResponse> result, UUni
...
+                CreateAccount = true,
+            };
+            var loginRequest2 = new LoginWithCustomIDRequest
+            {
+                TitleId = PlayFabSettings.TitleId,
+                CustomId = PlayFabSettings.BuildIdentifier,
+                CreateAccount = true,
+            };
+
+            var loginTask = PlayFabClientAPI.LoginWithCustomIDAsync(loginRequest, null, testTitleData.extraHeaders).Result;
```
pgilmorepf: Rename to loginTask1, since the other one is loginTask2

targets/csharp/source/source/Uunit/tests/PlayFabApiTest.cs
```diff
@@ -537,10 +480,169 @@ private void GetObjects2Continued(PlayFabResult<GetObjectsResponse> result, UUni
...
+            testContext.NotNull(loginTask.Result, "Login Result is null for loginRequest");
+            testContext.NotNull(loginTask2.Result, "Login Result is null for loginRequest2");
+            testContext.IsNull(loginTask.Error, "Login error occured for loginRequest: " + loginTask.Error?.ErrorMessage ?? string.Empty);
+            testContext.IsNull(loginTask2.Error, "Login error occured for loginRequest2: " + loginTask2.Error?.ErrorMessage ?? string.Empty);
+            testContext.NotNull(loginTask.Result.AuthenticationContext, "AuthenticationContext is not set for loginRequest");
+            testContext.NotNull(loginTask2.Result.AuthenticationContext, "AuthenticationContext is not set for loginRequest2");
+
+            if (string.Equals(loginTask.Result.AuthenticationContext.ClientSessionTicket, loginTask2.Result.AuthenticationContext.ClientSessionTicket))
+            {
+                testContext.Fail("Multiple Login Failed AuthenticationContexts are same");
```
pgilmorepf: Please add the ClientSessionTicket value to this fail message.

targets/csharp/source/source/Uunit/tests/PlayFabApiTest.cs
```diff
@@ -537,10 +480,169 @@ private void GetObjects2Continued(PlayFabResult<GetObjectsResponse> result, UUni
...
 #endif
+        /// <summary>
+        /// CLIENT API
+        /// Test that Multiple login can be done
+        /// </summary>
+        [UUnitTest]
+        public void MultipleLogin(UUnitTestContext testContext)
+        {
+            // If the setup failed to log in a user, we need to create one.
+            var loginRequest = new LoginWithCustomIDRequest
```
pgilmorepf: Likewise rename to _1

targets/csharp/source/source/Uunit/tests/PlayFabApiTest.cs
```diff
@@ -537,10 +480,169 @@ private void GetObjects2Continued(PlayFabResult<GetObjectsResponse> result, UUni
...
+            testContext.IsNull(loginTask.Error, "Login error occured for loginRequest: " + loginTask.Error?.ErrorMessage ?? string.Empty);
+            testContext.IsNull(loginTask2.Error, "Login error occured for loginRequest2: " + loginTask2.Error?.ErrorMessage ?? string.Empty);
+            testContext.NotNull(loginTask.Result.AuthenticationContext, "AuthenticationContext is not set for loginRequest");
+            testContext.NotNull(loginTask2.Result.AuthenticationContext, "AuthenticationContext is not set for loginRequest2");
+
+            if (string.Equals(loginTask.Result.AuthenticationContext.ClientSessionTicket, loginTask2.Result.AuthenticationContext.ClientSessionTicket))
+            {
+                testContext.Fail("Multiple Login Failed AuthenticationContexts are same");
+            }
+            testContext.EndTest(UUnitFinishState.PASSED, null);
```
pgilmorepf: At the end of this test, before passing, please save off AuthenticationContext1 and AuthenticationContext2 into private test class varaibles, and then in the next test...

targets/csharp/source/source/Uunit/tests/PlayFabApiTest.cs
```diff
@@ -537,10 +480,169 @@ private void GetObjects2Continued(PlayFabResult<GetObjectsResponse> result, UUni
...
+
+        /// <summary>
+        /// CLIENT API
+        /// Test that Multiple player can do api calls
+        /// </summary>
+        [UUnitTest]
+        public void MultiplePlayerApiCall(UUnitTestContext testContext)
+        {
+            // If the setup failed to log in a user, we need to create one.
+            var loginRequest = new LoginWithCustomIDRequest
```
pgilmorepf: This test shouldn't copy-paste the entire set of steps performed in the previous test, it should just check that the AuthenticationContext1 & AuthenticationContext2 are set from the previous test, and then use them.

targets/csharp/source/source/Uunit/tests/PlayFabApiTest.cs
```diff
@@ -537,10 +480,169 @@ private void GetObjects2Continued(PlayFabResult<GetObjectsResponse> result, UUni
...
+            var getPlayerProfileRequest = new GetPlayerProfileRequest()
+            {
+                AuthenticationContext = loginTask.Result.AuthenticationContext,
+                PlayFabId = PlayFabId
+            };
+            var getPlayerProfileRequest2 = new GetPlayerProfileRequest()
+            {
+                AuthenticationContext = loginTask2.Result.AuthenticationContext,
+                PlayFabId = PlayFabId
+            };
```
pgilmorepf: Reusing AuthenticationContext from the previous test, would remove this many lines of copy-paste, which I think is a greater sin than imperfect test isolation.
As long as the first step is verifying that the context from the previous test was set correctly, it's fine.

targets/csharp/source/source/Uunit/tests/PlayFabApiTest.cs
pgilmorepf: If the contexts aren't set, use the SKIPPED state, which is ideal to indicate "I just couldn't do this test"

targets/csharp/source/source/Uunit/tests/PlayFabApiTest.cs
```diff
@@ -537,10 +480,169 @@ private void GetObjects2Continued(PlayFabResult<GetObjectsResponse> result, UUni
...
+        /// CLIENT API
+        /// Tests async api calls with multiple users
+        /// </summary>
+        [UUnitTest]
+        public void AsyncApiCallWithMultipleUser(UUnitTestContext testContext)
+        {
+            var loginRequest = new LoginWithCustomIDRequest
+            {
+                TitleId = PlayFabSettings.TitleId,
+                CustomId = PlayFabSettings.BuildIdentifier,
```
pgilmorepf: This test seems very weird.
1. This test should be using three different players, and you're using the same one three times.

targets/csharp/source/source/Uunit/tests/PlayFabApiTest.cs
```diff
@@ -537,10 +480,169 @@ private void GetObjects2Continued(PlayFabResult<GetObjectsResponse> result, UUni
...
+
+            testContext.NotNull(loginTask.Result, "Login Result is null for loginRequest");
+            testContext.NotNull(loginTask2.Result, "Login Result is null for loginRequest2");
+            testContext.NotNull(loginTask3.Result, "Login Result is null for loginRequest3");
+            testContext.IsNull(loginTask.Error, "Login error occured for loginRequest: " + loginTask.Error?.ErrorMessage ?? string.Empty);
+            testContext.IsNull(loginTask2.Error, "Login error occured for loginRequest2: " + loginTask2.Error?.ErrorMessage ?? string.Empty);
+            testContext.IsNull(loginTask3.Error, "Login error occured for loginRequest3: " + loginTask3.Error?.ErrorMessage ?? string.Empty);
+            testContext.NotNull(loginTask.Result.AuthenticationContext, "AuthenticationContext is not set for loginRequest");
+            testContext.NotNull(loginTask2.Result.AuthenticationContext, "AuthenticationContext is not set for loginRequest2");
+            testContext.NotNull(loginTask3.Result.AuthenticationContext, "AuthenticationContext is not set for loginRequest3");
```
pgilmorepf: 2. Once again, you're copy-pasting a lot of login code, which is undesirable.

targets/csharp/source/source/Uunit/tests/PlayFabApiTest.cs
```diff
@@ -537,10 +480,169 @@ private void GetObjects2Continued(PlayFabResult<GetObjectsResponse> result, UUni
...
+            var task3 = PlayFabClientAPI.GetUserDataAsync(new GetUserDataRequest() { AuthenticationContext = loginTask3.Result.AuthenticationContext }, null, testTitleData.extraHeaders);
+            var task4 = PlayFabClientAPI.GetUserDataAsync(new GetUserDataRequest() { AuthenticationContext = loginTask3.Result.AuthenticationContext }, null, testTitleData.extraHeaders);
+            var task5 = PlayFabClientAPI.GetUserDataAsync(new GetUserDataRequest() { AuthenticationContext = loginTask2.Result.AuthenticationContext }, null, testTitleData.extraHeaders);
+            var task6 = PlayFabClientAPI.GetUserDataAsync(new GetUserDataRequest() { AuthenticationContext = loginTask.Result.AuthenticationContext }, null, testTitleData.extraHeaders);
+            var task7 = PlayFabClientAPI.GetUserDataAsync(new GetUserDataRequest() { AuthenticationContext = loginTask2.Result.AuthenticationContext }, null, testTitleData.extraHeaders);
+            var task8 = PlayFabClientAPI.GetUserDataAsync(new GetUserDataRequest() { AuthenticationContext = loginTask3.Result.AuthenticationContext }, null, testTitleData.extraHeaders);
```
pgilmorepf: 3. You're creating 8 tasks to load the same data repeatedly, which isn't really going to properly test the isolation of those logins. This part makes me think I should go look at the original test, because there's probably a design error with the test. For now, reduce it to just 3 calls, 1 for each [separate] player.

targets/csharp/make.js
melihceyhan: Sorry I made a change but I didn't clearly understand your expectation. What i changed is ok for you? Please update me for your this and below comments

targets/csharp/source/source/Uunit/tests/PlayFabApiTest.cs
melihceyhan: I did not changed this method yet because if I should test with 3 different player i need 3 different titleId. Am I wrong? I will modify this method after clarifying it in my mind

targets/csharp/source/source/Uunit/tests/PlayFabApiTest.cs
Vicodex: The other test (for instanced APIs) is in another test class, but i agree it's better to rename this method as you proposed

---
In reply to: [257055384](https://github.com/PlayFab/SDKGenerator/pull/404#discussion_r257055384) [](ancestors = 257055384)

targets/csharp/source/source/Uunit/tests/PlayFabApiTest.cs
Vicodex: Melih, i think you can use different CustomIds to distinguish them

---
In reply to: [257152426](https://github.com/PlayFab/SDKGenerator/pull/404#discussion_r257152426) [](ancestors = 257152426)

melihceyhan: Oh thanks! I will do this ASAP

melihceyhan: I commit it now. Does it meet the needs?

pgilmorepf:
Sorry for the delay. I just barely missed this one today. I'll get it first thing in the morning.

Vicodex:
Melih, please make sure that all tests pass in C# SDK when you generate it from SDK Generator with these changes

melihceyhan:
@Vicodex  thanks for your comments! There are unit test results from playfaball playfabclient and playfabserver sdks which are generated from sdkgen

playfabclient
playfabserver
playfaball

targets/csharp/make.js
```diff
@@ -262,36 +265,63 @@ function getAuthParams(apiCall, isApiInstance = false) {
...
     if (apiCall.auth === "SecretKey" && isApiInstance === true)
-        return tabbing + "var developerSecretKey = request.AuthenticationContext?.DeveloperSecretKey ?? (authenticationContext?.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey);\n            if (developerSecretKey == null) throw new PlayFabException(PlayFabExceptionCode.DeveloperKeyNotSet, \"DeveloperSecretKey is not found in Request, Server Instance or PlayFabSettings\");\n";
+        return tabbing + "#if ENABLE_PLAYFABSERVER_API || ENABLE_PLAYFABADMIN_API\n"
+			+ tabbing + "    var developerSecretKey = request.AuthenticationContext?.DeveloperSecretKey ?? (authenticationContext?.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey);\n"
+			+ tabbing + "    if (developerSecretKey == null) throw new PlayFabException(PlayFabExceptionCode.DeveloperKeyNotSet, \"DeveloperSecretKey is not found in Request, Server Instance or PlayFabSettings\");\n"
+			+ tabbing + "#else\n"    
+			+ tabbing + "    var developerSecretKey = PlayFabSettings.DeveloperSecretKey;\n"
+			+ tabbing + "#endif\n";
```
pgilmorepf: Theoretically, it should not be required to check ENABLE_PLAYFABSERVER_API  on every server-authenticated API method, because the entire API file should be wrapped properly.
If you're experiencing a problem that requires this, you should step back and try and figure out why the entire file isn't wrapped in ENABLE_PLAYFABSERVER_API.

targets/csharp/make.js
```diff
@@ -262,36 +265,63 @@ function getAuthParams(apiCall, isApiInstance = false) {
...
+			+ tabbing + "#else\n"    
+			+ tabbing + "    if ((PlayFabSettings.EntityToken) == null) throw new PlayFabException(PlayFabExceptionCode.EntityTokenNotSet, \"Must call GetEntityToken before calling this method\");\n"
+			+ tabbing + "#endif\n";
+	if (apiCall.auth === "SessionTicket")
+        return tabbing + "#if !DISABLE_PLAYFABCLIENT_API\n"
+	        + tabbing + "    var clientSessionTicket = request.AuthenticationContext?.ClientSessionTicket ?? PlayFabSettings.ClientSessionTicket;\n"
+            + tabbing + "    if (clientSessionTicket == null) throw new PlayFabException(PlayFabExceptionCode.NotLoggedIn, \"Must be logged in to call this method\");\n"
+			+ tabbing + "#else\n"    
+			+ tabbing + "    var clientSessionTicket = PlayFabSettings.ClientSessionTicket;\n"
+			+ tabbing + "#endif\n";
```
pgilmorepf: Theoretically, it should not be required to check !DISABLE_PLAYFABCLIENT_API on every sessionTicket-authenticated API method, because the entire API file should be wrapped properly.
If you're experiencing a problem that requires this, you should step back and try and figure out why the entire file isn't wrapped in !DISABLE_PLAYFABCLIENT_API.

targets/csharp/make.js
```diff
@@ -262,36 +265,63 @@ function getAuthParams(apiCall, isApiInstance = false) {
...
     if (apiCall.auth === "EntityToken")
-        return tabbing + "if (PlayFabSettings.EntityToken == null) throw new PlayFabException(PlayFabExceptionCode.EntityTokenNotSet, \"Must call GetEntityToken before calling this method\");\n";
-    if (apiCall.auth === "SessionTicket")
-        return tabbing + "if (PlayFabSettings.ClientSessionTicket == null) throw new PlayFabException(PlayFabExceptionCode.NotLoggedIn, \"Must be logged in to call this method\");\n";
+        return tabbing + "#if !DISABLE_PLAYFABENTITY_API\n"
+	        + tabbing + "    if ((request.AuthenticationContext?.EntityToken ?? PlayFabSettings.EntityToken) == null) throw new PlayFabException(PlayFabExceptionCode.EntityTokenNotSet, \"Must call GetEntityToken before calling this method\");\n"
+			+ tabbing + "#else\n"    
+			+ tabbing + "    if ((PlayFabSettings.EntityToken) == null) throw new PlayFabException(PlayFabExceptionCode.EntityTokenNotSet, \"Must call GetEntityToken before calling this method\");\n"
+			+ tabbing + "#endif\n";
```
pgilmorepf: Theoretically, it should not be required to check !DISABLE_PLAYFABENTITY_API on every entityToken-authenticated API method, because the entire API file should be wrapped properly.
If you're experiencing a problem that requires this, you should step back and try and figure out why the entire file isn't wrapped in !DISABLE_PLAYFABENTITY_API.

targets/csharp/make.js
```diff
@@ -262,36 +265,63 @@ function getAuthParams(apiCall, isApiInstance = false) {
...
-        return tabbing + "string authKey = null, authValue = null;\n"
-            + tabbing + "if (PlayFabSettings.ClientSessionTicket != null) { authKey = \"X-Authorization\"; authValue = PlayFabSettings.ClientSessionTicket; }\n"
-            + tabbing + "if (PlayFabSettings.DeveloperSecretKey != null) { authKey = \"X-SecretKey\"; authValue = PlayFabSettings.DeveloperSecretKey; }\n"
-            + tabbing + "if (PlayFabSettings.EntityToken != null) { authKey = \"X-EntityToken\"; authValue = PlayFabSettings.EntityToken; }\n";
+        return tabbing + "string authKey = null, authValue = null;\n" 
+            + tabbing + "#if !DISABLE_PLAYFABCLIENT_API\n"
+            + tabbing + "    var clientSessionTicket = request.AuthenticationContext?.ClientSessionTicket ?? PlayFabSettings.ClientSessionTicket;\n"
+            + tabbing + "    if (clientSessionTicket != null) { authKey = \"X-Authorization\"; authValue = clientSessionTicket; }\n"
+            + tabbing + "#endif\n"
+            + tabbing + "\n"
```
pgilmorepf: Whenever you have:
```+ tabbing + "\n"```
You should not do that.
You should instead end the previous line with ```".../n/n";```
You're creating an empty line with spaces, which does not match the standard VS spacing, IE the file spacing when you load the file in VS, and press "Ctrl + K + D"

File: targets/csharp/make.js [hunk 21]
```diff
@@ -262,36 +265,63 @@ function getAuthParams(apiCall, isApiInstance = false) {
...
+            + tabbing + "#if ENABLE_PLAYFABSERVER_API || ENABLE_PLAYFABADMIN_API\n"
+            + tabbing + "    var developerSecretKey = request.AuthenticationContext?.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey;\n"
+            + tabbing + "    if (developerSecretKey != null) { authKey = \"X-SecretKey\"; authValue = developerSecretKey; }\n"
+            + tabbing + "#endif\n"
+            + tabbing + "\n"
+            + tabbing + "#if !DISABLE_PLAYFABENTITY_API\n"
+            + tabbing + "    var entityToken = request.AuthenticationContext?.EntityToken ?? PlayFabSettings.EntityToken;\n"
+            + tabbing + "    if (entityToken != null) { authKey = \"X-EntityToken\"; authValue = entityToken; }\n"
+            + tabbing + "#endif\n";
```
pgilmorepf: The #if 's and flags defined for GetEntityToken here are correct.
You DO need to add them here, because GetEntityToken is an Entity-Group method, but Client and Server may be enabled or disabled independently.
So, GetEntityToken logic is good and correct here (except for the invisible spacing on empty lines).

targets/csharp/make.js
```diff
@@ -54,13 +52,18 @@ exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
...
+    else if (datatype.className.toLowerCase() === "loginresult")
+        parents.push("PlayFabLoginResultCommon");
+    else if (datatype.className.toLowerCase().endsWith("response") || datatype.className.toLowerCase().endsWith("result"))
         parents.push("PlayFabResultCommon");
+    //end classes - only 1 
+
     if (datatype.sortKey)
         parents.push("IComparable<" + datatype.name + ">");
-
+
```
pgilmorepf: Please don't add spaces to empty lines.

targets/csharp/make.js
melihceyhan: as we talked in #31's PR I don't think there is an error because we already generate projects with these macros but to prevent a future confision I add them. Should I remove?

image

pgilmorepf:
This looks good to me and Victor.
We're ready to land. Can you confirm there are no more pending changes on your side Melih?

melihceyhan:
Thanks! There is no more change

22 Şub 2019 Cum, saat 00:42 tarihinde Paul Gilmore <notifications@github.com>
şunu yazdı:

PR #405, opened 2019-02-15
Adding GSDK to PF Java SDK

haroniti:
Here's as far as I got, sending out a PR for feedback but most likely someone else will have to take it up to complete it. The branch is pushed though :) 

- It builds!
- Need to run a test app against the mock agent, will update this PR if I get a chance
- Sadly we never had Unit Tests for the Java GSDK, will add a task for this
- NOTE that this changes the PF Java SDK to use Java 1.8, need signoff from SDK team on whether this is ok

### DIFF
--- a/targets/java/source_shared/src/main/java/com/playfab/gsdk/FileLogger.java
+++ b/targets/java/source_shared/src/main/java/com/playfab/gsdk/FileLogger.java
@@ -1,0 +1,107 @@
+package com.playfab.gsdk;
+
+import java.io.BufferedOutputStream;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.nio.file.Files;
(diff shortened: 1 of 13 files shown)


targets/java/source_shared/src/main/java/com/playfab/gsdk/FileLogger.java
```diff
@@ -0,0 +1,107 @@
+package com.playfab.gsdk;
```
ravarnamsft: might wanna add the microsoft header in the files? (or we don't because this is open sourced?

pgilmorepf:
I'm gonna close this PR for now, with the idea that we'll be discussing the Common SDK, and plugins, and the overall project architecture more later, and we'll try to kick this off again after those prerequisites are cleared.

PR #409, opened 2019-02-21
Milestone 20: Xamarin project that runs PlayFab C# unit tests and displays the results

M-Pixel:
The csharp SDK Generator target now contains XamarinTestRunner, a Visual Studio solution for a Xamarin port of the UnittestRunner project (a win32 console application).  The purpose of this new solution is to test the C# SDK on Android, iOS, and UWP.

A few minor changes were made to the test framework in order to make it run properly on iOS.

The solution folder contains cmd files that produce application packages for the various target platforms using MSBuild, and a readme that explains the setup process required before the project can be built.  Those steps are not repeated in this documentation, please reference the readme file.  The build scripts are a supplemental alternative to building from Visual Studio, which is also viable.

The application itself is simple, consisting of only two elements:  A button to run the tests, and a scrolling text output that displays the test results.  If an exception occurs, the exception and its stack-trace will be displayed in the text output.


targets/csharp/make.js
```diff
@@ -1,16 +1,18 @@
...
 
 // Making resharper less noisy - These are defined in Generate.js
-if (typeof (copyFile) === "undefined") copyFile = function () { };
-if (typeof (copyTree) === "undefined") copyTree = function () { };
+if (typeof (copyOrTemplatizeFile) === "undefined") copyOrTemplatizeFile = function () { };
+if (typeof (templatizeTree) === "undefined") templatizeTree = function () { };
 if (typeof (generateApiSummaryLines) === "undefined") generateApiSummaryLines = function () { };
 if (typeof (getCompiledTemplate) === "undefined") getCompiledTemplate = function () { };
 
+var copyright = "\"Copyright © 2019\"";
```
pgilmorepf: Copyright Microsoft C 2019

targets/csharp/make.js
```diff
@@ -49,6 +52,11 @@ exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
     }
     generateSimpleFiles(apis, sourceDir, apiOutputDir);
     generateProject(apis, sourceDir, apiOutputDir, "All", ";ENABLE_PLAYFABADMIN_API;ENABLE_PLAYFABSERVER_API;ENABLE_PLAYFABCLIENT_API");
+
+    const xamarinOutputDir = path.join(rootOutputDir, "XamarinTestRunner");
+    var locals = {copyright: copyright};
```
pgilmorepf: Let's insert this line before line 44...

targets/csharp/make.js
```diff
@@ -35,12 +37,13 @@ exports.makeServerAPI = function (apis, sourceDir, apiOutputDir) {
...
 
 exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
+    const rootOutputDir = apiOutputDir;
     apiOutputDir = path.join(apiOutputDir, "PlayFabSDK");
     console.log("Generating C-sharp combined SDK to " + apiOutputDir);
 
-    copyTree(path.resolve(sourceDir, "source"), apiOutputDir);
-    copyTree(path.resolve(sourceDir, "UnittestRunner"), path.resolve(apiOutputDir, "UnittestRunner")); // Copy the actual unittest project in the CombinedAPI
-    copyFile(path.resolve(sourceDir, "PlayFabSDK+Unit.sln"), path.resolve(apiOutputDir, "PlayFabSDK+Unit.sln"));
+    templatizeTree({}, path.resolve(sourceDir, "source"), apiOutputDir);
```
pgilmorepf: ... and replace these {} first params with "locals" that should be moved up, so it can be ref'd here.
(I don't want to end up creating locals twice in a future commit)

targets/csharp/make.js
M-Pixel: I want to double-check before fulfilling this request.  "Declare when needed" is used consistently throughout the rest of this file.  If I change this declaration to "manually hoist", it will become inconsistent with the convention used throughout the rest of the file.

That, of course, begs the question of whether the solution is "then do it everywhere" instead of "never mind, don't do it".  However, I've found that "declare when needed" is generally favored in JS style guides nowadays.  Manual hoisting really only benefits JS-unfamiliar developers who don't understand hoisting, and only in specific scenarios where it makes a functional difference, while "declare when needed" improves the code's self-documenting qualities.

PR #411, opened 2019-02-22
Benran/sdkpowershell

veleek:
Added a powershell script to simplify generating SDKs from the command line a bit.  Automatically clones repositories that you don't have yet.  Makes it easier to customize the build (adding flags, etc)

This definitely doesn't NEED to be merged, just trying to get feedback on whether anybody thinks it's useful.  I found this easier than manually updating .bat files for one or more builds I was interested in generating and then needing to make sure that I don't check those in.


veleek:
Ping

pgilmorepf:
Hey there! I agree the merges have put this PR into a pretty weird state.

Obviously, this PR needs to be fixed, but looking at the changes that are specifically yours, I'm liking it.
I'll try to address them by looking at your individual commits, and not the merge stuff.

PS script looks pretty good. I'd totally land that.
I'm a big fan of the ContainsKey to TryGet conversions. Unity and CSharp.
My one complaint is that I would want to accept those as 3 separate Pull Requests, not one merged one. I consider SdkGenerator, CSharp, and Unity to be 3 separate products, and I try to avoid checkins that affect more than one product at a time.

pgilmorepf:
So what I'd say is, give me 3 fresh new code reviews for these things without the merge issues and I'll accept them.
Also, if I missed something you did, and it got mixed up with the merge, try and make sure that each sdk is in a separate PR, and that SdkGenerator changes are also in a separate pr.
I appreciate your contribution!

PR #412, opened 2019-02-22
Milestone #24: Multi-user support for Unity.

RustamMG:
PlayFab currently supports only a single login of a player in Unity SDK.
Added support multiple players logging from within the same SDK (Client).

- Introduce new class called PlayFabAuthenticationContext.
- Modify PlayFabRequestCommon to contain a pointer to PlayFabAuthenticationContext.
- PlayFabLoginResultCommon class based on an existing class PlayFabResultCommon. This’ll also have a reference to PlayFabAuthenticationContext.

### DIFF
--- a/targets/unity-v2/Testing/Tests/Client/ClientMultiUserTests.cs
+++ b/targets/unity-v2/Testing/Tests/Client/ClientMultiUserTests.cs
@@ -53,11 +53,11 @@ namespace PlayFab.UUnit
                 if (_player1 == null || _player2 == null) return;
 
                 // reset delegate to avoid calling this method again
-                _tickAction = delegate { };
+                _tickAction = null;
 
                 // players must not be equals
-                testContext.False(_player1 == _player2);
-                testContext.EndTest(UUnitFinishState.PASSED, "Two players are successfully login.");
+                testContext.False(_player1.PlayFabId == _player2.PlayFabId);
+                testContext.EndTest(UUnitFinishState.PASSED, "Two players are successfully login. " + _player1.PlayFabId + ", " + _player2.PlayFabId);
             };
         }
 
@@ -84,14 +84,14 @@ namespace PlayFab.UUnit
                 if (player1AccountInfo == null || player2AccountInfo == null) return;
 
                 // reset delegate to avoid calling this method again
-                _tickAction = delegate { };
+                _tickAction = null;
 
                 // compares received data with cached data to make sure everything works correctly
-                testContext.True(_player1.PlayFabId == player1AccountInfo.PlayFabId, "Player1 Ids not match!");
-                testContext.True(_player2.PlayFabId == player2AccountInfo.PlayFabId, "Player2 Ids not match!");
+                testContext.True(_player1.PlayFabId == player1AccountInfo.PlayFabId, "Player1 PlayFabIds not match!");
+                testContext.True(_player2.PlayFabId == player2AccountInfo.PlayFabId, "Player2 PlayFabIds not match!");
 
                 // playFabId at different players must be different
-                testContext.False(player1AccountInfo.PlayFabId == player2AccountInfo.PlayFabId, "Player ids is equals!");
+                testContext.False(player1AccountInfo.PlayFabId == player2AccountInfo.PlayFabId, "Players PlayFabId is equals!");
                 testContext.EndTest(UUnitFinishState.PASSED, _player1.PlayFabId + ", " + _player2.PlayFabId);
             };
         }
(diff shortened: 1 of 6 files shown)


targets/unity-v2/Testing/Tests/Client/ClientMultiUserTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            PlayFabClientAPI.LoginWithCustomID(player2LoginRequest, x => _player2 = x.AuthenticationContext,
+                PlayFabUUnitUtils.ApiActionWrapper<PlayFabError>(testContext, SharedErrorCallback), testContext);
+
+            _tickAction = () =>
+            {
+                // return if players not loaded yet
+                if (_player1 == null || _player2 == null) return;
+
+                // reset delegate to avoid calling this method again
+                _tickAction = delegate { };
```
pgilmorepf: An empty delegate performs work. Null would explicitly not perform work. And the end goal of "do nothing" would be accomplished either way. Please change _tickAction = null;

targets/unity-v2/Testing/Tests/Client/ClientMultiUserTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            PlayFabClientAPI.GetAccountInfo(player2Request, x => player2AccountInfo = x.AccountInfo,
+                PlayFabUUnitUtils.ApiActionWrapper<PlayFabError>(testContext, SharedErrorCallback), testContext);
+
+            _tickAction = () =>
+            {
+                // wait loading account info
+                if (player1AccountInfo == null || player2AccountInfo == null) return;
+
+                // reset delegate to avoid calling this method again
+                _tickAction = delegate { };
```
pgilmorepf: _tickAction = null;

targets/unity-v2/Testing/Tests/Client/ClientMultiUserTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            {
+                // return if players not loaded yet
+                if (_player1 == null || _player2 == null) return;
+
+                // reset delegate to avoid calling this method again
+                _tickAction = delegate { };
+
+                // players must not be equals
+                testContext.False(_player1 == _player2);
+                testContext.EndTest(UUnitFinishState.PASSED, "Two players are successfully login.");
```
pgilmorepf: Add the two IDs to the message.

targets/unity-v2/Testing/Tests/Client/ClientMultiUserTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            _tickAction = () =>
+            {
+                // return if players not loaded yet
+                if (_player1 == null || _player2 == null) return;
+
+                // reset delegate to avoid calling this method again
+                _tickAction = delegate { };
+
+                // players must not be equals
+                testContext.False(_player1 == _player2);
```
pgilmorepf: You're testing that the object references are different here, which doesn't actually confirm that they're different players.
You should be testing that _player1.PlayFabId is different from _player2.PlayFabId

targets/unity-v2/Testing/Tests/Client/ClientMultiUserTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            {
+                // wait loading account info
+                if (player1AccountInfo == null || player2AccountInfo == null) return;
+
+                // reset delegate to avoid calling this method again
+                _tickAction = delegate { };
+
+                // compares received data with cached data to make sure everything works correctly
+                testContext.True(_player1.PlayFabId == player1AccountInfo.PlayFabId, "Player1 Ids not match!");
+                testContext.True(_player2.PlayFabId == player2AccountInfo.PlayFabId, "Player2 Ids not match!");
```
pgilmorepf: Add the PlayFabIds to the messages here, so we can debug what went wrong.

targets/unity-v2/Testing/Tests/Client/ClientMultiUserTests.cs
```diff
@@ -0,0 +1,100 @@
...
+        private PlayFabAuthenticationContext _player1;
+        private PlayFabAuthenticationContext _player2;
+
+        public override void Tick(UUnitTestContext testContext)
+        {
+            if (_tickAction != null)
+                _tickAction();
+        }
+
+        private void SharedErrorCallback(PlayFabError error)
```
pgilmorepf: I'm pretty much at the point where this method should be in PlayFabUUnitUtils beside ApiActionWrapper.
Every TestCase is independently copy-pasting this method.

pgilmorepf: @Vicodex, can you make a tech debt task for us to do this someday?

targets/unity-v2/Testing/Tests/Client/ClientMultiUserTests.cs
pgilmorepf: By "performs work" I mean it still has the overall work of performing a runtime-allocated function call, with an empty body, which isn't much, but it's more than nothing.

PR #413, opened 2019-02-22
Cleanup for  Milestones 30a and 30b (Multi-Instance Support for UE4)

BrendanLoBuglio:
As requested, these are changes requested for the 30a and 30b code that didn't make it into [the original PR](https://github.com/PlayFab/SDKGenerator/pull/407)


Vicodex:
> 

It looks like this file doesn't have any changes. You may revert it in this PR

---
Refers to: targets/UnrealMarketplacePlugin/templates/PlayFabCpp/core/PlayFab_API.cpp.ejs:106 in a6b1eef. [](commit_id = a6b1eefd20eec65019f33f3645168c4626ce0c9c, deletion_comment = False)

PR #14, opened 2015-08-10
JavaScript SDK

sivakatir:
[Review on Reviewable](https://reviewable.io/reviews/playfab/sdkgenerator/14)

PR #414, opened 2019-02-22
(MS #14) Create a new automated, cross-platform test framework

HowardRoeske:
https://github.com/PlayFab/XPlatBetaSdk/pull/31

### DIFF
--- a/targets/XPlatCppSdk/source/test/TestApp/PlayFabApiTest.h
+++ b/targets/XPlatCppSdk/source/test/TestApp/PlayFabApiTest.h
@@ -1,0 +1,178 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+
+#pragma once
+
+#include <functional>
+#include <string>
+#include "TestCase.h"
+
+namespace PlayFab
+{
+    struct PlayFabError;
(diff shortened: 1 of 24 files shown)


targets/XPlatCppSdk/source/test/TestApp/PlayFabApiTest.h
```diff
@@ -0,0 +1,178 @@
...
+#pragma once
+
+#include <functional>
+#include <string>
+#include "TestCase.h"
+
+namespace PlayFab
+{
+    struct PlayFabError;
+    class PlayFabEventAPI;
```
Vicodex: >PlayFabEventAPI [](start = 10, length = 15)

This possibly is not needed in this file

pgilmorepf: Confirmed, doesn't seem necessary.

pgilmorepf:
It would be nice if you removed the one line, PlayFabEventAPI forward decl, that Victor found, but no review is required after that.

HowardRoeske:
Forward declaration has been removed.

PR #417, opened 2019-02-25
Update PlayFabHTTP.java

toddbellMSFT:
adding a timeout to http call


targets/java/source_shared/src/main/java/com/playfab/internal/PlayFabHTTP.java
```diff
@@ -13,6 +13,7 @@
 import com.playfab.PlayFabErrors.PlayFabJsonSuccess;
 
 public class PlayFabHTTP {
+    private static int timeoutInMilliseconds = 60000;
```
Vicodex: >private static int [](start = 4, length = 18)

You may use private const int here (even better)

PR #418, opened 2019-02-27
Remove string.empty

veleek:
When looking at the SDKGenerator solution, replacing "" with string.Empty doesn't complain.  However, when attempting to build the generated solution using this it does.  This was my fault from a previous commit so this is just reverting that specific change back.

### DIFF
--- a/targets/csharp/source/source/PluginManager.cs
+++ b/targets/csharp/source/source/PluginManager.cs
@@ -25,7 +25,7 @@ namespace PlayFab
         /// <param name="contract">The plugin contract.</param>
         /// <param name="instanceName">The optional plugin instance name. Instance names allow to have mulptiple plugins with the same contract.</param>
         /// <returns>The plugin instance.</returns>
-        public static T GetPlugin<T>(PluginContract contract, string instanceName = string.Empty) where T : IPlayFabPlugin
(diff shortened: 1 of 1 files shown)


veleek:
@Vicodex - You merged my previous change, but it looks like there's a bug that doesn't show up when looking at the code in the SDK Generator solution, only when using the generated output.  Can you please merge this ASAP as the SDK output is currently broken.

targets/csharp/source/source/PluginManager.cs
```diff
@@ -25,7 +25,7 @@ private PluginManager()
         /// <param name="contract">The plugin contract.</param>
         /// <param name="instanceName">The optional plugin instance name. Instance names allow to have mulptiple plugins with the same contract.</param>
         /// <returns>The plugin instance.</returns>
-        public static T GetPlugin<T>(PluginContract contract, string instanceName = string.Empty) where T : IPlayFabPlugin
+        public static T GetPlugin<T>(PluginContract contract, string instanceName = "") where T : IPlayFabPlugin
```
sanchitgarg: Consider using String.IsNullOrEmpty(s)
Not sure if we have one **s**tring.IsNullOrEmpty. If so, use that.

PR #419, opened 2019-02-27
Adding a user exception callback for background threads

toddbellMSFT:
we can potentially throw an exception on a background thread. We should allow users to handle these exceptions as they wish.

### DIFF
--- a/targets/XPlatCppSdk/source/code/include/playfab/PlayFabError.h.ejs
+++ b/targets/XPlatCppSdk/source/code/include/playfab/PlayFabError.h.ejs
@@ -40,4 +40,5 @@ namespace PlayFab
     };
 
     typedef std::function<void(const PlayFabError& error, void* customData)> ErrorCallback;
+    typedef std::function<void(const std::exception& exception)> ExceptionCallback;
 }
--- a/targets/XPlatCppSdk/source/code/include/playfab/PlayFabError.h.ejs
+++ b/targets/XPlatCppSdk/source/code/include/playfab/PlayFabError.h.ejs
@@ -40,5 +40,5 @@ namespace PlayFab
     };
 
     typedef std::function<void(const PlayFabError& error, void* customData)> ErrorCallback;
-    typedef std::function<void(const std::exception& exception)> ExceptionCallback;
+    typedef std::function<void(std::exception exception)> ExceptionCallback;
 }
--- a/targets/XPlatCppSdk/source/code/include/playfab/PlayFabEventPipeline.h
+++ b/targets/XPlatCppSdk/source/code/include/playfab/PlayFabEventPipeline.h
@@ -83,7 +83,7 @@ namespace PlayFab
         PlayFabEventBuffer buffer;
         std::thread workerThread;
         std::atomic<bool> isWorkerThreadRunning;
-        std::mutex userCallbackMutex;
+        std::mutex userExceptionCallbackMutex;
         ExceptionCallback userExceptionCallback;
     };
 }
--- a/targets/XPlatCppSdk/source/code/include/playfab/PlayFabEventPipeline.h
+++ b/targets/XPlatCppSdk/source/code/include/playfab/PlayFabEventPipeline.h
@@ -8,6 +8,7 @@
 #include <playfab/PlayFabEvent.h>
 #include <playfab/PlayFabEventBuffer.h>
 #include <playfab/PlayFabAuthenticationContext.h>
+#include <mutex>
 
 namespace PlayFab
 {
@@ -58,6 +59,8 @@ namespace PlayFab
         virtual void Start() override;
         virtual void IntakeEvent(std::shared_ptr<const IPlayFabEmitEventRequest> request) override;
 
+        void SetExceptionCallback(ExceptionCallback callback);
+
     protected:
         virtual void SendBatch(size_t& batchCounter);
 
@@ -80,6 +83,8 @@ namespace PlayFab
         PlayFabEventBuffer buffer;
         std::thread workerThread;
         std::atomic<bool> isWorkerThreadRunning;
+        std::mutex userCallbackMutex;
+        ExceptionCallback userExceptionCallback;
     };
 }
 
--- a/targets/XPlatCppSdk/source/code/include/playfab/PlayFabPluginManager.h
+++ b/targets/XPlatCppSdk/source/code/include/playfab/PlayFabPluginManager.h
@@ -2,6 +2,7 @@
 
 #include <playfab/PlayFabCallRequestContainer.h>
 #include <playfab/PlayFabError.h>
+#include <mutex>
 #include <unordered_map>
 
 namespace PlayFab
@@ -95,6 +96,12 @@ namespace PlayFab
         // If a plugin with specified contract and optional instance name already exists, it will be replaced with specified instance.
         void SetPluginInstance(std::shared_ptr<IPlayFabPlugin> plugin, const PlayFabPluginContract contract, const std::string& instanceName = "");
 
+        // Sets a custom exception handler for any possible background thread exceptions
+        void SetExceptionHandler(ExceptionCallback exceptionHandler);
+
+        // Called when an exception occured on a worker thread
+        void HandleException(const std::exception&);
+
     private:
         std::shared_ptr<IPlayFabPlugin> GetPluginInternal(const PlayFabPluginContract contract, const std::string& instanceName);
         void SetPluginInternal(std::shared_ptr<IPlayFabPlugin> plugin, const PlayFabPluginContract contract, const std::string& instanceName);
@@ -105,5 +112,7 @@ namespace PlayFab
 
     private:
         std::map<const std::pair<const PlayFabPluginContract, const std::string>, std::shared_ptr<IPlayFabPlugin>> plugins;
+        std::mutex userExceptionCallbackMutex;
+        ExceptionCallback userExceptionCallback;
     };
 }
--- a/targets/XPlatCppSdk/source/code/include/playfab/PlayFabWinHttpPlugin.h
+++ b/targets/XPlatCppSdk/source/code/include/playfab/PlayFabWinHttpPlugin.h
@@ -25,6 +25,8 @@ namespace PlayFab
         virtual void MakePostRequest(std::unique_ptr<CallRequestContainerBase> requestContainer) override;
         virtual size_t Update() override;
 
+        void SetExceptionHandler(ExceptionCallback exceptionHandler);
+
     protected:
         virtual void ExecuteRequest(std::unique_ptr<CallRequestContainer> requestContainer);
         virtual std::string GetUrl(CallRequestContainer& requestContainer) const;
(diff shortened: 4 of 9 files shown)


targets/XPlatCppSdk/source/code/include/playfab/PlayFabError.h.ejs
```diff
@@ -40,4 +40,5 @@ namespace PlayFab
     };
 
     typedef std::function<void(const PlayFabError& error, void* customData)> ErrorCallback;
+    typedef std::function<void(const std::exception& exception)> ExceptionCallback;
```
Vicodex: >& [](start = 51, length = 1)

Please pass the copy of exception here, not a reference. We had a long discussion with Deb about this (i initially also proposed to use a reference). The main reason is we don't want external user to accidentally use it AFTER the exception object was destroyed (e.g. they save it somewhere and then attempt to dereference later). Talk to Deb if you have a different opinion.

targets/XPlatCppSdk/source/code/include/playfab/PlayFabWinHttpPlugin.h
```diff
@@ -25,6 +25,8 @@ namespace PlayFab
         virtual void MakePostRequest(std::unique_ptr<CallRequestContainerBase> requestContainer) override;
         virtual size_t Update() override;
 
+        void SetExceptionHandler(ExceptionCallback exceptionHandler);
```
Vicodex: >void SetExceptionHandler(ExceptionCallback exceptionHandler); [](start = 8, length = 61)

Please move this method to PlayFabPluginManager (as an instance method)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabWinHttpPlugin.h
```diff
@@ -42,5 +44,6 @@ namespace PlayFab
         int activeRequestCount;
         std::deque<std::unique_ptr<CallRequestContainerBase>> pendingRequests;
         std::deque<std::unique_ptr<CallRequestContainerBase>> pendingResults;
+        ExceptionCallback userExceptionCallback;
```
Vicodex: >ExceptionCallback userExceptionCallback; [](start = 8, length = 40)

Please move this PlayFabPluginManager (as an instance field)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabWinHttpPlugin.h
Vicodex: >void SetExceptionHandler(ExceptionCallback exceptionHandler); [](start = 8, length = 61)

Please move this to PlayFabPluginManager as an instance method

targets/XPlatCppSdk/source/code/include/playfab/PlayFabWinHttpPlugin.h
Vicodex: >ExceptionCallback userExceptionCallback; [](start = 8, length = 40)

Please move this to PlayFabPluginManager as an instance field

targets/XPlatCppSdk/source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -32,39 +32,52 @@ namespace PlayFab
...
         }
+        catch (std::exception ex)
+        {
+            { // LOCK httpRequestMutex
+                std::unique_lock<std::mutex> lock(httpRequestMutex);
+                if (userExceptionCallback)
+                {
+                    userExceptionCallback(ex);
+                }
+            } // UNLOCK httpRequestMutex
```
Vicodex: Please call here PlayFabPluginManager.GetInstanceHandlerException(const std::exception&) method instead of all this. Add that method to PlayFabPluginManager. That method will have this mutex block and it will call the user's callback, how you are doing here.

targets/XPlatCppSdk/source/code/include/playfab/PlayFabEventPipeline.h
```diff
@@ -80,7 +83,9 @@ namespace PlayFab
         PlayFabEventBuffer buffer;
         std::thread workerThread;
         std::atomic<bool> isWorkerThreadRunning;
+        std::mutex userCallbackMutex;
```
Vicodex: >userCallbackMutex [](start = 19, length = 17)

Let's name it also userExceptionCallback for consistency

targets/XPlatCppSdk/source/code/include/playfab/PlayFabEventPipeline.h
```diff
@@ -8,6 +8,7 @@
 #include <playfab/PlayFabEvent.h>
 #include <playfab/PlayFabEventBuffer.h>
 #include <playfab/PlayFabAuthenticationContext.h>
+#include <mutex>
```
Vicodex: Please move this to PlayFabPluginManager only

targets/XPlatCppSdk/source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -32,39 +32,52 @@ namespace PlayFab
 
     void PlayFabWinHttpPlugin::WorkerThread()
     {
-        size_t queueSize;
-
-        while (this->threadRunning)
+        try
         {
```
Vicodex: Please apply this fix to all other HTTP plugins that have WorkerThreads (Curl, Xbox)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabWinHttpPlugin.cpp
```diff
@@ -32,39 +32,46 @@ namespace PlayFab
...
+                    CallRequestContainer* requestContainerPtr = dynamic_cast<CallRequestContainer*>(requestContainer.get());
+                    if (requestContainerPtr != nullptr)
+                    {
+                        requestContainer.release();
+                        ExecuteRequest(std::unique_ptr<CallRequestContainer>(requestContainerPtr));
+                    }
                 }
             }
         }
+        catch (std::exception ex)
```
debhaldarMS: >std::exception ex [](start = 15, length = 17)

Always throw an exception by value and catch by reference or const reference if possible

debhaldarMS: > catch [](start = 7, length = 6)

we should also catch (...)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabPluginManager.cpp
```diff
@@ -99,4 +99,23 @@ namespace PlayFab
         return std::make_shared<OneDSCurlHttpPlugin>();
 #endif // PLAYFAB_PLATFORM_XBOX
     }
+
+    void PlayFabPluginManager::HandleException(const std::exception& ex)
```
debhaldarMS: This is supposed to be implemneted by a user right ? Then don't pass the exception by reference

targets/XPlatCppSdk/source/code/include/playfab/PlayFabPluginManager.h
```diff
@@ -95,6 +96,12 @@ namespace PlayFab
         // If a plugin with specified contract and optional instance name already exists, it will be replaced with specified instance.
         void SetPluginInstance(std::shared_ptr<IPlayFabPlugin> plugin, const PlayFabPluginContract contract, const std::string& instanceName = "");
 
+        // Sets a custom exception handler for any possible background thread exceptions
+        void SetExceptionHandler(ExceptionCallback exceptionHandler);
+
+        // Called when an exception occured on a worker thread
+        void HandleException(const std::exception&);
```
debhaldarMS: >& [](start = 49, length = 1)

Give them a copy of the exception

targets/XPlatCppSdk/source/code/include/playfab/PlayFabWinHttpPlugin.h
```diff
@@ -43,4 +43,4 @@ namespace PlayFab
         std::deque<std::unique_ptr<CallRequestContainerBase>> pendingRequests;
         std::deque<std::unique_ptr<CallRequestContainerBase>> pendingResults;
     };
-}
\ No newline at end of file
+}
```
debhaldarMS: extra space ?

targets/XPlatCppSdk/source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -168,10 +176,18 @@ namespace PlayFab
                 std::this_thread::sleep_for(std::chrono::milliseconds(this->settings->readBufferWaitTime));
             }
         }
-        catch (...)
+        catch (std::exception ex)
```
debhaldarMS: >std::exception ex [](start = 15, length = 17)

Catch by const reference

targets/XPlatCppSdk/source/code/source/playfab/PlayFabEventPipeline.cpp
```diff
@@ -287,4 +303,4 @@ namespace PlayFab
     }
 }
 
-#endif
\ No newline at end of file
+#endif
```
debhaldarMS: space ?

targets/XPlatCppSdk/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -29,39 +29,46 @@ namespace PlayFab
...
+                    CallRequestContainer* requestContainerPtr = dynamic_cast<CallRequestContainer*>(requestContainer.get());
+                    if (requestContainerPtr != nullptr)
+                    {
+                        requestContainer.release();
+                        ExecuteRequest(std::unique_ptr<CallRequestContainer>(requestContainerPtr));
+                    }
                 }
             }
         }
+        catch (std::exception ex)
```
debhaldarMS: > catch (std::exception ex) [](start = 7, length = 26)

catch ...

targets/XPlatCppSdk/source/code/include/playfab/PlayFabWinHttpPlugin.h
toddbellMSFT: It's a force of habit of mine to have a single blank space at the end of a file due to some legacy languages expecting this.

I tried removing the change yesterday, but git isn't counting it as a change for the commit. I think I'll have to change the file directly in github.

---
In reply to: [261333906](https://github.com/PlayFab/SDKGenerator/pull/419#discussion_r261333906) [](ancestors = 261333906)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabEventPipeline.cpp
toddbellMSFT: Same issue as above. git doesn't seem to understand removing these additional lines.

---
In reply to: [261336313](https://github.com/PlayFab/SDKGenerator/pull/419#discussion_r261336313) [](ancestors = 261336313)

PR #420, opened 2019-02-28
Milestone #29: Melihceyhan/unity multi instance support for servers

melihceyhan:
Milestone #29: Melihceyhan/unity multi instance support for servers


targets/unity-v2/Testing/Tests/Server/ServerInstanceApiTests.cs
```diff
@@ -0,0 +1,276 @@
...
+
+        /// <summary>
+        /// SERVER API
+        /// Multiple instances of the same API class can be created
+        /// </summary>
+        [UUnitTest]
+        public void CreateMultipleServerInstance(UUnitTestContext testContext)
+        {
+            PlayFabApiSettings settings = new PlayFabApiSettings();
+            settings.TitleId = PlayFabSettings.TitleId;
```
pgilmorepf: It would be more appropriate for:
```settings.TitleId = testTitleData.TitleId;```
Theoretically any titleId could be in PlayFabSettings, but testTitleData.TitleId is the one we're explicitly trying to use.

File: targets/unity-v2/Testing/Tests/Server/ServerInstanceApiTests.cs [hunk 2]
```diff
@@ -0,0 +1,276 @@
...
+        /// <summary>
+        /// SERVER API
+        /// Each API instance can be used to login a player separately from any other API instances, 
+        /// and that player�s authentication context is stored in the API instance
+        /// </summary>
+        [UUnitTest]
+        public void ApiInstanceLogin(UUnitTestContext testContext)
+        {
+            PlayFabApiSettings settings = new PlayFabApiSettings();
+            settings.TitleId = PlayFabSettings.TitleId;
```
pgilmorepf: testTitleData.TitleId, not PlayFabSettings

targets/unity-v2/Testing/Tests/Server/ServerInstanceApiTests.cs
```diff
@@ -0,0 +1,276 @@
...
+
+        /// <summary>
+        /// SERVER API
+        /// If API settings object is not set by a customer developer for an instance of an API class then the
+        /// API instance always uses the static PlayFabSettings class (the same way as static API classes)
+        /// </summary>
+        [UUnitTest]
+        public void CheckWithNoSettings(UUnitTestContext testContext)
+        {
+            PlayFabSettings.DeveloperSecretKey = testTitleData.developerSecretKey;
```
pgilmorepf: This should be two lines:
PlayFabSettings.TitleId = testTitleData.titleId;
PlayFabSettings.DeveloperSecretKey = testTitleData.developerSecretKey;
And both lines should be in Setup(), not in the test.

targets/unity-v2/Testing/Tests/Server/ServerInstanceApiTests.cs
```diff
@@ -0,0 +1,276 @@
...
+            testContext.EndTest(UUnitFinishState.PASSED, null);
+        }
+        /// <summary>
+        /// SERVER API
+        /// Try to check server instance with authentication context and without authentication context
+        /// </summary>
+        [UUnitTest]
+        public void CheckWithAuthContextAndWithoutAuthContext(UUnitTestContext testContext)
+        {
+            PlayFabSettings.DeveloperSecretKey = testTitleData.developerSecretKey;
```
pgilmorepf: This should be two lines:
PlayFabSettings.TitleId = testTitleData.titleId;
PlayFabSettings.DeveloperSecretKey = testTitleData.developerSecretKey;
And both lines should be in Setup(), not in the test.

targets/unity-v2/Testing/Tests/Server/ServerInstanceApiTests.cs
```diff
@@ -0,0 +1,276 @@
...
+		/// <summary>
+        /// SERVER API
+        /// Try to parallel request at same time
+        /// </summary>
+        [UUnitTest]
+        public void ParallelRequest(UUnitTestContext testContext)
+        {
+            
+            PlayFabApiSettings settings = new PlayFabApiSettings();
+            settings.TitleId = PlayFabSettings.TitleId;
```
pgilmorepf: testTitleData.TitleId

targets/unity-v2/make.js
```diff
@@ -378,7 +418,14 @@ function getAuthParams(apiCall) {
     return "AuthType.None";
 }
 
-function getRequestActions(tabbing, apiCall) {
+function getInstanceParams(apiCall) {
+    if (apiCall.auth === "SecretKey")
+        return "authenticationContext, apiSettings"
```
pgilmorepf: @Vicodex This condition seems VERY weird to me.
Why is this the case?
@melihceyhan, Feel free to answer or ignore. I just want to know why this choice is made.

targets/unity-v2/make.js
```diff
@@ -389,12 +436,11 @@ function getRequestActions(tabbing, apiCall) {
             tabbing + "if (authType == AuthType.None && !string.IsNullOrEmpty(PlayFabSettings.DeveloperSecretKey))\n" +
             tabbing + "    authType = AuthType.DevSecretKey;\n" +
             "#endif\n";
-
```
pgilmorepf: Let's keep this empty line.
GetEntityToken is a single API method that serves as a very special case, where all the ones below are standard api authentication handling.

targets/unity-v2/source/Shared/Public/PlayFabApiSettings.cs.ejs
```diff
@@ -0,0 +1,31 @@
...
+        public string VerticalName = null;
+        /// <summary> Session token for Entity API. Auto-Populated by GetEntityToken method. </summary>
+        internal string EntityToken = null;
+        /// <summary> Session ticket for Client API. Auto-Populated by any login or registration call. </summary>
+        internal string ClientSessionTicket = null;
+        /// <summary> You must set this value for PlayFabSdk to work properly (Found in the Game Manager for your title, at the PlayFab Website) </summary>
+        public string DeveloperSecretKey = null;
+        /// <summary> You must set this value for PlayFabSdk to work properly (Found in the Game Manager for your title, at the PlayFab Website) </summary>
+        public string TitleId;
+        public ErrorCallback GlobalErrorHandler;
```
pgilmorepf: Remove GlobalErrorHandler.
You're clearly not hooking it up to anything, and I don't think it's a good idea to have lots of "global" error handlers.

targets/unity-v2/make.js
melihceyhan: we use authenticationContext in MakeApiCall for only ServerInstance so we do not need to pass it. By the way apiSettings is needed for all Instance-able ones. I can also pass authenticationContext for all instance-able ones if you want. We should also wait for @Vicodex's comment

melihceyhan:
@pgilmorepf I did some changes is that what you want? please let me know if i got any misunderstanding

targets/unity-v2/templates/InstanceAPI.cs.ejs
```diff
@@ -0,0 +1,91 @@
...
+
+        public PlayFab<%- api.name %>InstanceAPI(PlayFabApiSettings settings) 
+        {
+            apiSettings = settings;
+        }
+
+        public PlayFab<%- api.name %>InstanceAPI(PlayFabAuthenticationContext context) 
+        {
+            authenticationContext = context;
+        }
```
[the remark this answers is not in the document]
Vicodex: Generally i didn't see harm in multiple constructors, to be honest, and actually prefer cleaner API designs where users are not forced to explicitly specify nulls for stuff that they are not using. I am seeing multiple constructors all over in standard C# library classes and i like the convenience of it. We also already have these constructors in C# SDK so i'd prefer to stay consistent - i.e. either remove in both, or have in both. Do you have any concerns about this?

---
In reply to: [261837292](https://github.com/PlayFab/SDKGenerator/pull/420#discussion_r261837292) [](ancestors = 261837292)

targets/unity-v2/templates/InstanceAPI.cs.ejs
```diff
@@ -0,0 +1,91 @@
...
+
+        public void SetSettings(PlayFabApiSettings settings)
+        {
+            apiSettings = settings;
+        }
+
+        public PlayFabApiSettings GetSettings()
+        {
+            return apiSettings;
+        }
```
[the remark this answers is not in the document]
Vicodex: I've seen really a lot of opinions on this :) It seems that majority of C# guides and tools like StyleCop consider hiding private fields behind public/internal/protected properties or methods a recommended practice. I understand that there can be some perf cost. If you ask me directly what i prefer then my answer would be to follow recommended standards in situations where performance is not an issue (like this one). An additional benefit of this wrapping is that we are hiding implementation details which is a good thing. One day SetSettings(...) may potentially do more than just setting a variable.

---
In reply to: [261837352](https://github.com/PlayFab/SDKGenerator/pull/420#discussion_r261837352) [](ancestors = 261837352)

targets/unity-v2/templates/InstanceAPI.cs.ejs
```diff
@@ -0,0 +1,91 @@
...
+            return authenticationContext;
+        }
+
+<% if (hasClientOptions) { %>
+        /// <summary>
+        /// Check to See if the client is logged in.
+        /// </summary>
+        public bool IsClientLoggedIn()
+        {
+            return PlayFabHttp.IsClientLoggedIn();
```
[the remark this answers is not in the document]
Vicodex: Hmm, isn't that what this method is doing? It returns authenticationContext.IsClientLoggedIn() when authenticationContext is set, and false (i.e. "not logged in") when authenticationContext is not set. The naming of method maybe not great. It really meant to be IsInstanceAuthenticated() or something along that line. But i'm fine keeping this name, for consistency with other SDKs

---
In reply to: [261837405](https://github.com/PlayFab/SDKGenerator/pull/420#discussion_r261837405) [](ancestors = 261837405)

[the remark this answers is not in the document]
Vicodex: Oh, he fixed that. Thanks Melih!

---
In reply to: [261837405](https://github.com/PlayFab/SDKGenerator/pull/420#discussion_r261837405) [](ancestors = 261837405)

targets/unity-v2/templates/InstanceAPI.cs.ejs
```diff
@@ -0,0 +1,91 @@
...
+        }
+<% }%>
+<% if (hasEntityTokenOptions) { %>
+        /// <summary>
+        /// Check to See if the entity is logged in.
+        /// </summary>
+        public bool IsEntityLoggedIn()
+        {
+            var transport = PluginManager.GetPlugin<IPlayFabTransportPlugin>(PluginContract.PlayFab_Transport);
+            return !string.IsNullOrEmpty(transport.EntityToken);
```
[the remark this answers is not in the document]
Vicodex: Really good catch, Paul. Thank you! I checked other SDKs and only C# has a problem (a bit different one: this method is checking PlayFabSettings there instead of checking authentication context). C++ and Unreal are fine, they have correct code.

---
In reply to: [261837473](https://github.com/PlayFab/SDKGenerator/pull/420#discussion_r261837473) [](ancestors = 261837473)

targets/unity-v2/make.js
Vicodex: If we can pass both arguments in call cases i think it will make it less confusing. Even if authenticationContext is not used in one of the cases, it looks like it won't harm if we pass it?

---
In reply to: [261843518](https://github.com/PlayFab/SDKGenerator/pull/420#discussion_r261843518) [](ancestors = 261843518)

targets/unity-v2/templates/InstanceAPI.cs.ejs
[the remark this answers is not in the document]
melihceyhan: Sure I will

targets/unity-v2/make.js
melihceyhan: Yes i think so. It wont harm anything. I commited it. Is there anything more to correct for this task?

targets/unity-v2/templates/InstanceAPI.cs.ejs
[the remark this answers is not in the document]
pgilmorepf: My thoughts are:
If we make it a public field today, it's the simplest possible code.
If we want it to be a property tomorrow, we can replace the field with a property, and it is not an SDK breaking change.
Basically I don't want to over complicate the code.
We have 10 lines today, where we only need 1 line.
I don't want this to be 10 lines until it NEEDS to be 10 lines.

targets/unity-v2/Testing/Tests/Server/ServerInstanceApiTests.cs
```diff
@@ -101,13 +103,13 @@ public void MultipleInstanceWithDifferentSettings(UUnitTestContext testContext)
         /// <summary>
         /// SERVER API
         /// Each API instance can be used to login a player separately from any other API instances, 
-        /// and that player�s authentication context is stored in the API instance
+        /// and that player�s authentication context is stored in the API instance
```
pgilmorepf: Please remove the unicode apostrophe and replace with an ansi ' character.

targets/unity-v2/templates/InstanceAPI.cs.ejs
```diff
@@ -58,7 +58,7 @@ namespace PlayFab
         /// </summary>
         public bool IsClientLoggedIn()
         {
-            return PlayFabHttp.IsClientLoggedIn();
+            return authenticationContext == null ? false : authenticationContext.IsClientLoggedIn();
```
pgilmorepf: What does an instance without an authenticationContext even mean?

melihceyhan:
Thanks @pgilmorepf  I did your update. I'm waiting for Victor's comments :)

targets/unity-v2/templates/InstanceAPI.cs.ejs
Vicodex: I think we talked already about the same thing in C#. One of the goals of the initial design was to provide a smooth upgrade path for existing customers from static APIs to instance APIs. Their code is not instrumented to use authentication contexts, but they can easily create instances of APIs using a default constructor. In that case authentication context is set to null (not instantiated by default!). It allows us to fallback to PlayFabSettings values when calling APIs, and everything just works as it was before with static APIs. BUT, if a customer decides to use authentication context explicitly, they can set it, and then it will be used for authentication.

---
In reply to: [262255530](https://github.com/PlayFab/SDKGenerator/pull/420#discussion_r262255530) [](ancestors = 262255530)

targets/unity-v2/source/Shared/Internal/PlayFabHttp/PlayFabHTTP.cs
```diff
@@ -186,17 +186,39 @@ public static void SimplePostCall(string fullUrl, byte[] payload, Action<byte[]>
...
+                if(developerSecretKey == null && authenticationContext != null && !string.IsNullOrEmpty(authenticationContext.DeveloperSecretKey))
+                {
+                    developerSecretKey = authenticationContext.DeveloperSecretKey;
+                }
+                if(developerSecretKey == null)
+                {
+                    developerSecretKey = PlayFabSettings.DeveloperSecretKey;
+                }
+            #else
+                developerSecretKey = PlayFabSettings.DeveloperSecretKey;
```
Vicodex: Please remove these two lines. If server or admin APIs are not enabled, we should not even be using developer secret keys

targets/unity-v2/make.js
```diff
@@ -378,7 +418,11 @@ function getAuthParams(apiCall) {
     return "AuthType.None";
 }
 
-function getRequestActions(tabbing, apiCall) {
+function getInstanceParams(apiCall) {
+    return "authenticationContext, apiSettings"
+}
```
Vicodex: Please remove this method entirely as it seems we don't need it at all. Just use those two parameters directly in template

targets/unity-v2/make.js
```diff
@@ -394,7 +438,7 @@ function getRequestActions(tabbing, apiCall) {
         return tabbing + "request.TitleId = request.TitleId ?? PlayFabSettings.TitleId;\n";
     if (apiCall.auth === "SessionTicket")
         return tabbing + "if (!IsClientLoggedIn()) throw new PlayFabException(PlayFabExceptionCode.NotLoggedIn,\"Must be logged in to call this method\");\n";
-    if (apiCall.auth === "SecretKey")
+    if (apiCall.auth === "SecretKey" && isApiInstance === false)
```
Vicodex: >if (apiCall.auth === "SecretKey" && isApiInstance === false) [](start = 4, length = 60)

We need to add correct login here when isApiInstance === true. Please take a look here how it is done for C#:

[https://github.com/PlayFab/SDKGenerator/blob/master/targets/csharp/make.js#L300](https://github.com/PlayFab/SDKGenerator/blob/master/targets/csharp/make.js#L300)

Please add that.

targets/unity-v2/make.js
```diff
@@ -378,7 +418,11 @@ function getAuthParams(apiCall) {
     return "AuthType.None";
 }
 
-function getRequestActions(tabbing, apiCall) {
+function getInstanceParams(apiCall) {
+    return "authenticationContext, apiSettings"
+}
+
+function getRequestActions(tabbing, apiCall, isApiInstance = false) {
```
Vicodex: Ok i see multiple things here that are wrong (we may actually fix some of them in C#, too). Please apply the following to EACH of the if (...) cases in this function:

1) If isApiInstance == false, then:
  a) Use values from request.AuthenticationContext (if it is not null), otherwise:
  b) Use values from PlayFabSettings.

2) If isApiInstance == true, then:
  a) Use values from request.AuthenticationContext (if it is not null), otherwise:
  b) Use values from this.AuthenticationContext (if it is not null), otherwise:
  c) Use values from PlayFabSettings

Using IsClientLoggedIn is ok if you want to have those functions, but please make sure you are using them right: instance method (this.IsClientLoggedIn()) should be called only when isApiInstance == true. <ClassName>.IsClientLoggedIn() must be called only when isApiInstance == false. Please make sure you are calling the right static method from the right class. If this template is meant to be used to generate each API class then please be careful to call specific static methods as you may end up calling it from the wrong API class.

Vicodex:
I added some important comments. @Melih Ceyhan, please make sure to generate the SDK, build it and run the tests to ensure that it works with your changes.

targets/unity-v2/make.js
melihceyhan: We check this control in PlayFabHttp.MakeApiCall method. you said me that to move that there. So I think we should not add isApiInstance === true, we should remove  isApiInstance === false condition also. What is our final decision what should I do?

melihceyhan: I removed it for now.  After your review if you want I can add it again.

melihceyhan:
Thanks @Vicodex! I did required changes. Please let me know if have any misunderstanding. Also I ran all unit tests again they passed.

targets/unity-v2/templates/InstanceAPI.cs.ejs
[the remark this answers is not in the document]
Vicodex: Ok. Melih, please feel free to make it a public field.

---
In reply to: [262254578](https://github.com/PlayFab/SDKGenerator/pull/420#discussion_r262254578) [](ancestors = 262254578)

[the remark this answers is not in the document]
melihceyhan: Thats ok. Done

targets/unity-v2/source/Shared/Internal/PlayFabHttp/PlayFabHTTP.cs
```diff
@@ -186,17 +186,35 @@ public static void SimplePostCall(string fullUrl, byte[] payload, Action<byte[]>
...
+                }
+                if(developerSecretKey == null && authenticationContext != null && !string.IsNullOrEmpty(authenticationContext.DeveloperSecretKey))
+                {
+                    developerSecretKey = authenticationContext.DeveloperSecretKey;
+                }
+                if(developerSecretKey == null)
+                {
+                    developerSecretKey = PlayFabSettings.DeveloperSecretKey;
+                }
+
```
Vicodex: Please remove these tabs

melihceyhan: removed

targets/unity-v2/make.js
```diff
@@ -378,24 +417,34 @@ function getAuthParams(apiCall) {
     return "AuthType.None";
 }
 
-function getRequestActions(tabbing, apiCall) {
-    if (apiCall.name === "GetEntityToken")
+function getRequestActions(tabbing, apiCall, isApiInstance = false) {
+    if (apiCall.name === "GetEntityToken" && isApiInstance === false)
         return tabbing + "AuthType authType = AuthType.None;\n" +
             "#if !DISABLE_PLAYFABCLIENT_API\n" +
             tabbing + "if (authType == AuthType.None && PlayFabClientAPI.IsClientLoggedIn())\n" +
```
Vicodex: >PlayFabClientAPI. [](start = 56, length = 17)

Please remove "PlayFabClientAPI." prefix here. This code will be used to run any other API template, so we want to keep just the static method name here.

targets/unity-v2/make.js
```diff
@@ -378,24 +417,34 @@ function getAuthParams(apiCall) {
...
             "#endif\n" +
             "#if ENABLE_PLAYFABSERVER_API || ENABLE_PLAYFABADMIN_API || UNITY_EDITOR\n" +
-            tabbing + "if (authType == AuthType.None && !string.IsNullOrEmpty(PlayFabSettings.DeveloperSecretKey))\n" +
+            tabbing + "if (authType == AuthType.None && !string.IsNullOrEmpty(request.AuthenticationContext.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey))\n" +
             tabbing + "    authType = AuthType.DevSecretKey;\n" +
             "#endif\n";
+	if (apiCall.name === "GetEntityToken" && isApiInstance === true)
+        return tabbing + "AuthType authType = AuthType.None;\n" +
+            "#if !DISABLE_PLAYFABCLIENT_API\n" +
+            tabbing + "if (authType == AuthType.None && authenticationContext.IsClientLoggedIn())\n" +
```
Vicodex: >authenticationContext. [](start = 56, length = 22)

Please remove "authenticationContext." prefix here and call just the method IsClientLoggedIn on this API instance.

targets/unity-v2/make.js
Vicodex: Ok, my bad. I see that some of the logic that i wanted is indeed already moved to MakeApiCall. Please see my comment there.

---
In reply to: [262365219](https://github.com/PlayFab/SDKGenerator/pull/420#discussion_r262365219) [](ancestors = 262365219)

targets/unity-v2/make.js
```diff
@@ -378,24 +417,34 @@ function getAuthParams(apiCall) {
...
+            "#endif\n" +
+            "#if ENABLE_PLAYFABSERVER_API || ENABLE_PLAYFABADMIN_API || UNITY_EDITOR\n" +
+            tabbing + "if (authType == AuthType.None && !string.IsNullOrEmpty(request.AuthenticationContext.DeveloperSecretKey ?? (authenticationContext.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey)))\n" +
+            tabbing + "    authType = AuthType.DevSecretKey;\n" +
+            "#endif\n";		
 
     if (apiCall.result === "LoginResult" || apiCall.request === "RegisterPlayFabUserRequest")
         return tabbing + "request.TitleId = request.TitleId ?? PlayFabSettings.TitleId;\n";
-    if (apiCall.auth === "SessionTicket")
+    if (apiCall.auth === "SessionTicket" && isApiInstance === false)
```
Vicodex: >SessionTicket [](start = 26, length = 13)

Ok, you are right: for this:
if (apiCall.auth === "SessionTicket") block let's not make a separation by isApiInstance as they will be the same. Just keep one single if (apiCall.auth === "SessionTicket") block which will have this:
return tabbing + "if (!IsClientLoggedIn()) throw new PlayFabException(PlayFabExceptionCode.NotLoggedIn,\"Must be logged in to call this method\");\n";

Both static and instance API classes have IsClientLoggedIn method so it should work.

targets/unity-v2/make.js
melihceyhan: For example PlayFabAuthenticationAPI has no IsClientLoggedInMethod what should we do? Replace PlayFabClientAPI with PlayFabHttp?

targets/unity-v2/make.js
melihceyhan: All api instances has no IsClientLoggedInMethod for example: PlayFabAuthenticationInstanceAPI same question with above one

Vicodex:
>             }

I would like to use authenticationContext.ClientSessionTicket and authenticationContext.EntityToken here for cases AuthType.LoginSession and AuthType.EntityToken correspondingly, when authenticationContext is not null (it is a new argument that we pass to MakeApiCall). This will be the intended priority logic:

1) Request.AuthenticationContext
2) authenticationContext (if passed as argument to MakeApiCall)
3) PlayFabSettings (which is what really comes from transport)

---
Refers to: targets/unity-v2/source/Shared/Internal/PlayFabHttp/PlayFabHTTP.cs:261 in 7d22ac5. [](commit_id = 7d22ac55ee1af8fc7622a1c4cd56e1f6c39ac82d, deletion_comment = False)

targets/unity-v2/make.js
Vicodex: Ah, i see. Ok, let's do what you did with developerSecretKey below, something like:

!string.IsNullOrEmpty(request.AuthenticationContext.ClientSessionKey ?? (authenticationContext.ClientSessionKey ?? PlayFabSettings.ClientSessionKey)

---
In reply to: [262696786](https://github.com/PlayFab/SDKGenerator/pull/420#discussion_r262696786) [](ancestors = 262696786)

targets/unity-v2/make.js
Vicodex: Please try to do what you did with developer secret key below: 

!string.IsNullOrEmpty(request.AuthenticationContext.ClientSessionKey ?? PlayFabSettings.ClientSessionKey)



---
In reply to: [262695893](https://github.com/PlayFab/SDKGenerator/pull/420#discussion_r262695893) [](ancestors = 262695893)

targets/unity-v2/make.js
Vicodex: Since you showed me that we can't simply call IsClientLoggedIn everywhere, we may have to keep the separation by isApiInstance. Please use direct checks of ClientSessionTicked fields here instead of callign IsClientLogged in.

---
In reply to: [262695780](https://github.com/PlayFab/SDKGenerator/pull/420#discussion_r262695780) [](ancestors = 262695780)

targets/unity-v2/templates/InstanceAPI.cs.ejs
```diff
@@ -0,0 +1,83 @@
...
+<% }%>
+<% if (hasEntityTokenOptions) { %>
+        /// <summary>
+        /// Check to See if the entity is logged in.
+        /// </summary>
+        public bool IsEntityLoggedIn()
+        {
+            return authenticationContext == null ? false : authenticationContext.IsEntityLoggedIn();
+        }
+<% }%>
```
Vicodex: In this case, you can remove these methods if they are not used anywhere.

targets/unity-v2/make.js
melihceyhan: We can use it there because it is for session ticket and it is used for ClientApi i think. Am i wrong?

targets/unity-v2/templates/InstanceAPI.cs.ejs
melihceyhan: It's up to your decision on if (apiCall.auth === "SessionTicket").  As I asked you on previous comment

targets/unity-v2/make.js
Vicodex: I think this check is used for any API call that uses "Session Ticket" authentication (instead of, for example, "Secret Key" authentication). For "Session Ticket" authentication a ClientSessionTicket value will be used, which is why i think we can simply check it directly, instead of calling IsClientLoggedIn (which is doing exactly that - it checks if client session ticket value was set).

---
In reply to: [262704021](https://github.com/PlayFab/SDKGenerator/pull/420#discussion_r262704021) [](ancestors = 262704021)

melihceyhan: I commit it with single if block please let me know if i should revert it to double if blocks

melihceyhan: sorry there is a little confusion for me :) as a final decision what should I do

Vicodex: Give me a moment, please. I will write it :)

---
In reply to: [262709936](https://github.com/PlayFab/SDKGenerator/pull/420#discussion_r262709936) [](ancestors = 262709936)

targets/unity-v2/make.js
```diff
@@ -378,24 +417,32 @@ function getAuthParams(apiCall) {
...
 }
 
-function getRequestActions(tabbing, apiCall) {
-    if (apiCall.name === "GetEntityToken")
+function getRequestActions(tabbing, apiCall, isApiInstance = false) {
+    if (apiCall.name === "GetEntityToken" && isApiInstance === false)
         return tabbing + "AuthType authType = AuthType.None;\n" +
             "#if !DISABLE_PLAYFABCLIENT_API\n" +
-            tabbing + "if (authType == AuthType.None && PlayFabClientAPI.IsClientLoggedIn())\n" +
+            tabbing + "if (authType == AuthType.None && (request.AuthenticationContext != null ? request.AuthenticationContext.IsClientLoggedIn() : PlayFabClientAPI.IsClientLoggedIn()))\n" +
```
Vicodex: >(request.AuthenticationContext != null ? request.AuthenticationContext.IsClientLoggedIn() : PlayFabClientAPI.IsClientLoggedIn()) [](start = 56, length = 128)

Replace

(request.AuthenticationContext != null ? request.AuthenticationContext.IsClientLoggedIn() : PlayFabClientAPI.IsClientLoggedIn())

with

!string.IsNullOrEmpty(request.AuthenticationContext.ClientSessionTicket ?? PluginManager.GetPlugin<IPlayFabTransportPlugin>(PluginContract.PlayFab_Transport).AuthKey)

targets/unity-v2/make.js
```diff
@@ -378,24 +417,32 @@ function getAuthParams(apiCall) {
...
             "#endif\n" +
             "#if ENABLE_PLAYFABSERVER_API || ENABLE_PLAYFABADMIN_API || UNITY_EDITOR\n" +
-            tabbing + "if (authType == AuthType.None && !string.IsNullOrEmpty(PlayFabSettings.DeveloperSecretKey))\n" +
+            tabbing + "if (authType == AuthType.None && !string.IsNullOrEmpty(request.AuthenticationContext.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey))\n" +
             tabbing + "    authType = AuthType.DevSecretKey;\n" +
             "#endif\n";
+	if (apiCall.name === "GetEntityToken" && isApiInstance === true)
+        return tabbing + "AuthType authType = AuthType.None;\n" +
+            "#if !DISABLE_PLAYFABCLIENT_API\n" +
+            tabbing + "if (authType == AuthType.None && (request.AuthenticationContext != null ? request.AuthenticationContext.IsClientLoggedIn() : (authenticationContext != null ? authenticationContext.IsClientLoggedIn() : PlayFabClientAPI.IsClientLoggedIn())))\n" +
```
Vicodex: >(request.AuthenticationContext != null ? request.AuthenticationContext.IsClientLoggedIn() : (authenticationContext != null ? authenticationContext.IsClientLoggedIn() : PlayFabClientAPI.IsClientLoggedIn())) [](start = 56, length = 205)

Replace 

(request.AuthenticationContext != null ? request.AuthenticationContext.IsClientLoggedIn() : (authenticationContext != null ? authenticationContext.IsClientLoggedIn() : PlayFabClientAPI.IsClientLoggedIn()))

with

!string.IsNullOrEmpty(request.AuthenticationContext.ClientSessionTicket ?? (authenticationContext.ClientSessionTicket ?? PluginManager.GetPlugin<IPlayFabTransportPlugin>(PluginContract.PlayFab_Transport).AuthKey))

targets/unity-v2/make.js
```diff
@@ -378,24 +417,32 @@ function getAuthParams(apiCall) {
...
+            "#endif\n" +
+            "#if ENABLE_PLAYFABSERVER_API || ENABLE_PLAYFABADMIN_API || UNITY_EDITOR\n" +
+            tabbing + "if (authType == AuthType.None && !string.IsNullOrEmpty(request.AuthenticationContext.DeveloperSecretKey ?? (authenticationContext.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey)))\n" +
+            tabbing + "    authType = AuthType.DevSecretKey;\n" +
+            "#endif\n";		
 
     if (apiCall.result === "LoginResult" || apiCall.request === "RegisterPlayFabUserRequest")
         return tabbing + "request.TitleId = request.TitleId ?? PlayFabSettings.TitleId;\n";
     if (apiCall.auth === "SessionTicket")
         return tabbing + "if (!IsClientLoggedIn()) throw new PlayFabException(PlayFabExceptionCode.NotLoggedIn,\"Must be logged in to call this method\");\n";
```
Vicodex: >if (!IsClientLoggedIn()) [](start = 26, length = 24)

Replace 

if (!IsClientLoggedIn())

with two different cases:

when isApiInstance === true:

if (string.IsNullOrEmpty(request.AuthenticationContext.ClientSessionTicket ?? (authenticationContext.ClientSessionTicket ?? PluginManager.GetPlugin<IPlayFabTransportPlugin>(PluginContract.PlayFab_Transport).AuthKey)))

when isApiInstance === false:

if (string.IsNullOrEmpty(request.AuthenticationContext.ClientSessionTicket ?? PluginManager.GetPlugin<IPlayFabTransportPlugin>(PluginContract.PlayFab_Transport).AuthKey))

targets/unity-v2/make.js
Vicodex: Yeah it will be double blocks, see my other comment with direct replacement code

---
In reply to: [262709039](https://github.com/PlayFab/SDKGenerator/pull/420#discussion_r262709039) [](ancestors = 262709039)

targets/unity-v2/templates/InstanceAPI.cs.ejs
Vicodex: Yeah if my replacements work, please remove these methods (if they are not used)

---
In reply to: [262707701](https://github.com/PlayFab/SDKGenerator/pull/420#discussion_r262707701) [](ancestors = 262707701)

targets/unity-v2/make.js
melihceyhan: isApiInstance === false case got null pointer exception when ReportDeviceInfo called so I keep instanceapi.ejs.cs and isApiInstance  === false case as :"if (!IsClientLoggedIn()) . Committed last version

Vicodex:
Looks good now, thanks Melih! Please make sure that tests pass and everything is working :) Good job!

---
In reply to: [469567340](https://github.com/PlayFab/SDKGenerator/pull/420#issuecomment-469567340) [](ancestors = 469567340)

PR #422, opened 2019-02-28
Minor unreal update tweaks

pgilmorepf:
Remove 4.18
Templatize the WhitelistPlatforms, so we can add the consoles easily in the next checkin.

### DIFF
--- a/targets/UnrealMarketplacePlugin/make.js
+++ b/targets/UnrealMarketplacePlugin/make.js
@@ -15,7 +15,8 @@ var copyright =
 
 exports.makeCombinedAPI = function (apis, sourceDir, baseApiOutputDir) {
     // The list of current supported UE versions - Intended to be the latest 3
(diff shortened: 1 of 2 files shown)


targets/UnrealMarketplacePlugin/make.js
```diff
@@ -15,7 +15,8 @@ var copyright =
 
 exports.makeCombinedAPI = function (apis, sourceDir, baseApiOutputDir) {
     // The list of current supported UE versions - Intended to be the latest 3
-    var ueTargetVersions = ["4.18", "4.19", "4.20", "4.21"];
+    var ueTargetVersions = ["4.19", "4.20", "4.21"];
+    var ueWhitelistPlatforms = '[ "Win64", "Win32", "Mac", "IOS", "Android" ]'; // TODO: Add PS4, Switch, and XBox once the testing stories are done
```
Vicodex: >XBox [](start = 110, length = 4)

nit: Xbox

PR #425, opened 2019-03-02
Toddbell/unreal inherit

toddbellMSFT:
Possible answer to fixing an Unreal issue


toddbellMSFT:
This is definitley not the correct answer to the issue. This will be fixed at a later date.

PR #426, opened 2019-03-02
Toddbell/unreal rename

toddbellMSFT:
Simple possible fix to Unreal SDK


toddbellMSFT:
We will won't fix the issue until after GDC.

PR #430, opened 2019-03-04
small fix

melihceyhan:
Is that what we need. Please let me know If I am doing it wrong way


Vicodex:
This is great, but we also need a fix for this:
[https://github.com/PlayFab/SDKGenerator/blob/master/targets/csharp/templates/APIInstance.cs.ejs#L82](https://github.com/PlayFab/SDKGenerator/blob/master/targets/csharp/templates/APIInstance.cs.ejs#L82)

It should be using authentication context instead of PlayFabSettings in that template (i.e. make it how you just did in this template for Unity).

melihceyhan:
Thanks @Vicodex  i did it

Vicodex:
Please hold on this as we may need to fix a few more things here for C#, in getRequestActions(...) JS function

PR #431, opened 2019-03-04
Add support for local debugging of Azure Functions in Unity and C# SDKs

nima-ap:
The following changes in this PR facilitate the ability to have local debugging with the Azure Functions capabilities added to the SDK. 
These changes are implemented for the Unity and C# SDKs and are implemented through various changes in a dozen files. 
To make this work, a few changes have also been made to the global API.cs.ejs templates for these two SDKs to allow conditional `Property` vs. `Field` generation in classes based on which class it is that is being generated as we need Properties for some deserializations to work properly on the Azure Function side which uses the C# SDK.

### DIFF
--- a/targets/csharp/templates/PlayFabSettings.cs.ejs
+++ b/targets/csharp/templates/PlayFabSettings.cs.ejs
@@ -24,25 +24,6 @@ namespace PlayFab
         public static string DeveloperSecretKey = null;
         /// <summary> You must set this value for PlayFabSdk to work properly (Found in the Game Manager for your title, at the PlayFab Website) </summary>
         public static string TitleId;
-        /// <summary> Placeholder for the Local Debugging URI. </summary>
-        private static string _localApiServer;
-        /// <summary> Must be manually set by user on .NETStandard 1.1 and .NETPortable for local debugging to work properly </summary>
-        public static string LocalApiServer 
-        {
-            get 
-            {
-#if NET45 || NETCOREAPP2_0
-                return _localApiServer ?? PlayFabUtil.GetLocalSettingsFileProperty("LocalApiServer");
-#else
-                return _localApiServer;
-#endif
-            }
-            set
-            {
-                _localApiServer = value;
-            }
-        }
-
         public static ErrorCallback GlobalErrorHandler;
 <% if (hasClientOptions) {
 %>        /// <summary> Set this to the appropriate AD_TYPE_X constant below </summary>
--- a/targets/csharp/templates/PlayFabSettings.cs.ejs
+++ b/targets/csharp/templates/PlayFabSettings.cs.ejs
@@ -13,7 +13,7 @@ namespace PlayFab
         };
 
         /// <summary> This is only for customers running a private cluster.  Generally you shouldn't touch this </summary>
-        public static string ProductionEnvironmentUrl = ".playfabapi.com";
+        public static string ProductionEnvironmentUrl = "playfabapi.com";
         /// <summary> The name of a customer vertical. This is only for customers running a private cluster.  Generally you shouldn't touch this </summary>
         public static string VerticalName = <%- getVerticalNameDefault() %>;
         /// <summary> Session token for Entity API. Auto-Populated by GetEntityToken method. </summary>
@@ -24,6 +24,27 @@ namespace PlayFab
         public static string DeveloperSecretKey = null;
         /// <summary> You must set this value for PlayFabSdk to work properly (Found in the Game Manager for your title, at the PlayFab Website) </summary>
         public static string TitleId;
+		/// <summary> Placeholder for the Local Debugging URI. </summary>
+		private static string _debugUri;
+		/// <summary> Must be manually set by user on .NETStandard 1.1 and .NETPortable for local debugging to work properly </summary>
+        public static string DebugUri 
+        {
+            get 
+            {
+#if NET45 || NETCOREAPP2_0
+                return PlayFabUtil.GetLocalSettingsFileProperty("ExecuteFunctionDebugUri");
+#else
+                return _debugUri;
+#endif
+            }
+#if !NET45 && !NETCOREAPP2_0
+            set
+            {
+                _debugUri = value;
+            }
+#endif
+        }
+
         public static ErrorCallback GlobalErrorHandler;
 <% if (hasClientOptions) {
 %>        /// <summary> Set this to the appropriate AD_TYPE_X constant below </summary>
@@ -40,17 +61,18 @@ namespace PlayFab
         public static string GetFullUrl(string apiCall, Dictionary<string, string> getParams, PlayFabApiSettings apiSettings = null)
         {
             StringBuilder sb = new StringBuilder(1000);
-        
+
             var baseUrl = apiSettings?.ProductionEnvironmentUrl ?? ProductionEnvironmentUrl;
             if (!baseUrl.StartsWith("http"))
             {
+                sb.Append("https://");
+                if ((apiSettings?.TitleId ?? TitleId) != null)
+                {
+                    sb.Append(apiSettings?.TitleId ?? TitleId).Append(".");
+                }
                 if ((apiSettings?.VerticalName ?? VerticalName) != null)
                 {
-                    sb.Append("https://").Append(apiSettings?.VerticalName ?? VerticalName);
-                }
-                else
-                {
-                    sb.Append("https://").Append(apiSettings?.TitleId ?? TitleId);
+                    sb.Append(apiSettings?.VerticalName ?? VerticalName).Append(".");
                 }
             }
                 
--- a/targets/csharp/templates/PlayFabSettings.cs.ejs
+++ b/targets/csharp/templates/PlayFabSettings.cs.ejs
@@ -25,22 +25,22 @@ namespace PlayFab
         /// <summary> You must set this value for PlayFabSdk to work properly (Found in the Game Manager for your title, at the PlayFab Website) </summary>
         public static string TitleId;
         /// <summary> Placeholder for the Local Debugging URI. </summary>
-        private static string _debugUri;
+        private static string _localApiUri;
         /// <summary> Must be manually set by user on .NETStandard 1.1 and .NETPortable for local debugging to work properly </summary>
-        public static string DebugUri 
+        public static string LocalApiUri 
         {
             get 
             {
 #if NET45 || NETCOREAPP2_0
-                return PlayFabUtil.GetLocalSettingsFileProperty("ExecuteFunctionDebugUri");
+                return PlayFabUtil.GetLocalSettingsFileProperty("LocalApiUri");
 #else
-                return _debugUri;
+                return _localApiUri;
 #endif
             }
 #if !NET45 && !NETCOREAPP2_0
             set
             {
-                _debugUri = value;
+                _localApiUri = value;
             }
 #endif
         }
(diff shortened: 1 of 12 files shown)


targets/csharp/templates/PlayFabSettings.cs.ejs
```diff
@@ -24,6 +24,27 @@ namespace PlayFab
         public static string DeveloperSecretKey = null;
         /// <summary> You must set this value for PlayFabSdk to work properly (Found in the Game Manager for your title, at the PlayFab Website) </summary>
         public static string TitleId;
+		/// <summary> Placeholder for the Local Debugging URI. </summary>
```
pgilmorepf: You're using tab characters here.
Please replace all tab characters with 4 spaces.

targets/csharp/templates/PlayFabSettings.cs.ejs
```diff
@@ -24,6 +24,27 @@ namespace PlayFab
...
+		private static string _debugUri;
+		/// <summary> Must be manually set by user on .NETStandard 1.1 and .NETPortable for local debugging to work properly </summary>
+        public static string DebugUri 
+        {
+            get 
+            {
+#if NET45 || NETCOREAPP2_0
+                return PlayFabUtil.GetLocalSettingsFileProperty("ExecuteFunctionDebugUri");
```
pgilmorepf: I think you should do:
return _debugUri ??  PlayFabUtil.GetLocalSettingsFileProperty("ExecuteFunctionDebugUri");
Basically, you are ignoring _debugUri if it's Net4x or higher, which seems odd.

targets/csharp/templates/PlayFabUtil.cs.ejs
```diff
@@ -81,5 +83,78 @@ namespace PlayFab
...
+#if NET45 || NETCOREAPP2_0
+        [ThreadStatic]
+        private static StringBuilder _sb;
+        /// <summary>
+        /// A threadsafe way to block and load a text file
+        /// 
+        /// Load a text file, and return the file as text.
+        /// Used for small (usually json) files.
+        /// </summary>
+        public static string ReadAllFileText(string filename)
```
pgilmorepf: This utility was just removed.
It looks like you did a bad merge.
Please look through this diff and make sure you are only adding your own changes.

targets/csharp/templates/PlayFabSettings.cs.ejs
```diff
@@ -24,6 +24,27 @@ namespace PlayFab
         public static string DeveloperSecretKey = null;
         /// <summary> You must set this value for PlayFabSdk to work properly (Found in the Game Manager for your title, at the PlayFab Website) </summary>
         public static string TitleId;
+		/// <summary> Placeholder for the Local Debugging URI. </summary>
+		private static string _debugUri;
+		/// <summary> Must be manually set by user on .NETStandard 1.1 and .NETPortable for local debugging to work properly </summary>
+        public static string DebugUri
```
pgilmorepf: I'm concerned about this name.
The name implies a much more generic purpose than it actually serves. It's only used by one method, and can't be used by another.
Either we need to fix it so that the URI is shared by any potential CloudScript API call (or relevant thing), or we need to   rename it to match the fact that it's very specific to just one method.

targets/csharp/templates/PlayFabUtil.cs.ejs
nima-ap: So we are aware of the fact that it was removed but since we need to read a config file in net45 and core22 scenarios, we decided to improve the method by incorporating more `using` statements to dispose of disposables and bring it back so that we can actually read our config file.

targets/csharp/templates/PlayFabSettings.cs.ejs
nima-ap: Maybe @MGudgin can comment on this better cause I recall having the conversation with him and we decided to always pull the URI live from the file when requested to reflect potential changes in the file in the middle of a session, instead of caching it.

targets/csharp/templates/PlayFabSettings.cs.ejs
nima-ap: 🤦‍♂️ I thought I fixed these, but I guess not. I'll look into finding a way to do a global check on the project so that I don't miss any more of these.

targets/unity-v2/make.js
```diff
@@ -147,6 +147,7 @@ function makeApi(api, sourceDir, apiOutputDir) {
         generateApiSummary: generateApiSummary,
         getDeprecationAttribute: getDeprecationAttribute,
         getRequestActions: getRequestActions,
+        getCustomApiLogic: getCustomApiLogic,
```
nima-ap: I'd like to bring something to your attention regarding this.
In Unity it seems like you already had some logic built for custom behavior, using `getCustomApiFunction`. For C# however, this is not the case. 
However, your unity `getCustomApiFunction` placeholder is inserted _after_ the api call in the Unity case. In our case, we need it to be _before_ the API call.
How does the SDK team feel about if we renamed these two functions to `getPreAPICustomActions` and `getPostAPICustomActions` to keep the naming a bit more consistent here? We could even have the `pre` and `post` be a parameter of a single function like `getApiCustomActions` or something of the sort. 
@pgilmorepf

targets/csharp/templates/PlayFabSettings.cs.ejs
nima-ap: How about `ExecuteFunctionLocalUri` and `_executeLocalFunctionUri`?

targets/csharp/templates/PlayFabUtil.cs.ejs
```diff
@@ -81,5 +83,78 @@ namespace PlayFab
...
+
+        internal static string GetLocalSettingsFileProperty(string propertyKey)
+        {
+            string envFileContent = null;
+            string currDir = Directory.GetCurrentDirectory();
+            string currDirEnvFile = Path.Combine(currDir, _localSettingsFileName);
+
+            if (File.Exists(currDirEnvFile))
+            {
+                envFileContent = ReadAllFileText(currDirEnvFile);
```
nima-ap: @pgilmorepf  This and line 144 are where we are using `ReadAllfileText` and hence the reason why we brought it back.

targets/csharp/templates/PlayFabSettings.cs.ejs
pgilmorepf: My concern here is that we have the interface to set the variable in code, but if the env-var is set, the local configuration will be ignored. If you have a local configuration, it should be obeyed. You're setting up a scenario where a user is unable to accomplish what they want to do, and the fix seems really quite simple.

targets/csharp/templates/PlayFabUtil.cs.ejs
pgilmorepf: I feel like if the function exists, it should be available for all platforms possible, not just a very limited subset. That imbalanced limitation was exactly why we removed it.

targets/csharp/templates/PlayFabSettings.cs.ejs
pgilmorepf: I'm happy with a limited name that you propose.
I'd be slightly more happy with a generic name, if that variable was able to serve a more generic purpose.

targets/csharp/templates/PlayFabUtil.cs.ejs
nima-ap: Okay well Net Portable doesn't provide IO and we're not in favor of throwing away the whole concept of reading out config file because we can't do IO on portable or standard 1.1.
I already spoke to you about this. You said allow the user to use these features on 45 and 22 and on portable we'll remove it since we can't do IO. This is exactly what that is.

targets/csharp/templates/PlayFabSettings.cs.ejs
nima-ap: The user's primary interface with changing this variable was meant to be through the VS Code extension and not to have to write `PlayFabSettings.DebugUri = xxxxx` in their code and that was a design decision made from the beginning of this feature's development.
This is more-so a placeholder for the rare case that the user is not able to go through the main interface, since they are are standard1.1 or Portable, and hence we decided to allow such 'guru' users who want to debug their code on an android device (per say) to be able to do it, as per our previous conversation in your office.

nima-ap: I wouldn't necessarily consider this a 'bug' to 'fix' as it's a design decision and we wanted to be able to pull the URL live from the config file instead of cache it at the start of every run-time.

targets/unity-v2/make.js
pgilmorepf: I agree with everything you're saying here, both your methodology and your proposal.

targets/csharp/templates/PlayFabUtil.cs.ejs
pgilmorepf: Yes, I saw this. And I understand your need.

targets/csharp/templates/PlayFabUtil.cs.ejs
pgilmorepf: Let's call this one ok. Clearly it's not a choice for you.
I'll talk to Deb about redoing the cross-platform testing.

pgilmorepf: As discussed, we're looping people in and discussing the cross-platform testing.

targets/csharp/templates/PlayFabSettings.cs.ejs
MGudgin: Actually, there is no interface to set the variable in code in the net45/netstandard20  case the way the code is written today. That said, I'm find with having the setter in all cases and doing the null coalesce in the net45/netstandard20 case.

---
In reply to: [262691931](https://github.com/PlayFab/SDKGenerator/pull/431#discussion_r262691931) [](ancestors = 262691931)

targets/csharp/templates/PlayFabSettings.cs.ejs
MGudgin: If we go with ExecuteFunctionLocalUri as the public property the field should be _executeFunctionLocalUri, no?

---
In reply to: [262266567](https://github.com/PlayFab/SDKGenerator/pull/431#discussion_r262266567) [](ancestors = 262266567)

nima-ap:
@pgilmorepf Please re-review this as a rebase on your current master branch has been done and there were some merge conflicts that were manually resolved by me today.

targets/csharp/templates/PlayFabUtil.cs.ejs
nima-ap: Reverting changes to C# SDK related to local-debugging for GDC.

nima-ap: Resolving as changes reverted.

targets/csharp/templates/PlayFabUtil.cs.ejs
nima-ap: Resolving as changes reverted.

targets/csharp/templates/PlayFabUtil.cs.ejs
nima-ap: Resolving as changes reverted.

targets/unity-v2/make.js
nima-ap: Resolving this for now as I realized you use `getCustomApiFunction` **after the whole method** and not **after the request**. My original suggestion was in regards to **pre and post request actions** and so we'll need to perhaps figure out a better name for `getCustomApiLogic` to avoid it from being confused with `getCustomApiFunction`.

nima-ap:
All C# Changes to SDK that include I/O have been reverted.

targets/unity-v2/make.js
pgilmorepf: I agree with everything you're saying here, both your methodology and your proposal.
Come up with good and appropriate names, just make sure you test it.

nima-ap: Let's put this in backlog for post GDC then.

PR #432, opened 2019-03-06
Milestone #29: sdkgen unity bug fix

melihceyhan:
Hey @Vicodex  I got an null pointer exception error while testing so I made this change. On the other hand I got missing or invalid X-SecretKey Http error which didnt occur yesterday. Is there a problem with these values : https://github.com/PlayFab/pf-main/blob/master/tools/SDKBuildScripts/testTitleData.json

### DIFF
--- a/targets/unity-v2/make.js
+++ b/targets/unity-v2/make.js
@@ -430,16 +430,21 @@ function getRequestActions(tabbing, apiCall, isApiInstance = false) {
             tabbing + "    authType = AuthType.LoginSession;\n" +
             "#endif\n" +
             "#if ENABLE_PLAYFABSERVER_API || ENABLE_PLAYFABADMIN_API || UNITY_EDITOR\n" +
-            tabbing + "if (authType == AuthType.None && !string.IsNullOrEmpty(request.AuthenticationContext.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey))\n" +
+            tabbing + "string developerSecretKey = null;\n" +
+            tabbing + "if (request.AuthenticationContext != null && !string.IsNullOrEmpty(request.AuthenticationContext.DeveloperSecretKey))\n" +
+            tabbing + "    developerSecretKey = request.AuthenticationContext.DeveloperSecretKey;\n" +
+            tabbing + "if (developerSecretKey == null)\n" +
+            tabbing + "    developerSecretKey = PlayFabSettings.DeveloperSecretKey;\n" +
+            tabbing + "if (authType == AuthType.None && !string.IsNullOrEmpty(developerSecretKey))\n" +
             tabbing + "    authType = AuthType.DevSecretKey;\n" +
             "#endif\n";
 	if (apiCall.name === "GetEntityToken" && isApiInstance === true)
         return tabbing + "AuthType authType = AuthType.None;\n" +
             "#if !DISABLE_PLAYFABCLIENT_API\n" +
+			tabbing + "string clientSessionTicket = null;\n" +
             tabbing + "if (request.AuthenticationContext != null && !string.IsNullOrEmpty(request.AuthenticationContext.ClientSessionTicket))\n" +
             tabbing + "    clientSessionTicket = request.AuthenticationContext.ClientSessionTicket;\n" +
             tabbing + "if (clientSessionTicket == null && authenticationContext != null && !string.IsNullOrEmpty(authenticationContext.ClientSessionTicket))\n" +
-            tabbing + "string clientSessionTicket = null;\n" +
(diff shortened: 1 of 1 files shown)


Vicodex:
Hey Melih, the problem with invalid developer secret key is a separate issue, we had an incident yesterday and had to invalidate the key. You will get a new test title data json file soon. 

Now, about this change: you are returning back explicit call of PlayFabClientAPI class, which was the reason why we replaced it. Can you please tell which part is reason of NullPointerException that you are getting? Let's get to the root of it, because the code that you are trying to change should be working by design.

melihceyhan:
Thanks for clarification!

!string.IsNullOrEmpty(request.AuthenticationContext.ClientSessionTicket ?? PluginManager.GetPlugin<IPlayFabTransportPlugin>(PluginContract.PlayFab_Transport).AuthKey)  here is the part which I removed request.AuthenticationContext.ClientSessionTicket is null and also PluginManager.GetPlugin<IPlayFabTransportPlugin>(PluginContract.PlayFab_Transport) is null so calling AuthKey property from it causes null pointer exception

melihceyhan:
if we call PlayFabClientAPI.IsClientLoggedIn() it works smt like transport.IsInitialized. I think this prevents null pointer exception. But I am not sure

Vicodex:
Can you please debug it and see if this method in PluginManager works as expected: CreatePlayFabTransportPlugin

melihceyhan:
Hey Victor,
Sorry it is my bad. 
PluginManager.GetPlugin(PluginContract.PlayFab_Transport) is not null
request.AuthenticationContext is null so it gets null pointer exception.

I add checks for it and commit

targets/unity-v2/make.js
```diff
@@ -431,7 +436,14 @@ function getRequestActions(tabbing, apiCall, isApiInstance = false) {
 	if (apiCall.name === "GetEntityToken" && isApiInstance === true)
         return tabbing + "AuthType authType = AuthType.None;\n" +
             "#if !DISABLE_PLAYFABCLIENT_API\n" +
-            tabbing + "if (authType == AuthType.None && !string.IsNullOrEmpty(request.AuthenticationContext.ClientSessionTicket ?? (authenticationContext.ClientSessionTicket ?? PluginManager.GetPlugin<IPlayFabTransportPlugin>(PluginContract.PlayFab_Transport).AuthKey)))\n" +
+            tabbing + "if (request.AuthenticationContext != null && !string.IsNullOrEmpty(request.AuthenticationContext.ClientSessionTicket))\n" +
+            tabbing + "    clientSessionTicket = request.AuthenticationContext.ClientSessionTicket;\n" +
+            tabbing + "if (clientSessionTicket == null && authenticationContext != null && !string.IsNullOrEmpty(authenticationContext.ClientSessionTicket))\n" +
+            tabbing + "string clientSessionTicket = null;\n" +
```
Vicodex: >tabbing + "string clientSessionTicket = null;\n" + [](start = 12, length = 50)

This line with declaration:
string clientSessionTicket = null
should be moved 3 lines above, right?

Vicodex:
>             tabbing + "if (authType == AuthType.None && !string.IsNullOrEmpty(request.AuthenticationContext.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey))\n" +

This may have the same issue with null reference exception. Please add code that checks if request.AuthenticationContext is not null

---
Refers to: targets/unity-v2/make.js:433 in 643adca. [](commit_id = 643adcab5be322f159445e5e2540c3c6eb4b6e95, deletion_comment = False)

Vicodex:
>             tabbing + "if (authType == AuthType.None && !string.IsNullOrEmpty(request.AuthenticationContext.DeveloperSecretKey ?? (authenticationContext.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey)))\n" +

This may have the same problem

---
Refers to: targets/unity-v2/make.js:450 in 643adca. [](commit_id = 643adcab5be322f159445e5e2540c3c6eb4b6e95, deletion_comment = False)

Vicodex:
>             tabbing + "if (authType == AuthType.None && !string.IsNullOrEmpty(request.AuthenticationContext.DeveloperSecretKey ?? (authenticationContext.DeveloperSecretKey ?? PlayFabSettings.DeveloperSecretKey)))\n" +

This may have the same problem

---
Refers to: targets/unity-v2/make.js:450 in 643adca. [](commit_id = 643adcab5be322f159445e5e2540c3c6eb4b6e95, deletion_comment = False)

targets/unity-v2/make.js
melihceyhan: yes you're right it moved below accidently

melihceyhan:
Done! Thanks!

PR #434, opened 2019-03-06
Revert "Milestone 28: Added multi-user authenticationContext support in JavaScript"

Vicodex:
Reverts PlayFab/SDKGenerator#403


Vicodex:
My bad, I committed prematurely.

PR #436, opened 2019-03-07
Milestone 33.a: Lightweight events for Unity 2018 Delivery 1

cobalt910:
Lightweight events are custom high-volume, high-throughput events sent to Microsoft's OneDS (One Collector) service account associated with PlayFab user's entity ID.

Changes:

- PlayFabEventApi class (the main API to emit custom events, such as Lightweight events)
- PlayFabEvent class (custom event, such as a Lightweight event)
- PlayFabEventRouter class
- IEventPipeline interface
- OneDSEventPipeline class (and any support classes)
- OneDSEventPipelineSettings class
- OneDSEventsApi class (and any support classes)
-Bond serializer and Common Schema 3.0 composer/decorator classes for OneDS calls. These are external dependencies.
- OneDSSysHttpPlugin class

Pull-request in [UnityBetaSDK](https://github.com/PlayFab/UnityBeta/pull/15)


targets/unity-v2/source/Shared/Internal/OneDsUtility.cs
```diff
@@ -0,0 +1,111 @@
+using System;
+using System.Net.Http;
```
Vicodex: Please remove this using System.Net.Http.
I know this may work in more recent Unity 2018, but in Unity 2017 (with 4.6 runtime) it doesn't support out-of-the box. Please see another comment below.

targets/unity-v2/source/Shared/Internal/OneDsUtility.cs
```diff
@@ -0,0 +1,111 @@
...
+                    httpResponseString = getText.Invoke();
+                }
+                catch (HttpRequestException exception)
+                {
+                    var error = new PlayFabError();
+                    error.Error = PlayFabErrorCode.ConnectionError;
+                    error.ErrorMessage = exception.InnerException.Message;
+                    callback?.Invoke(error);
+                    return;
+                }
```
Vicodex: Please remove this, the catch-all Exception block below should be sufficient. HttpRequestException is not supported in Unity 2017 (with 4.6 runtime) out-of-the-box, so we can remove this part without sacrificing much.

targets/unity-v2/source/Shared/Public/Pipeline/OneDSEventPipeline.cs
```diff
@@ -0,0 +1,352 @@
...
+        /// <summary>
+        /// Creates an instance of EventPipeline.
+        /// </summary>
+        /// <param name="settings">The configuration settings for event pipeline.</param>
+        public OneDSEventPipeline(
+            OneDSEventPipelineSettings settings,
+            ILogger logger)
+        {
+            this.settings = settings ?? throw new ArgumentNullException(nameof(settings));
+            this.logger = logger ?? throw new ArgumentNullException(nameof(logger));
```
Vicodex: Please replace these two lines with old-fashioned code:

            if (settings == null)
            {
                throw new ArgumentNullException(nameof(settings));
            }
            else
            {
                this.settings = settings;
            }

            if (logger == null)
            {
                throw new ArgumentNullException(nameof(logger));
            }
            else
            {
                this.logger = logger;
            }

to make it compatible with older C# 6.0 (which is used in Unity 2017 (with 4.6 runtime)).

Vicodex:
Looks good! I generated UnitySDK with your PR and verified the tests. Just a heads-up: you may need to merge it with the latest SDK Gen master before running tests, because it has some bug fixes, without which you'll see a couple tests failing (not yours)

Vicodex:
Can we please add one more important thing: since we generate the same Unity SDK code for newer and older versions of Unity, we need to make sure that our SDK remains compilable when loaded in the older Unity. Some of the classes that we added use newer runtime (4.6+) so we have to do the following:
1) Wrap PlayFabEventAPI class with NET_4_6 macro to make sure it won't be used with 3.5 runtime
2) Wrap the newly added tests with NET_4_6 macro to make sure they won't be used with 3.5 runtime
3) Switch runtime in Unity's Build Settings --> Player Settings to 3.5
4) Get rid of all compilation errors caused by incompatibility runtime version by wrapping necessary classes with NET_4_6 macro. The project must be compilable and tests must be runnable with 3.5 (albeit without the Lightweight events, of course). 
5) When we bring Lightweight events to runtime 3.5 (as part of the next milestone) we will adjust these macroses accordingly

Vicodex:
Looks good and verified. If someone wants to test, please make sure to merge with current master which has recent fixes.

targets/unity-v2/source/Shared/Public/OneDSEventsAPI.cs
```diff
@@ -0,0 +1,207 @@
...
+                    },
+                    CustomData = customData
+                };
+
+                PlayFabSettings.GlobalErrorHandler?.Invoke(apiMethodResult.Error);
+                return apiMethodResult;
+            }
+
+            // get transport plugin for OneDS
+            var transport = PluginManager.GetPlugin<IOneDSTransportPlugin>(PluginContract.PlayFab_Transport, PluginManager.PLUGIN_TRANSPORT_ONEDS);
```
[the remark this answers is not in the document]
pgilmorepf: Why doesn't PluginContract just have another enum value for OneDsTransport. We already agreed it's a totally separate thing. I think this is abusing the original intent of the enum and the string, and using them incorrectly.

targets/unity-v2/source/Shared/Public/OneDSEventsAPI.cs
```diff
@@ -0,0 +1,207 @@
...
+                }
+            }
+
+            PlayFabResult<WriteEventsResponse> result = null;
+            transport.DoPost(serializedBatch, headers, httpResult =>
+            {
+                if (httpResult is PlayFabError)
+                {
+                    var error = (PlayFabError)httpResult;
+                    PlayFabSettings.GlobalErrorHandler?.Invoke(error);
```
pgilmorepf: Doesn't transport.DoPost already invoke GlobalErrorHandler?
If not, it should, and it shouldn't be here.

targets/unity-v2/source/Shared/Public/Pipeline/EventBatchingStage.cs
```diff
@@ -0,0 +1,172 @@
...
+            if (!this.batches.ContainsKey(titleId))
+            {
+                this.CreateNewBatch(titleId);
+            }
+        }
+
+        private static bool ValidationCheck(IPlayFabEmitEventRequest request)
+        {
+            // Payload size validation
+            // TODO
```
pgilmorepf: This seems important.

targets/unity-v2/source/Shared/Public/Pipeline/EventSendingStage.cs
```diff
@@ -0,0 +1,112 @@
...
+                    if (response != null)
+                    {
+                        oneDSEventsApi.SetCredentials("o:" + response.TenantId, response.IngestionKey, response.TelemetryJwtToken, response.TelemetryJwtHeaderKey, response.TelemetryJwtHeaderPrefix);
+                    }
+                    else
+                    {
+                        throw new Exception($"Failed to get OneDS authentication token from PlayFab service");
+                    }
+                }
+                catch (Exception e)
```
pgilmorepf: I'm not a fan of putting large blocks of code in try-catch, especially if it seems that the whole point is to just catch the exception thrown 3 lines up (but maybe that's not the only case?)
It's a very dangerous way to code, and it invites problems.
Like, maintaining this code, I have no idea where all the possible exceptions can come from, or why they might happen.

targets/unity-v2/source/Shared/Public/Pipeline/EventSendingStage.cs
```diff
@@ -0,0 +1,112 @@
...
+                }
+            }
+
+            Task<PlayFabResult<WriteEventsResponse>> apiTask = oneDSEventsApi.WriteTelemetryEventsAsync(writeEventsRequest);
+            try
+            {
+                apiTask.Wait(cts.Token);
+                FulfillPromises(batch.Events, apiTask.Result);
+            }
+            catch (Exception e)
```
pgilmorepf: This is better, because it's just two lines, so I have a pretty good idea where the exception will occur.

targets/unity-v2/source/Shared/Public/Pipeline/EventSendingStage.cs
pgilmorepf: If this is LITERALLY just for the exception 3 lines up, then that's TERRIBLY lazy, and should be fixed.

targets/unity-v2/source/Shared/Public/Pipeline/OneDSEventPipeline.cs
```diff
@@ -0,0 +1,357 @@
...
+        // references to individual stages of the pipeline task
+        private Task batchingTask; 
+        private Task sendingTask;
+
+        // The composition of all pipeline tasks including stages
+        private Task pipelineTask;
+
+        private ILogger logger;
+
+        public OneDSEventPipelineSettings Settings
```
pgilmorepf: This could just be:
public OneDSEventPipelineSettings Settings { get; private set; }
It reduces complexity and more strongly indicates intent.
It could also be:
public readonly OneDSEventPipelineSettings Settings = new OneDSEventPipelineSettings();
Which would be even better, if that's a viable workflow.

targets/unity-v2/source/Shared/Public/Pipeline/OneDSEventPipeline.cs
```diff
@@ -0,0 +1,357 @@
...
+        {
+            get
+            {
+                return this.settings;
+            }
+        }
+
+        public Task PipelineTask
```
pgilmorepf: Ditto above.

targets/unity-v2/source/Shared/Public/Pipeline/OneDSEventPipeline.cs
```diff
@@ -0,0 +1,357 @@
...
+
+        private void Cancel()
+        {
+            this.eventBuffer?.CompleteAdding();
+
+            // This should also direct all stages to mark their output buffers as complete for adding:
+            this.pipelineCancellationTokenSource?.Cancel();
+        }
+
+        private void ThrowIfDisposed()
```
pgilmorepf: I don't like using an exception when an if will do, and I VERY much do not like explicitly turning an if info an exception.
It serves no purpose other than making code harder to maintain, and introducing a bunch of unintended consequences.
Every place where this function is used, convert it to be an if-return.

targets/unity-v2/source/Shared/Public/Pipeline/OneDSEventPipeline.cs
```diff
@@ -0,0 +1,357 @@
...
+                disposed = true;
+            }
+        }
+
+        public void Dispose()
+        {
+            try
+            {
+                this.Dispose(true);
+                GC.SuppressFinalize(this);
```
pgilmorepf: This seems dangerous... you never turn it back on?

targets/unity-v2/source/Shared/Public/Pipeline/OneDSEventPipeline.cs
```diff
@@ -0,0 +1,357 @@
...
+        }
+
+        #region IDisposable Support
+        private bool disposed = false;
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!disposed)
+            {
+                if (disposing)
```
pgilmorepf: disposing is always true.
What is this for?

targets/unity-v2/source/Shared/Public/PlayFabSettings.cs.ejs
```diff
@@ -76,6 +76,7 @@ namespace PlayFab
             }
         }
 
+        public static ErrorCallback GlobalErrorHandler;
```
pgilmorepf: I'm super confused by this, because GlobalErrorHandler should already exist...

pgilmorepf: If it doesn't exist, I'd VERY MUCH like to not add it.

targets/unity-v2/source/Shared/Public/PluginManager.cs
```diff
@@ -103,5 +112,26 @@ private ITransportPlugin CreatePlayFabTransportPlugin()
...
+        private IOneDSTransportPlugin CreateOneDSTransportPlugin()
+        {
+            IOneDSTransportPlugin transport = null;
+#if !UNITY_WSA && !UNITY_WP8
+            if (PlayFabSettings.RequestType == WebRequestType.HttpWebRequest)
+                transport = new OneDsWebRequestPlugin();
+#endif
+#if UNITY_2017_2_OR_NEWER
+            if (PlayFabSettings.RequestType == WebRequestType.UnityWww)
+                transport = new OneDsWwwPlugin();
```
pgilmorepf: Todd updated the standard IHTTP equivalent of this function, and simplified and improved the logic with the idea that we're shifting towards UnityWebRequest being the more stable default. (WWW doesn't even exist anymore in 2018.3)

Please check that out, and update this sequence to match.

targets/unity-v2/Testing/Tests/Client/ClientApiTests.cs
```diff
@@ -465,6 +472,158 @@ private void WriteEventCallback(WriteEventResponse result)
...
+        private PlayFabEvent CreateSamplePlayFabEvent(string name, PlayFabEventType eventType)
+        {
+            var customEvent = new PlayFabEvent { Name = name, EventType = eventType };
+            customEvent.SetProperty("Prop-A", true);
+            customEvent.SetProperty("Prop-B", "custom");
+            customEvent.SetProperty("Prop-C", 567);
+            return customEvent;
+        }
+
+        private static void ContinueWithContext<T>(Task<PlayFabResult<T>> srcTask, UUnitTestContext testContext,
```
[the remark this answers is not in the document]
Vicodex: Update: if your goal was to make this code ALSO work in runtime 3.5 you may keep it this way for the sake of having one codebase for both runtimes. If that was not the reason, please consider addressing this feedback.

---
In reply to: [264019106](https://github.com/PlayFab/SDKGenerator/pull/436#discussion_r264019106) [](ancestors = 264019106)

targets/unity-v2/source/Shared/Public/Pipeline/EventBatchingStage.cs
Vicodex: It was reserved for the future. Please keep it as is for now.

---
In reply to: [264019975](https://github.com/PlayFab/SDKGenerator/pull/436#discussion_r264019975) [](ancestors = 264019975)

targets/unity-v2/source/Shared/Public/Pipeline/EventSendingStage.cs
Vicodex: No, it's for any exception including CancellationException that may happen at authResult.Wait(...) above. This try-catch is justified. Marking it as resolved.

---
In reply to: [264020062](https://github.com/PlayFab/SDKGenerator/pull/436#discussion_r264020062) [](ancestors = 264020062)

targets/unity-v2/source/Shared/Public/Pipeline/OneDSEventPipeline.cs
Vicodex: If it's needed to support runtime 3.5 and older version of C# please feel free to keep it. I'd rather have one codebase.

---
In reply to: [264020099](https://github.com/PlayFab/SDKGenerator/pull/436#discussion_r264020099) [](ancestors = 264020099)

PR #438, opened 2019-03-08
ifdefing out www

toddbellMSFT:
Removing the possibility of using www after 2018.3

### DIFF
--- a/targets/unity-v2/source/Shared/Internal/PlayFabHttp/PlayFabWWW.cs
+++ b/targets/unity-v2/source/Shared/Internal/PlayFabHttp/PlayFabWWW.cs
@@ -1,4 +1,4 @@
-#if !UNITY_2018_3_OR_NEWER // Unity has deprecated Www
+#if !UNITY_2018_2_OR_NEWER // Unity has deprecated Www
 using System;
 using System.Collections;
 using System.IO;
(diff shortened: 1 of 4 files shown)


targets/unity-v2/source/Shared/Internal/PlayFabHttp/PlayFabWWW.cs
```diff
@@ -1,3 +1,4 @@
+#if !UNITY_2018_3_OR_NEWER // Unity has deprecated Www
```
pgilmorepf: UNITY_2018_2_OR_NEWER

targets/unity-v2/source/Shared/Public/PluginManager.cs
```diff
@@ -92,7 +92,11 @@ private ITransportPlugin CreatePlayFabTransportPlugin()
 #endif
 #if UNITY_2017_2_OR_NEWER
             if (PlayFabSettings.RequestType == WebRequestType.UnityWww)
+#if !UNITY_2018_3_OR_NEWER // PlayFabWww will throw warnings as Unity has deprecated Www
```
pgilmorepf: UNITY_2018_2_OR_NEWER
And.... you're really screwing with the logic by removing the body of the if statement, without removing the if statement.
The #if and the #endif need to be inside the brackets, specifically and only around the line:
```transport = new PlayFabWww();```
Because otherwise you're changing the whole logic flow of this method, likely breaking everything.

Vicodex:
Please hold on this, don't merge yet

Vicodex:
Thanks. Unblocked.

PR #441, opened 2019-03-09
Fix broken PlayFabApiSettings

Vicodex:
It was giving me so much grief today

### DIFF
--- a/targets/unity-v2/source/Shared/Public/PlayFabApiSettings.cs.ejs
+++ b/targets/unity-v2/source/Shared/Public/PlayFabApiSettings.cs.ejs
@@ -10,7 +10,7 @@ namespace PlayFab
         };
 
         /// <summary> This is only for customers running a private cluster.  Generally you shouldn't touch this </summary>
-        public string ProductionEnvironmentUrl = ".playfabapi.com";
+        public string ProductionEnvironmentUrl = "playfabapi.com";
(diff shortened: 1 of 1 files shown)


targets/unity-v2/source/Shared/Public/PlayFabApiSettings.cs.ejs
```diff
@@ -10,7 +10,7 @@ namespace PlayFab
         };
 
         /// <summary> This is only for customers running a private cluster.  Generally you shouldn't touch this </summary>
-        public string ProductionEnvironmentUrl = ".playfabapi.com";
+        public string ProductionEnvironmentUrl = "playfabapi.com";
```
pgilmorepf: public string ProductionEnvironmentUrl = PlayFabSettings.ProductionEnvironmentUrl;
I don't want that same string hard coded in many places.

Vicodex: Ok let me try that. Thanks!!

Vicodex: That property is private and named differently with a "private" word in it. I don't want to do rename refactoring across the board now, and I need to urgently unblock failing tests now. Going to commit as it is, and will file a bug to address minor thing later.

PR #443, opened 2019-03-09
CSharp refactor & Test resolution

pgilmorepf:
Updating core Settings, PlayFabAuthenticationContext, and PlayFabApiSettings, to reduce duplication, and enable a better data flow.
Did a regex to remove all trailing spaces
Using async more/better in the test runner, which fixes some out-of-order outputs in the test output.
Removed some deprecated defines
Fixed other defines
Moved a template to a more appropriate place
The fallback mechanism for API instances and statics now works a lot more smoothly and consistently.
Fixed failing tests, and failing test upload process, and upload process reporting
Far better define wrappers around sensitive lines of API files.


targets/csharp/templates/PlayFabAuthenticationContext.cs.ejs
```diff
@@ -2,24 +2,23 @@ namespace PlayFab
...
+            return !string.IsNullOrEmpty(EntityToken);
+        }
 #endif
 #if !DISABLE_PLAYFABCLIENT_API
         /// <summary> Session ticket for Client API. Auto-Populated by any login or registration call. </summary>
         public string ClientSessionTicket = null;
-#endif
-#if ENABLE_PLAYFABSERVER_API || ENABLE_PLAYFABADMIN_API 
-        /// <summary> You must set this value for PlayFabSdk to work properly (Found in the Game Manager for your title, at the PlayFab Website) </summary>
-        public string DeveloperSecretKey = null;
```
Vicodex: >DeveloperSecretKey [](start = 22, length = 18)

Moving DeveloperSecretKey from authentication context to API settings will have some consequences:
1) Users can't do a scenario anymore where they could set it for one particular call (via an authentication context in request), without having to change either static settings or API settings of an API instance
2) We are splitting pieces of authentication data (credentials) between two different places
3) Inconsistency in design and feature documentation between C# SDK and C++, Unreal BP/C++, Unity, JavaScript SDKs. 

This is my biggest concern about this PR. I like the other changes.

PR #448, opened 2019-03-12
Fix Plugin build

Vallantyn:
* Moved Cpp/Blueprint Tests into the related modules
* Fixed includes
* Added empty file to generate TestTitleData folder
* Fixed some modules dependency

@pgilmorepf I realized the errors Jenkins had were because it was using the build script form the plugin.
After fixing the include errors (somehow the example project doesn't need full include paths, but the plugin alone does), the errors you were having originally with the BlueprintTests (not being able to compile outside of the Blueprint Module) appeared (but only when building the plugin alone).
So... I did what I should've done from the beginning: moved the BlueprintTests in the Blueprint module.
I also moved the CppTests into the cpp module, and to be really logic with all that process, the TestActor and the commandline tool were moved in the test project.
So now the TestFramework is **really** a Test Framework. No test implementation inside, just the base classes. 
The Test Project implements the Test Runners.
The Tests are implemented in the proper modules.
After that a bunch of other errors forced me to further cleanup the includes.

I tested this code from 4.18 to 4.21, both the plugin build script and the example project build script (+ Tests).


targets/UnrealMarketplacePlugin/examplesource/ExampleProject/Source/ExampleProject/RunTestsCommandlet.cpp.ejs
```diff
@@ -36,7 +34,7 @@ int32 URunTestsCommandlet::Main(const FString& Params)
     }
     else
     {
-        UE_LOG(LogPlayFabTests, Error, TEXT("Invalid Engine instance."));
+        UE_LOG(LogTemp, Error, TEXT("Invalid Engine instance."));
```
pgilmorepf: Please create a PlayFabTestLog.
Unreal frowns on people using LogTemp.

pgilmorepf: Or use a PlayFab one that's already available in whichever module you're in.

targets/UnrealMarketplacePlugin/source/PlayFab/Source/PlayFabCpp/Private/Tests/PlayFabCppTests.cpp.ejs
```diff
@@ -1,9 +1,8 @@
 <%- copyright %>
 
-#include "CppTests/PlayFabCppTests.h"
+#include "Tests/PlayFabCppTests.h"
```
pgilmorepf: In the case of the CPP interface tests, it's important that the tests exist OUTSIDE of the cpp interface module, because that's an important part of testing them. The CPP module MUST work when accessed from the outside, and the tests have to demonstrate that.

The CPP tests have been confirmed as working already in that state, so let's not move these tests into the CPP module.

targets/UnrealMarketplacePlugin/source/PlayFab/Source/PlayFabCpp/PlayFabCpp.Build.cs
```diff
@@ -27,5 +27,12 @@ public PlayFabCpp(ReadOnlyTargetRules Target) : base(Target)
             "JsonUtilities",
             "PlayFabCommon"
         });
+
+        PrivateDependencyModuleNames.AddRange(
```
pgilmorepf: The CPP module should not depend on the Tests Module.

targets/UnrealMarketplacePlugin/source/PlayFab/Source/PlayFabTests/PlayFabTests.Build.cs
```diff
@@ -24,12 +24,18 @@ public PlayFabTests(ReadOnlyTargetRules Target) : base(Target)
...
             );
+
+//            PrivateDependencyModuleNames.AddRange(
+//                new string[]
+//                {
+//                    "PlayFab",
+//                    "PlayFabCpp",
+//                    "PlayFabCommon"
+//                }
+//            );
```
pgilmorepf: Please don't add commented code.

PR #450, opened 2019-03-12
Fix for Unity in Jenkins, part 2

Vicodex:
- Prior code had a compilation bug for versions of Unity earlier than 2017.2 (PlayFabUnityHttp is available only after 2017.2, so earlier Unity could not find it)
- Fixed a few yellow warnings in Unity

### DIFF
--- a/targets/unity-v2/source/Shared/Public/PluginManager.cs
+++ b/targets/unity-v2/source/Shared/Public/PluginManager.cs
@@ -100,19 +100,24 @@ namespace PlayFab
                 transport = new PlayFabWebRequest();
 #endif
 
-#if !UNITY_2018_2_OR_NEWER // PlayFabWww will throw warnings as Unity has deprecated Www
-            if (PlayFabSettings.RequestType == WebRequestType.UnityWww)
-            {
-                transport = new PlayFabWww();
-            }
-#endif
-
+#if UNITY_2018_2_OR_NEWER // PlayFabWww will throw warnings as Unity has deprecated Www
             if (transport == null)
                 transport = new PlayFabUnityHttp();
+#elif UNITY_2017_2_OR_NEWER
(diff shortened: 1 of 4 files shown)


targets/unity-v2/source/Shared/Public/PluginManager.cs
```diff
@@ -100,32 +100,45 @@ private ITransportPlugin CreatePlayFabTransportPlugin()
...
-            }
-#endif
-
+            
             if (transport == null)
                 transport = new PlayFabUnityHttp();
+#else
+            if (transport == null)
+                transport = new PlayFabWww();
+#endif
```
pgilmorepf: This looks like a bad merge.
Unless you're rolling it back on purpose.

targets/unity-v2/source/Shared/Public/PluginManager.cs
```diff
@@ -100,32 +100,45 @@ private ITransportPlugin CreatePlayFabTransportPlugin()
...
                 transport = new OneDsWwwPlugin();
-#endif
-            if(transport == null)
+            
+            if (transport == null)
                 transport = new OneDsUnityHttpPlugin();
+#else
+            if (transport == null)
+                transport = new OneDsWwwPlugin();
+#endif
```
pgilmorepf: Likewise.

targets/unity-v2/source/Shared/Public/PluginManager.cs
toddbellMSFT: Victor was mentioning there were some issues with other versions of Unity where the Www class should probably still work

Vicodex: Me and Paul just discuss this offline, and I explained why this is a necessary fix

PR #452, opened 2019-03-13
Milestone #32: Authentication service in Unity.

cobalt910:
- Frictionless login with Android, iOS, XBox, Nintendo etc..
- Ability to link/unlink device or service to existing account.

### DIFF
--- a/targets/unity-v2/Testing/Tests/Client/PlayFabAuthServiceTests.cs
+++ b/targets/unity-v2/Testing/Tests/Client/PlayFabAuthServiceTests.cs
@@ -26,7 +26,7 @@ public class PlayFabAuthServiceTests : UUnitTestCase
         _emailAuthService.OnLoginSuccess += success =>
         {
             testContext.True(!string.IsNullOrEmpty(success.PlayFabId));
-            testContext.NotNull(_emailAuthService.AuthenticationContext);
+            testContext.NotNull(_emailAuthService.IsClientLoggedIn());
             testContext.EndTest(UUnitFinishState.PASSED, "Email & password auth success. " + success.PlayFabId);
         };
         _emailAuthService.OnPlayFabError += error =>
@@ -37,7 +37,7 @@ public class PlayFabAuthServiceTests : UUnitTestCase
         {
             testContext.EndTest(UUnitFinishState.FAILED, "Email & password auth failed.");
         };
-        _emailAuthService.Authenticate(AuthTypes.EmailAndPassword);
+        _emailAuthService.Authenticate(AuthTypes.EmailPassword);
     }
 
     [UUnitTest]
@@ -47,7 +47,7 @@ public class PlayFabAuthServiceTests : UUnitTestCase
         _emailAuthService.OnPlayFabLink += (auth, error) =>
         {
             if (error == null) testContext.EndTest(UUnitFinishState.PASSED, "Link deviceId success.");
-            else testContext.EndTest(UUnitFinishState.FAILED, "Link deviceId failed with error: " + error.GenerateErrorReport());
+            else testContext.Fail("Link deviceId failed with error: " + error.GenerateErrorReport());
         };
         _emailAuthService.Link(AuthTypes.Silent);
     }
@@ -61,16 +61,16 @@ public class PlayFabAuthServiceTests : UUnitTestCase
         }, response =>
         {
 #if UNITY_ANDROID && !UNITY_EDITOR
-            testContext.True(response.AccountInfo.AndroidDeviceInfo.AndroidDeviceId.Equals(PlayFabSettings.DeviceUniqueIdentifier), "Android deviceID not match!");
+            testContext.StringEquals(response.AccountInfo.AndroidDeviceInfo.AndroidDeviceId, PlayFabSettings.DeviceUniqueIdentifier, "Android deviceID not match!");
 #elif UNITY_IPHONE || UNITY_IOS && !UNITY_EDITOR
-            testContext.True(response.AccountInfo.IosDeviceInfo.IosDeviceId.Equals(PlayFabSettings.DeviceUniqueIdentifier), "iOS deviceID not match!");
+            testContext.StringEquals(response.AccountInfo.IosDeviceInfo.IosDeviceId, PlayFabSettings.DeviceUniqueIdentifier, "iOS deviceID not match!");
 #else
-            testContext.True(response.AccountInfo.CustomIdInfo.CustomId.Equals(PlayFabSettings.DeviceUniqueIdentifier), "customId not match!");
+            testContext.StringEquals(response.AccountInfo.CustomIdInfo.CustomId,PlayFabSettings.DeviceUniqueIdentifier, "customId not match!");
 #endif
             testContext.EndTest(UUnitFinishState.PASSED, "DeviceId successfully linked!");
         }, error =>
         {
-            testContext.EndTest(UUnitFinishState.FAILED, "GetAccountInfo error: " + error.ErrorMessage);
+            testContext.Fail("GetAccountInfo error: " + error.ErrorMessage);
         });
     }
 
@@ -80,12 +80,12 @@ public class PlayFabAuthServiceTests : UUnitTestCase
         var silentAuth = new PlayFabAuthService();
         silentAuth.OnLoginSuccess += success =>
         {
-            testContext.True(_emailAuthService.AuthenticationContext.PlayFabId == success.PlayFabId);
+            testContext.StringEquals(_emailAuthService.AuthenticationContext.PlayFabId, success.PlayFabId);
             testContext.EndTest(UUnitFinishState.PASSED, "Silent auth success with playFabId: " + success.PlayFabId);
         };
         silentAuth.OnPlayFabError += error =>
         {
-            testContext.EndTest(UUnitFinishState.FAILED, "Silent auth failed with error: " + error.ErrorMessage);
+            testContext.Fail("Silent auth failed with error: " + error.ErrorMessage);
         };
         silentAuth.Authenticate(AuthTypes.Silent);
     }
@@ -96,7 +96,7 @@ public class PlayFabAuthServiceTests : UUnitTestCase
         _emailAuthService.OnPlayFabUnlink += (auth, error) =>
         {
             if(error == null) testContext.EndTest(UUnitFinishState.PASSED, "UnLink deviceId success.");
-            else testContext.EndTest(UUnitFinishState.FAILED, "UnLink deviceId failed with error: " + error.ErrorMessage);
+            else testContext.Fail("UnLink deviceId failed with error: " + error.ErrorMessage);
         };
         _emailAuthService.Unlink(AuthTypes.Silent);
     }
@@ -110,16 +110,16 @@ public class PlayFabAuthServiceTests : UUnitTestCase
         }, response =>
         {
 #if UNITY_ANDROID && !UNITY_EDITOR
-            testContext.True(response.AccountInfo.AndroidDeviceInfo == null, "Android deviceID should be null!");
+            testContext.IsNull(response.AccountInfo.AndroidDeviceInfo, "Android deviceID should be null!");
 #elif UNITY_IPHONE || UNITY_IOS && !UNITY_EDITOR
-            testContext.True(response.AccountInfo.IosDeviceInfo == null, "iOS deviceID should be null!");
+            testContext.IsNull(response.AccountInfo.IosDeviceInfo, "iOS deviceID should be null!");
 #else
-            testContext.True(response.AccountInfo.CustomIdInfo == null, "customID should be null!");
+            testContext.IsNull(response.AccountInfo.CustomIdInfo, "customID should be null!");
 #endif
             testContext.EndTest(UUnitFinishState.PASSED, "DeviceId successfully unlinked!");
         }, error =>
         {
-            testContext.EndTest(UUnitFinishState.FAILED, "GetAccountInfo error: " + error.ErrorMessage);
+            testContext.Fail("GetAccountInfo error: " + error.ErrorMessage);
         });
     }
 
@@ -134,7 +134,7 @@ public class PlayFabAuthServiceTests : UUnitTestCase
         };
         silentAuth.OnPlayFabError += error =>
         {
-            testContext.EndTest(UUnitFinishState.PASSED, "Silent auth abort with error: " + error.Error.ToString());
+            testContext.Fail("Silent auth abort with error: " + error.Error.ToString());
         };
         silentAuth.Authenticate(AuthTypes.Silent);
     }
--- a/targets/unity-v2/Testing/Tests/Client/PlayFabAuthServiceTests.cs
+++ b/targets/unity-v2/Testing/Tests/Client/PlayFabAuthServiceTests.cs
@@ -1,0 +1,141 @@
+﻿using PlayFab;
+using PlayFab.ClientModels;
+using PlayFab.UUnit;
+
+public class PlayFabAuthServiceTests : UUnitTestCase
+{
+    public override void Tick(UUnitTestContext testContext)
+    {
+        // No async work needed
+    }
+
+    private PlayFabAuthService _emailAuthService;
+
+    [UUnitTest]
+    public void EmailPasswordLoginTest(UUnitTestContext testContext)
+    {
+        const string email = "LoginTest@gmail.com";
+        const string password = "395847";
+        const string username = "LoginTest";
+
+        _emailAuthService = new PlayFabAuthService();
+        _emailAuthService.Email = email;
+        _emailAuthService.Password = password;
+        _emailAuthService.Username = username;
+
+        _emailAuthService.OnLoginSuccess += success =>
+        {
+            testContext.True(!string.IsNullOrEmpty(success.PlayFabId));
+            testContext.NotNull(_emailAuthService.AuthenticationContext);
+            testContext.EndTest(UUnitFinishState.PASSED, "Email & password auth success. " + success.PlayFabId);
+        };
+        _emailAuthService.OnPlayFabError += error =>
+        {
+            testContext.EndTest(UUnitFinishState.FAILED, "Email & password auth failed with error: " + error.GenerateErrorReport());
+        };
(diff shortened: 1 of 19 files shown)


targets/unity-v2/source/Authentication/PlayFabAuthService.cs
```diff
@@ -0,0 +1,228 @@
...
+                OnPlayFabUnlink.Invoke(unlinkType, error);
+        }
+
+        internal void InvokeDisplayAuthentication()
+        {
+            if (OnDisplayAuthentication != null)
+                OnDisplayAuthentication.Invoke();
+        }
+
+        public void Link(AuthTypes linkType)
```
pgilmorepf: This method reinforces my concern with calling Link(FB) and then Link(Google), and AuthToken being a shared value.
In fact, it tips the scales. It's not an @MarcoWilliamsPF question anymore. We must delete the shared AuthToken value, and break out separate inputs for every authentication type.
Otherwise we can't support two simultaneous recoverable logins.

targets/unity-v2/source/Authentication/Strategies/EmailPasswordAuthStrategy.cs
```diff
@@ -0,0 +1,69 @@
...
+                    TitleId = PlayFabSettings.TitleId,
+                    CustomId = authService.RememberMeId,
+                    CreateAccount = true,
+                    InfoRequestParameters = authService.InfoRequestParams
+                }, resultCallback, errorCallback);
+
+                return;
+            }
+
+            // a good catch: If username & password is empty, then do not continue, and Call back to Authentication UI Display
```
pgilmorepf: I agree it's a good catch, but please don't praise yourself in your comments. Haha.
Keep the comment drop the praise.

targets/unity-v2/source/Authentication/Strategies/EmailPasswordAuthStrategy.cs
```diff
@@ -0,0 +1,69 @@
...
+                Email = authService.Email,
+                Password = authService.Password,
+                InfoRequestParameters = authService.InfoRequestParams
+            }, loginResult =>
+            {
+                // Note: At this point, they already have an account with PlayFab using a Username (email) & Password
+                // If RememberMe is checked, then generate a new Guid for Login with CustomId.
+                if (authService.RememberMe)
+                {
+                    authService.RememberMeId = Guid.NewGuid().ToString();
```
pgilmorepf: RememberMeId is public, and given the async-delay, it's no longer guaranteed to be unset.
Also, simply retrieving it when it's not set will generate a new guid, so we don't need to repeat that logic here. (Delete this line)

targets/unity-v2/source/Authentication/PlayFabAuthService.cs
```diff
@@ -0,0 +1,228 @@
...
+        public bool RememberMe
+        {
+            get { return PlayerPrefs.GetInt(_LoginRememberKey, 0) != 0; }
+            set { PlayerPrefs.SetInt(_LoginRememberKey, value ? 1 : 0); }
+        }
+
+        /// <summary>
+        /// Remember the type of authenticate for the user
+        /// </summary>
+        public AuthTypes AuthType
```
pgilmorepf: @MarcoWilliamsPF Looping back around to this:
Every interface function accepts an AuthTypes as a parameter.
I am concerned that this variable, and the preference both:

1. Doesn't serve a purpose
2. Conflicts with the AuthType being manually set in almost every input method.

I would opt to completely remove this property, as it is more flexible, and removes the conflict.

targets/unity-v2/source/Authentication/Strategies/FacebookAuthStrategy.cs
```diff
@@ -0,0 +1,55 @@
+using System;
+using PlayFab.ClientModels;
+
+namespace PlayFab.Authentication.Strategies
+{
+    internal sealed class FacebookAuthStrategy : IAuthenticationStrategy
```
pgilmorepf: Why doesn't Facebook have a CustomId silent auth fallback like email?

targets/unity-v2/source/Authentication/Strategies/FacebookInstantGameAuthStrategy.cs
```diff
@@ -0,0 +1,56 @@
+using System;
+using PlayFab.ClientModels;
+
+namespace PlayFab.Authentication.Strategies
+{
+    internal sealed class FacebookInstantGameAuthStrategy : IAuthenticationStrategy
```
pgilmorepf: Why doesn't FBInstant have a CustomId silent auth fallback like email?

targets/unity-v2/source/Authentication/Strategies/GameCenterAuthStrategy.cs
```diff
@@ -0,0 +1,55 @@
+using System;
+using PlayFab.ClientModels;
+
+namespace PlayFab.Authentication.Strategies
+{
+    internal sealed class GameCenterAuthStrategy : IAuthenticationStrategy
```
pgilmorepf: Why doesn't GameCenter have a CustomId silent auth fallback like email?
Basically, now I'm thinking that the silent auth backup should be an automatic call to another strategy, and it should be part of the core service to pick an appropriate silent auth backup, and just do it. (Added a comment to email to this effect)

targets/unity-v2/source/Authentication/Strategies/EmailPasswordAuthStrategy.cs
```diff
@@ -0,0 +1,69 @@
+using System;
+using PlayFab.ClientModels;
+using UnityEngine;
+
+namespace PlayFab.Authentication.Strategies
+{
+    internal sealed class EmailPasswordAuthStrategy : IAuthenticationStrategy
```
pgilmorepf: Why is this duplicating/hard-coding in a bunch of customId silent auth stuff, instead of just invoking the customId strategy?
(duplication of code, and code maintenance)

targets/unity-v2/source/Authentication/PlayFabAuthService.cs
```diff
@@ -0,0 +1,228 @@
...
+        public bool ForceLink;
+        
+        public string AuthTicket;
+        public string OpenIdConnectionId;
+        public string WindowsHelloChallengeSignature;
+        public string WindowsHelloPublicKeyHint;
+        
+        public string PlayFabId { get; private set; }
+        public string SessionTicket { get; private set; }
+        public static EntityKey EntityToken { get; private set; }
```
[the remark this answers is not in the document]
pgilmorepf: Why is this static when nothing else is?

targets/unity-v2/source/Authentication/PlayFabAuthService.cs
```diff
@@ -0,0 +1,228 @@
...
+        {
+            get { return PlayerPrefs.GetString(_PlayFabRememberMeIdKey, ""); }
+            set
+            {
+                var guid = string.IsNullOrEmpty(value) ? Guid.NewGuid().ToString() : value;
+                PlayerPrefs.SetString(_PlayFabRememberMeIdKey, guid);
+            }
+        }
+
+        private static readonly Dictionary<AuthTypes, IAuthenticationStrategy> AuthStrategies = new Dictionary<AuthTypes, IAuthenticationStrategy>();
```
pgilmorepf: This should also not be static.

targets/unity-v2/source/Authentication/PlayFabAuthService.cs
```diff
@@ -0,0 +1,228 @@
...
+
+        static PlayFabAuthService()
```
pgilmorepf: We should move the contents of this to the instance constructor, rather than the static one.

targets/unity-v2/source/Authentication/Strategies/SilentAuthStrategy.cs
```diff
@@ -0,0 +1,140 @@
...
+                CreateAccount = true,
+                InfoRequestParameters = authService.InfoRequestParams
+            }, resultCallback, errorCallback);
+#elif UNITY_IPHONE || UNITY_IOS && !UNITY_EDITOR
+            PlayFabClientAPI.LoginWithIOSDeviceID(new LoginWithIOSDeviceIDRequest
+            {
+                TitleId = PlayFabSettings.TitleId,
+                DeviceModel = SystemInfo.deviceModel,
+                OS = SystemInfo.operatingSystem,
+                DeviceId = SystemInfo.deviceUniqueIdentifier,
```
pgilmorepf: PlayFabSettings.DeviceUniqueIdentifier

targets/unity-v2/source/Authentication/Strategies/SilentAuthStrategy.cs
```diff
@@ -0,0 +1,140 @@
...
+                CreateAccount = true,
+                InfoRequestParameters = authService.InfoRequestParams
+            }, resultCallback, errorCallback);
+#else
+            PlayFabClientAPI.LoginWithCustomID(new LoginWithCustomIDRequest
+            {
+                TitleId = PlayFabSettings.TitleId,
+                CustomId = SystemInfo.deviceUniqueIdentifier,
```
pgilmorepf: RememberMeId

targets/unity-v2/source/Authentication/Strategies/SilentAuthStrategy.cs
```diff
@@ -0,0 +1,140 @@
...
+                {
+                    authService.InvokeLink(AuthTypes.Silent);
+                }, errorCallback =>
+                {
+                    authService.InvokeLink(AuthTypes.Silent, errorCallback);
+                });
+#else
+                PlayFabClientAPI.LinkCustomID(new LinkCustomIDRequest
+                {
+                    CustomId = PlayFabSettings.DeviceUniqueIdentifier,
```
pgilmorepf: RememberMeId

targets/unity-v2/source/Authentication/Strategies/SilentAuthStrategy.cs
```diff
@@ -0,0 +1,140 @@
...
+            {
+                TitleId = PlayFabSettings.TitleId,
+                CustomId = SystemInfo.deviceUniqueIdentifier,
+                CreateAccount = true,
+                InfoRequestParameters = authService.InfoRequestParams
+            }, resultCallback, errorCallback);
+#endif
+        }
+
+        public virtual void Link(PlayFabAuthService authService)
```
pgilmorepf: You should note that your deviceIds mismatched between Authenticate, and Link.
Be very careful with that.

targets/unity-v2/source/Authentication/Strategies/SilentAuthStrategy.cs
```diff
@@ -0,0 +1,140 @@
...
+                {
+                    authService.InvokeUnlink(AuthTypes.Silent);
+                }, errorCallback =>
+                {
+                    authService.InvokeUnlink(AuthTypes.Silent, errorCallback);
+                });
+#else
+                PlayFabClientAPI.UnlinkCustomID(new UnlinkCustomIDRequest
+                {
+                    CustomId = PlayFabSettings.DeviceUniqueIdentifier,
```
pgilmorepf: RememberMeId

targets/unity-v2/source/Authentication/Strategies/UsernamePasswordAuthStrategy.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        public override void Authenticate(PlayFabAuthService authService, Action<LoginResult> resultCallback, Action<PlayFabError> errorCallback)
+        {
+            if (string.IsNullOrEmpty(authService.Email) || string.IsNullOrEmpty(authService.Password))
+            {
+                authService.InvokeDisplayAuthentication();
+                return;
+            }
+
+            base.Authenticate(authService, silentCallback =>
```
pgilmorepf: rename silentCallback to resultCallback so that it matches the other authStrategies and doesn't confuse me.

targets/unity-v2/source/Authentication/PlayFabAuthService.cs
```diff
@@ -131,67 +121,79 @@ public void ClearRememberMe()
...
+                {
+                    TitleId = PlayFabSettings.TitleId,
+                    CustomId = RememberMeId,
+                    InfoRequestParameters = InfoRequestParams,
+                    CreateAccount = true
+                }, InvokeLoginSuccess, InvokePlayFabError);
                 return;
             }
+            
+            var auth = _authStrategies[AuthType];
```
pgilmorepf: If AuthType is invalid, then this will just throw an exception rather than assigning auth to null, so the null check 2 lines down doesn't really serve the purpose you expect it to.

Go ahead and just remove the whole ```if (auth == null) { log, return }``` block below, because it will never go there.

targets/unity-v2/source/Authentication/PlayFabAuthService.cs
```diff
@@ -201,28 +203,52 @@ internal void InvokeDisplayAuthentication()
...
+            auth.Unlink(this, authKeys);
         }
     }
-}
\ No newline at end of file
+
+    [Serializable]
+    public sealed class AuthKeys
+    {
+        public string AuthTicket;
```
pgilmorepf: AuthTicket is inherently bound to the ```AuthTypes authType``` that almost always comes with this.
I really like this structure, and I think moving the authType adjacent to AuthTicket would make it perfect.

targets/unity-v2/source/Shared/Internal/Util.cs
```diff
@@ -101,6 +101,23 @@ public static string ReadAllFileText(string filename)
...
 
+        public static T TryEnumParse<T>(string value)
+        {
+            try
+            {
+                return (T) Enum.Parse(typeof(T), value);
+            }
+            catch (InvalidCastException castException)
+            {
+                return default(T);
```
pgilmorepf: There are some fairly significant performance problems with default(T).
It also doesn't always cross compile as expected.
Please add an additional parameter, which is to be used if the parse fails.
That way the caller provides the default.

targets/unity-v2/source/Shared/Internal/Util.cs
```diff
@@ -101,6 +101,23 @@ public static string ReadAllFileText(string filename)
...
+            }
+            catch (Exception e)
+            {
+                UnityEngine.Debug.LogError("Enum cast failed with unknown error: " + e.Message);
+                return default(T);
```
pgilmorepf: (see above)

targets/unity-v2/source/Shared/Internal/Util.cs
```diff
@@ -101,6 +101,23 @@ public static string ReadAllFileText(string filename)
             return _sb.ToString();
         }
 
+        public static T TryEnumParse<T>(string value)
+        {
+            try
+            {
+                return (T) Enum.Parse(typeof(T), value);
```
pgilmorepf: This is great!

targets/unity-v2/Testing/Tests/Client/PlayFabAuthServiceTests.cs
```diff
@@ -0,0 +1,141 @@
...
+
+        _emailAuthService.OnLoginSuccess += success =>
+        {
+            testContext.True(!string.IsNullOrEmpty(success.PlayFabId));
+            testContext.NotNull(_emailAuthService.IsClientLoggedIn());
+            testContext.EndTest(UUnitFinishState.PASSED, "Email & password auth success. " + success.PlayFabId);
+        };
+        _emailAuthService.OnPlayFabError += error =>
+        {
+            testContext.EndTest(UUnitFinishState.FAILED, "Email & password auth failed with error: " + error.GenerateErrorReport());
```
pgilmorepf: testContext.Fail(msg)

targets/unity-v2/Testing/Tests/Client/PlayFabAuthServiceTests.cs
```diff
@@ -0,0 +1,141 @@
...
+        };
+        _emailAuthService.OnDisplayAuthentication += () =>
+        {
+            testContext.EndTest(UUnitFinishState.FAILED, "Email & password auth failed.");
```
pgilmorepf: testContext.Fail(msg)

targets/unity-v2/source/Authentication/PlayFabAuthService.cs
```diff
@@ -0,0 +1,254 @@
...
+
+        public string Email;
+        public string Username;
+        public string Password;
+        public GetPlayerCombinedInfoRequestParams InfoRequestParams;
+        public bool ForceLink;
+
+        public PlayFabAuthenticationContext AuthenticationContext { get; private set; }
+
+        private const string _LoginRememberKey = "PlayFabLoginRemember";
```
pgilmorepf: Can you put a comment that these pref-keys are incompatible with split-screen/multi-instance play?
There's no good solution to that, because it's dependent on information we can't have.
So just put a comment here saying there's an issue to be fixed with split-screen/multi-instance play regarding these pref-keys.

targets/unity-v2/source/Shared/Internal/Util.cs
pgilmorepf: OOps, one more thing!

targets/unity-v2/source/Authentication/PlayFabAuthService.cs
```diff
@@ -0,0 +1,268 @@
...
+            {
+                if (OnDisplayAuthentication != null)
+                    OnDisplayAuthentication.Invoke();
+
+                return;
+            }
+
+            if (RememberMe && !string.IsNullOrEmpty(RememberMeId))
+            {
+                PlayFabClientAPI.LoginWithCustomID(new LoginWithCustomIDRequest
```
MarcoWilliamsPF: I am not sure why we are doing this.  I do understand the scenario where a user logs in with a email & password and we have to hash that into a remember me id,  but this insinuates that under all conditions when you use the remember me feature that it is a login with a custom id, which is a little misleading in the "auto-login" as it remembers your previous auth type.   This issue isn't a blocker for release but we should revisit it from a design perspective.

PR #460, opened 2019-03-28
Removing tests from the Plugin folder

toddbellMSFT:
we are running into weird issues for tests on the Plugin. We will remove the tests for the moment.


toddbellMSFT:
wrong approach

PR #464, opened 2019-04-03
Implemented iOS HTTP Plugin

v-hoccha:
Implemented iOS HTTP Plugin.
Added iOS build projects and a test application for the UnitTest.


targets/XPlatCppSdk/source/code/source/playfab/PlayFabIOSHttpPlugin.mm
```diff
@@ -0,0 +1,346 @@
+#include <stdafx.h>
```
pgilmorepf: Shouldn't this whole file be wrapped in:
#if defined(PLAYFAB_PLATFORM_IOS)
?

targets/XPlatCppSdk/source/code/source/playfab/OneDSIOSHttpPlugin.mm
```diff
@@ -0,0 +1,114 @@
+#include <stdafx.h>
```
pgilmorepf: Shouldn't this whole file be wrapped in:
#if defined(PLAYFAB_PLATFORM_IOS)
?

targets/XPlatCppSdk/source/code/include/playfab/PlayFabIOSHttpPlugin.h
```diff
@@ -0,0 +1,75 @@
+#pragma once
```
pgilmorepf: Shouldn't this whole file be wrapped in:
#if defined(PLAYFAB_PLATFORM_IOS)
?

targets/XPlatCppSdk/source/code/include/playfab/OneDSIOSHttpPlugin.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+
+#ifndef DISABLE_ONEDS_API
```
pgilmorepf: Shouldn't this whole file be wrapped in:
#if defined(PLAYFAB_PLATFORM_IOS)
?

targets/XPlatCppSdk/source/code/source/playfab/PlayFabIOSHttpPlugin.mm
v-hoccha: I followed the implementation of other platforms. But, yes it will be better to be wrapped or make subdirectories. It will be good to do once later.

targets/XPlatCppSdk/source/code/source/playfab/OneDSIOSHttpPlugin.mm
v-hoccha: I followed the implementation of other platforms. But, yes it will be better to be wrapped or make subdirectories. It will be good to do once later.

targets/XPlatCppSdk/source/code/include/playfab/PlayFabIOSHttpPlugin.h
v-hoccha: I followed the implementation of other platforms. But, yes it will be better to be wrapped or make subdirectories. It will be good to do once later.

targets/XPlatCppSdk/source/code/include/playfab/OneDSIOSHttpPlugin.h
v-hoccha: I followed the implementation of other platforms. But, yes it will be better to be wrapped or make subdirectories. It will be good to do once later.

targets/XPlatCppSdk/source/code/include/playfab/PlayFabIOSHttpPlugin.h
```diff
@@ -0,0 +1,75 @@
...
+        {
+            RequestTask();
+            virtual ~RequestTask();
+
+            bool Initialize(std::unique_ptr<CallRequestContainerBase>& requestContainer);
+            
+            enum State:int
+            {
+                None = 0,
+                Pending = RequestTask::None,
```
PushpadantK: >Pending = RequestTask::None [](start = 16, length = 27)

not sure why you need none, if you assign none here, both will have same value

targets/XPlatCppSdk/source/code/source/playfab/PlayFabIOSHttpPlugin.mm
```diff
@@ -0,0 +1,346 @@
...
+            if(sessionTask)
+            {
+                [sessionTask cancel];
+            }
+        }
+
+        __strong NSURLSessionUploadTask* sessionTask;
+    };
+
+}
```
PushpadantK: can you not include whole code in same { }? sorry for my lack of understanding of .mm file

targets/XPlatCppSdk/source/code/source/playfab/PlayFabIOSHttpPlugin.mm
```diff
@@ -0,0 +1,346 @@
...
+            this->impl->Cancel();
+        }
+    }
+
+    
+}
+
+namespace PlayFab
+{
+    PlayFabIOSHttpPlugin::PlayFabIOSHttpPlugin() :
```
PushpadantK: >PlayFabIOSHttpPlugin [](start = 26, length = 20)

no error logging or reporting :(

targets/XPlatCppSdk/source/code/include/playfab/PlayFabIOSHttpPlugin.h
v-hoccha: Not separated for the status of the value, but clearly for the programmer to recognize it when setting the value. This reduces mistakes in terms of maintenance.

targets/XPlatCppSdk/source/code/source/playfab/PlayFabIOSHttpPlugin.mm
```diff
@@ -0,0 +1,346 @@
...
+        }
+        catch (...)
+        {
+
+        }
+        if(requestTask != nullptr)
+        { // LOCK httpRequestMutex
+            std::unique_lock<std::mutex> lock(httpRequestMutex);
+            requestTask->state = RequestTask::Pending;
+            pendingRequests.push_back(std::move(requestTask));
```
PushpadantK: >pendingRequests.push_back(std::move(requestTask)); [](start = 12, length = 50)

isn't it better to make post request and then add to the queue only to process result? if current SDK design is same then its ok, otherwise this is something to keep in mind for future.

targets/XPlatCppSdk/source/code/include/playfab/PlayFabIOSHttpPlugin.h
PushpadantK: then why assign pending = none, instead intialize with none and then when you get the request mark it pending?

---
In reply to: [272404052](https://github.com/PlayFab/SDKGenerator/pull/464#discussion_r272404052) [](ancestors = 272404052)

v-hoccha: It is because the task created and added to pending queue immediatly. there is actually no state change on automata perspective.

targets/XPlatCppSdk/source/code/include/playfab/PlayFabIOSHttpPlugin.h
```diff
@@ -0,0 +1,75 @@
...
+            {
+                return *dynamic_cast<OneDSCallRequestContainer*>(requestContainer.get());
+            }
+#endif
+            void Cancel();
+            std::atomic<State> state;
+            std::unique_ptr<CallRequestContainerBase> requestContainer;
+            RequestImpl* impl;
+        };
+        std::unique_ptr<std::thread> workerThread;
```
PushpadantK: >std::unique_ptr<std::thread> workerThread; [](start = 8, length = 42)

From purely pattern point of view, other plug-ins use thread variable and not pointer to thread, just want to know the reason to move away from that pattern?

targets/XPlatCppSdk/source/code/source/playfab/PlayFabIOSHttpPlugin.mm
```diff
@@ -0,0 +1,346 @@
...
+                    if (!pendingRequests.empty())
+                    {
+                        this->requestingTask = this->pendingRequests[0];
+                        this->pendingRequests.pop_front();
+                    }
+                    else
+                    {
+                        threadRunning = false;
+                        workerThread->detach();
+                        workerThread = nullptr;
```
PushpadantK: If there are no pending requests then this will exit the thread, is that correct? and then when a new PostRequest comes we will start a new thread? if yes then is it not expensive to start a new thread for every request?

targets/XPlatCppSdk/source/code/source/playfab/PlayFabIOSHttpPlugin.mm
```diff
@@ -0,0 +1,346 @@
...
+        while(threadRunning)
+        {
+            RequestTask::State state;
+            { // LOCK httpRequestMutex
+                std::unique_lock<std::mutex> lock(httpRequestMutex);
+                if(this->requestingTask == nullptr)
+                {
+                    if (!pendingRequests.empty())
+                    {
+                        this->requestingTask = this->pendingRequests[0];
```
PushpadantK: >this->pendingRequests[0]; [](start = 47, length = 25)

std::move?

PushpadantK: this is shared_ptr, so should be ok?

---
In reply to: [272408735](https://github.com/PlayFab/SDKGenerator/pull/464#discussion_r272408735) [](ancestors = 272408735)

PR #466, opened 2019-04-04
Milestone 34: Lightweight Events Unity 2017

cobalt910:
Port of [milestone #33](https://github.com/PlayFab/SDKGenerator/pull/436) to .net 3.5
Pull-request in UnityBeta [here](https://github.com/PlayFab/UnityBeta/pull/17)

Also have additional changes was not present in [TPL_3_5_Mirror](https://github.com/PlayFab/TPL_3_5_Mirror):
* In CustomInterlocked.cs null compare replaced to null pointer compare.
More info in [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.compareexchange?view=netframework-3.5). Exception will thrown if `The address of "location1" is a null pointer.` 
As i understand it's little different then null compare.
* Changed `#if NINTENDO` to `#if UNITY_SWITCH`


targets/unity-v2/Testing/Tests/Client/LightweightEventsTests.cs
```diff
@@ -47,7 +54,11 @@ public void WriteOneDSEvents(UUnitTestContext testContext)
     /// ONEDS API
     /// Test that a batch of custom events can be sent to OneDS server
     /// </summary>
+    #if TPL_35
```
toddbellMSFT: nitpick: can this be left hand aligned to the file?

targets/unity-v2/source/Shared/Internal/OneDsUtility.cs
```diff
@@ -21,49 +22,74 @@ public static void ParseResponse(long httpCode, Func<string> getText, string err
...
+                    {
+                        Error = PlayFabErrorCode.ConnectionError,
+                        ErrorMessage = exception.Message
+                    });
                     return;
                 }
 
                 if (httpCode >= 200 && httpCode < 300)
                 {
+                    Action<Exception> printErrorMessage = e =>
```
toddbellMSFT: would be nice to see a more detailed named parameter instead of just "e".

toddbellMSFT: Why would we want to change the errorMsg here? UnityEngine.Debug.LogException already accepts an Exception which we could plug right in. I'm assuming that may look gross, hence explicilty calling out the source of the error and the message.

targets/unity-v2/source/Shared/External/aria/Schema3_types.cs
```diff
@@ -684,19 +697,24 @@ internal void Reset()
             extApp = null;
             extUtc = null;
             extNet = null;
-            extDevice = null;
```
debhaldarMS: > extDevice = null; [](start = 11, length = 18)

I understand the property is duplicated but does this not change the event schema ? Have we validated that we can send this data and see it in ARIA protal ?

targets/unity-v2/Testing/Tests/Client/LightweightEventsTests.cs
```diff
@@ -1,4 +1,6 @@
-#if NET_4_6
+#if !NET_4_6 && (NET_2_0_SUBSET || NET_2_0)
```
debhaldarMS: Did we do regression test on Unity 2018 and make sure it's not broken ?

targets/unity-v2/source/Shared/Public/Pipeline/OneDSEventPipeline.cs
```diff
@@ -76,11 +79,19 @@ public Task PipelineTask
         public OneDSEventPipeline(OneDSEventPipelineSettings settings, ILogger logger)
         {
             if (settings == null)
```
toddbellMSFT: Can we ensure these are surrounded with brackets { } ?

Would make adding anything in the future easier if necessary.

targets/unity-v2/source/Shared/Public/Pipeline/OneDSEventPipeline.cs
```diff
@@ -76,11 +79,19 @@ public Task PipelineTask
...
+#if TPL_35
+                throw new ArgumentNullException("settings");
+#else
                 throw new ArgumentNullException(nameof(settings));
+#endif
             this.settings = settings;
 
             if (logger == null)
```
toddbellMSFT: Similar situation here with ensuring we visibly define the scope with brackets.

targets/unity-v2/source/Shared/Public/OneDSEventsAPI.cs
```diff
@@ -183,24 +207,60 @@ internal static async Task<PlayFabResult<TelemetryIngestionConfigResponse>> GetT
...
+                        Task.Delay(frequency).Await();
+                    }
+                });
+
+                if(waitTask != Task.WhenAny(waitTask, Task.Delay(timeout)).Await())
+                    throw new TimeoutException();
+            });
+        }
+#else
+        public static async Task WaitWhile(Func<bool> condition, int frequency = WaitWhileFrequencyDefault, int timeout = WaitWhileTimeoutDefault)
```
toddbellMSFT: Should this go in a shared Utils file? It seems fairly generic. It looks like it could be useful in other parts of C# legacy code.

targets/unity-v2/Testing/Tests/Client/LightweightEventsTests.cs
cobalt910: Everything is working.

ryanpcsm:
@cobalt910 Please do regression testing to verify that the current implementation still works in Unity 2018.

targets/unity-v2/source/Shared/External/tpl35/System.Collections.Generic/DefaultEqualityComparer.cs
```diff
@@ -54,7 +54,9 @@ public int GetHashCode(object obj)
             return GetHashCode ((T)obj);
         }
 
+#pragma warning disable 108,114
```
pgilmorepf: Can you please leave comments about which warnings are being disabled here?
It saves me the time of doing an internet search every time.

PR #470, opened 2019-04-06
Implemented Android Plug-in. Added Android test app

alex-xbox:
Implemented Android Plug-in. Added Android test app.


targets/XPlatCppSdk/source/code/include/playfab/OneDSAndroidHttpPlugin.h
```diff
@@ -0,0 +1,32 @@
...
+#ifndef DISABLE_ONEDS_API
+
+#include <playfab/PlayFabCallRequestContainer.h>
+#include <playfab/PlayFabPluginManager.h>
+#include <playfab/PlayFabAndroidHttpPlugin.h>
+
+namespace PlayFab
+{
+    /// <summary>
+    /// OneDSAndroidHttpPlugin is an https implementation to interact with OneDS (One Data Collector) services using WinHTTP.
```
debhaldarMS: >WinHTTP. [](start = 117, length = 8)

this is not accurate here right ?

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
```diff
@@ -0,0 +1,78 @@
...
+
+        struct RequestImpl;
+        struct RequestTask
+        {
+            RequestTask();
+            virtual ~RequestTask();
+
+            bool Initialize(std::unique_ptr<CallRequestContainerBase>& requestContainer);
+            
+            enum State:int
```
debhaldarMS: >enum [](start = 12, length = 4)

can we use an enumclass ?

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
```diff
@@ -0,0 +1,78 @@
...
+namespace PlayFab
+{
+    /// <summary>
+    /// PlayFabAndroidHttpPlugin is an https implementation to interact with PlayFab services using WinHTTP.
+    /// </summary>
+    class PlayFabAndroidHttpPlugin : public IPlayFabHttpPlugin
+    {
+    public:
+        PlayFabAndroidHttpPlugin();
+        PlayFabAndroidHttpPlugin(const PlayFabAndroidHttpPlugin& other) = delete;
```
debhaldarMS: >PlayFabAndroidHttpPlugin(const PlayFabAndroidHttpPlugin& other) [](start = 8, length = 63)

Do you need the copy assignment operator ?

targets/XPlatCppSdk/source/code/include/playfab/QoS/XPlatSocket.h
```diff
@@ -66,6 +66,9 @@ namespace PlayFab
 #if defined(PLAYFAB_PLATFORM_WINDOWS) || defined(PLAYFAB_PLATFORM_XBOX)
             int s, slen;
             WSADATA wsa;
+#elif defined(PLAYFAB_PLATFORM_ANDROID)
+            unsigned int s;
+            socklen_t slen;
```
debhaldarMS: Did we check that QoS actually works on Android ? Or is this just to get past compile ?

targets/XPlatCppSdk/source/code/source/playfab/OneDSAndroidHttpPlugin.cpp
```diff
@@ -0,0 +1,109 @@
+#include <stdafx.h>
+
+#ifndef DISABLE_ONEDS_API
+
+#include <playfab/OneDSAndroidHttpPlugin.h>
+
+namespace PlayFab
+{
+    constexpr auto ONEDS_SERVICE_URL = "https://self.events.data.microsoft.com/OneCollector/1.0/";
```
debhaldarMS: please use fully typed constants

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -0,0 +1,743 @@
...
+
+            JNIEnv* jniEnv;
+            this->jvm->GetEnv(reinterpret_cast<void**>(&jniEnv), JNI_VERSION_1_6);
+            assert(jniEnv);
+            if (jniEnv == nullptr)
+            {
+                return false;
+            }
+
+            jclass localClass = jniEnv->FindClass("com/microsoft/xplatcppsdk/HttpRequest");
```
debhaldarMS: >com/microsoft/xplatcppsdk/HttpRequest" [](start = 51, length = 38)

please make it a constexpr

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -0,0 +1,743 @@
...
+            assert(localClass);
```
debhaldarMS: >assert(localClass); [](start = 12, length = 19)

are these debug asserts or retail asserts ?

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -0,0 +1,743 @@
...
+    {
+        __android_log_print(ANDROID_LOG_INFO, "XPlatCppSdk", "Failed to create PlayFabAndroidHelper");
+        throw e;
+    }
+    if (!(s_helper->Initialize(vm)))
+    {
+        throw "Initializing to load classes failed";
+    }
+
+    return JNI_VERSION_1_6;
```
debhaldarMS: >JNI_VERSION_1_6 [](start = 11, length = 15)

Does this need to be changed for different version of NDK / Android OS versions ?

targets/XPlatCppSdk/source/code/source/playfab/PlayFabEvent.cpp
```diff
@@ -48,7 +48,11 @@ namespace PlayFab
 
     void PlayFabEvent::SetProperty(const std::string& name, const int64_t value)
     {
+#if defined(PLAYFAB_PLATFORM_ANDROID)
+        this->eventContents.Payload[name] = (Int64)value;
+#else // PLAYFAB_PLATFORM_ANDROID
```
debhaldarMS: >// PLAYFAB_PLATFORM_ANDROID [](start = 6, length = 27)

remove since it seems like a little misleading

targets/XPlatCppSdk/source/code/source/playfab/PlayFabEvent.cpp
```diff
@@ -68,7 +72,11 @@ namespace PlayFab
 
     void PlayFabEvent::SetProperty(const std::string& name, const uint64_t value)
     {
+#if defined(PLAYFAB_PLATFORM_ANDROID)
+        this->eventContents.Payload[name] = (Uint64)value;
+#else // PLAYFAB_PLATFORM_ANDROID
```
debhaldarMS: >// PLAYFAB_PLATFORM_ANDROID [](start = 6, length = 27)

remove

targets/XPlatCppSdk/source/build/Android/XPlatCppSdk/src/main/java/com/microsoft/xplatcppsdk/HttpRequest.java
```diff
@@ -0,0 +1,141 @@
...
+            return 400; // Bad request
+        }
+    }
+
+    public byte[] getResponseHttpBody()
+    {
+        if (this.response != null && this.requestCompleted) {
+            try {
+                return this.response.body().bytes();
+            } catch (IOException e) {
```
debhaldarMS: >(IOException e) { [](start = 20, length = 17)

Do you also need to catch a generic expection? We should not let a http stack throw an unhandled exception right since we're communicating read failure via null return values ?

targets/XPlatCppSdk/source/build/Android/UnitTest/TestAppAndroid.cpp
```diff
@@ -0,0 +1,83 @@
...
+        return true;
+    }
+
+    void TestApp::LogPut(const char* message)
+    {
+        // Write to Android Log.
+        __android_log_print(ANDROID_LOG_INFO, "XPlatCppSdk", "%s", message);
+
+        // Write to UI log list.
+        //OutputDebugString(message);
```
debhaldarMS: please remove commented out code

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -0,0 +1,743 @@
...
+#include <android/log.h>
+
+#include <playfab/PlayFabAndroidHttpPlugin.h>
+#include <playfab/PlayFabSettings.h>
+#include <stdexcept>
+#include <vector>
+
+namespace PlayFab
+{
+    struct PlayFabAndroidHelper
```
debhaldarMS: What's the reason for using a struct here instead of a class ? Was this a pattern you were following from other plugin examples ? Usually struct is a container for POD types.

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -0,0 +1,743 @@
...
+
+        httpRequestMutex.lock();
+
+        if(workerThread)
+        {
+            httpRequestMutex.unlock();
+            workerThread->join();
+            httpRequestMutex.lock();
+        }
+        httpRequestMutex.unlock();
```
debhaldarMS: can we just lock once ?

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -0,0 +1,743 @@
...
+            { // LOCK httpRequestMutex
+                std::unique_lock<std::mutex> lock(httpRequestMutex);
+                if (this->requestingTask == nullptr) {
+                    if (!pendingRequests.empty()) {
+                        this->requestingTask = this->pendingRequests[0];
+                        this->pendingRequests.pop_front();
+                        this->requestingTask->impl->SetJniEnv(static_cast<JNIEnv*>(jniEnv));
+                    } else {
+                        threadRunning = false;
+                        workerThread->detach();
```
debhaldarMS: > workerThread->detach(); [](start = 23, length = 24)

Doesn't this mean you can fork any number of threads and exhaust the threadpool ?

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -0,0 +1,743 @@
...
+            jvm->DetachCurrentThread();
+        }
+    }
+
+    void PlayFabAndroidHttpPlugin::WorkerThread(void* jniEnv)
+    {
+        { // LOCK httpRequestMutex
+            std::unique_lock<std::mutex> lock(httpRequestMutex);
+            this->requestingTask = nullptr;
+        }
```
debhaldarMS: Insert an empty line between code blocks

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -0,0 +1,743 @@
...
+        while (threadRunning) {
+            RequestTask::State state;
+            { // LOCK httpRequestMutex
+                std::unique_lock<std::mutex> lock(httpRequestMutex);
+                if (this->requestingTask == nullptr) {
```
debhaldarMS: >{ [](start = 53, length = 1)

Please follow the current convention of code braces on next line. there are other places in code which needs to be changed in this aspect

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -0,0 +1,743 @@
...
+        jniEnv->CallVoidMethod(httpRequestObject, methodId, nameJstr, valueJstr);
+
+        jniEnv->DeleteLocalRef(nameJstr);
+        jniEnv->DeleteLocalRef(valueJstr);
+    }
+
+    bool PlayFabAndroidHttpPlugin::GetBinaryPayload(RequestTask& requestTask, void*& payload, size_t& payloadSize) const
+    {
+        return false;
+    }
```
debhaldarMS: Add a comment if this is not implemented

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -0,0 +1,743 @@
...
+
+    void PlayFabAndroidHttpPlugin::HandleResults(RequestTask& requestTask)
+    {
+        CallRequestContainer& requestContainer = requestTask.RequestContainer();
+
+        auto callback = requestContainer.GetCallback();
+        if (callback != nullptr)
+        {
+            callback(
```
debhaldarMS: start on the first line after parenthesis

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -0,0 +1,743 @@
+#include <stdafx.h>
+
+#include <jni.h>
+#include <android/log.h>
```
debhaldarMS: >log.h> [](start = 18, length = 6)

Do you need this ?

targets/XPlatCppSdk/source/code/include/playfab/OneDSAndroidHttpPlugin.h
v-hoccha: Yes, you are right.
It should be changed to "okhttp3 though JNI"

---
In reply to: [273684820](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r273684820) [](ancestors = 273684820)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
v-hoccha: Yes, you are right. And to use "enum class" will be better on C++, because enum class is more strictly considered type and conversion by the compiler.

---
In reply to: [273685101](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r273685101) [](ancestors = 273685101)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
v-hoccha: No, "PlayFabAndroidHttpPlugin& operator=(PlayFabAndroidHttpPlugin&& other) = delete;" is removing copy assignment operator.

---
In reply to: [273688270](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r273688270) [](ancestors = 273688270)

targets/XPlatCppSdk/source/code/include/playfab/QoS/XPlatSocket.h
v-hoccha: Yes, the test case of QoS passed.

---
In reply to: [273691254](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r273691254) [](ancestors = 273691254)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
v-hoccha: This is debug assert if there defined NDEBUG. But I didn't implemented Release mode, so this will make assert on release mode.

---
In reply to: [273698342](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r273698342) [](ancestors = 273698342)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
v-hoccha: This time, Android uses 1.6 version. And this is for just let JNI system to know how this JNI plug-in should be handled. So don't need to change this version until we uses new version architectre. And new version of JNI system will be load using 1.6 interface.

---
In reply to: [273701150](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r273701150) [](ancestors = 273701150)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabEvent.cpp
v-hoccha: this change is due to different compiler's type conversion on template. In the template, the compiler strictly checking the type and the auto type conversion is different each compiler, and this "long long int" isn't matched any template types. Because JsonCPP didn't implement "long long int" type explicitly.

---
In reply to: [274661025](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r274661025) [](ancestors = 274661025)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabEvent.cpp
v-hoccha: this change is due to different compiler's type conversion on template. In the template, the compiler strictly checking the type and the auto type conversion is different each compiler, and this "long long int" isn't matched any template types. Because JsonCPP didn't implement "long long int" type explicitly.

---
In reply to: [274661166](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r274661166) [](ancestors = 274661166)

targets/XPlatCppSdk/source/build/Android/XPlatCppSdk/src/main/java/com/microsoft/xplatcppsdk/HttpRequest.java
v-hoccha: This is due to the this method called from JNI. And the exception from JAVA didn't forward to JNI. And after got exception form Java side, all JNI calling is failed. And the JNI is implemented base on C, so there is no exception on it. So we need to just return error with null. And there is only one exception which is throw is "IOException" in bytes().

---
In reply to: [274664851](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r274664851) [](ancestors = 274664851)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
v-hoccha: The reason of using structure are 
1. This is internal data structure which is used by PlayFabAndroidHttpPlugin, so no need encapsulating because this members should be fully accessed.
2. It doesn't need inheritance. And in this pimpl design pattern is usually used straightforward. If it needs to use inheritance this should be used different design pattern.
3. Structure and Class are exactly the same structurally.

---
In reply to: [274670039](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r274670039) [](ancestors = 274670039)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
v-hoccha: No, to call join method, the lock shouldn't be locked or this will be make dead lock. because the thread is trying to lock during working.

---
In reply to: [274690883](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r274690883) [](ancestors = 274690883)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
v-hoccha: No, this is to reduce waiting time to close thread when the plug-in destroyed. This detach is release thread immediately, and if there MakePostRequest() is called during the closing thread, a new thread will create immediately without waiting to closing previouse thread.

---
In reply to: [274691406](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r274691406) [](ancestors = 274691406)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
v-hoccha: Ok, It was changed by Android Studio automatically. I tried to disable the option, but I couldn't failed.

---
In reply to: [274692123](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r274692123) [](ancestors = 274692123)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
v-hoccha: This is for using __android_log_print() to output log. and the __android_log_print are used only Shared Object loading and unloading.
This will be good to keep this log to check the shared object is successfully loaded or not.

---
In reply to: [274693173](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r274693173) [](ancestors = 274693173)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
```diff
@@ -0,0 +1,78 @@
...
+        virtual size_t Update() override;
+
+    protected:
+        struct RequestTask;
+
+        void WorkerThreadEntry();
+        void WorkerThread(void* jniEnv);
+        bool ExecuteRequest(RequestTask& requestTask);
+        bool CheckResponse(RequestTask& requestTask);
+        void SetResponceAsBadRequest(RequestTask& requestTask);
```
MichaelMackendMS:
```suggestion
        void SetResponseAsBadRequest(RequestTask& requestTask);
```

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -0,0 +1,743 @@
...
+            void* payload = nullptr;
+            jbyteArray bodyArray = nullptr;
+
+            if (!GetBinaryPayload(requestTask, payload, payloadSize))
+            {
+                // set string payload if binary wasn't provided
+                std::string requestBody = requestContainer.GetRequestBody();
+                payloadSize = (size_t)requestBody.size();
+
+                if (payloadSize > 0)
```
MichaelMackendMS: The full contents of the code inside this 'if' statement are almost perfectly duplicated starting at line 527.  We should factor out the differing parts and implement the block once.

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -0,0 +1,743 @@
...
+                    std::this_thread::yield();
+                    break;
+                }
+                case RequestTask::Finished: {
+                    if (PlayFabSettings::threadedCallbacks) {
+                        HandleResults(*(this->requestingTask));
+                    } else {
+                        { // LOCK httpRequestMutex
+                            std::unique_lock<std::mutex> lock(httpRequestMutex);
+                            this->pendingResults.push_back(this->requestingTask);
```
MichaelMackendMS:
```suggestion
                            this->pendingResults.push_back(std::move(this->requestingTask));
```

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
```diff
@@ -0,0 +1,743 @@
...
+            return ;
+        }
+        else
+        {
+            JNIEnv* jniEnv = nullptr;
+            jvm->AttachCurrentThread(&jniEnv, nullptr);
+
+            WorkerThread(static_cast<void*>(jniEnv));
+
+            jvm->DetachCurrentThread();
```
MichaelMackendMS: The JVM does full garbage collection on the allocated JVM thread whenever DetatchCurrentThread is called.  Since we only appear to ever operate a single thread at a time, we might consider pooling the thread and keeping it around.  I have no data to back it up, but I strongly suspect that our pendingRequests will most of the time be empty, which means the vast majority of requests will incur the full cost of JVM thread allocation, construction, and garbage collection.

MichaelMackendMS:
I should note that I did build and run the UnitTests on my Android device and all but one test passed.  The failing test was not related to the code changes in this PR.

targets/XPlatCppSdk/source/build/Android/UnitTest/TestAppAndroid.cpp
```diff
@@ -0,0 +1,83 @@
...
+static const char* c_developerSecretKey = "YOUR_DEVELOPER_SECRET_KEY";
+static const char* c_userEmail = "YOUR_EMAIL";
+#endif // USE_EXTERNAL_JSON_FILE
+
+namespace PlayFabUnit
+{
+    bool TestApp::LoadTitleDataJson(std::shared_ptr<char*>& testDataJson, size_t& testDataJsonLen)
+    {
+#ifdef USE_EXTERNAL_JSON_FILE
+        NSString *filePath = [[NSBundle mainBundle] pathForResource:c_jsonFileName ofType:@"json"];
```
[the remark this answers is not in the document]
v-hoccha: This is my mistake I refered iOS implementation. This code should be removed or implemented for Android.

---
In reply to: [275535741](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275535741) [](ancestors = 275535741)

v-hoccha:
Can you give me result?

---
In reply to: [483479318](https://github.com/PlayFab/SDKGenerator/pull/470#issuecomment-483479318) [](ancestors = 483479318)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
```diff
@@ -0,0 +1,78 @@
...
+
+        virtual std::string GetUrl(RequestTask& requestTask) const;
+        virtual void SetPredefinedHeaders(RequestTask& requestTask);
+        virtual void SetHeader(RequestTask& requestTask, const char* name, const char* value);
+        virtual bool GetBinaryPayload(RequestTask& requestTask, void*& payload, size_t& payloadSize) const;
```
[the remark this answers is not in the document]
v-hoccha: This code is working on this case, because we are not refer the returned pointer later. the return value is immediatly copied to new memory. And there will be no misuse.
And your recommendation is duplicate memory.
But I totally agree with you this is little gangerous. And if we have to change this, we also have to change another platform too at once.

---
In reply to: [275580593](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275580593) [](ancestors = 275580593)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
v-hoccha: this is depend on the perspective. I think the garbage collection isn't expensive than keep the resource. If this PlayFab is required time-sensitive response such as web server or game server, I will keep the thread alive. But I think PlayFab isn't time sensitive, and the keep resources until finishing isn't a good pattern if it can return resources to the system. But yes, some times to keep the resource is better, but in this case, if we keep the thread alive, we have to choose between the working thread should keep polling or suspend when there is no request. And the best case is suspending. but to make suspending implementation makes more complex. I think this pattern I implemented is good to keep simple and efficient.  and actually, the garbage collection isn't run every time, it is running when it is required. And, I checked the source code and it seems that to calling DetachCurrentThread() isn't related Java Object. This only related native resources. Anyway, such as using Kernel memory API can make kernel panic when the Kernel memory is completely exhausted. and on the Android easily reaches maximum memory usage.

---
In reply to: [275602778](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275602778) [](ancestors = 275602778)

targets/XPlatCppSdk/source/build/Android/app/src/main/java/com/microsoft/xplatcppsdk/unittest/MainActivity.java
```diff
@@ -0,0 +1,57 @@
...
+        //initialize the TimerTask's job
+        initializeTimerTask();
+
+        //schedule the timer, after the first 5000ms the TimerTask will run every 10000ms
+        timer.schedule(timerTask, 5000, 10000); //
+    }
+
+    public void initializeTimerTask() {
+
+        timerTask = new TimerTask() {
```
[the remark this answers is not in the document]
v-hoccha: I don't mean to bother you, but can you tell me why you prefer Lambda? Yes, Android Studio preferred to use this, and it shows this pattern as Lambda. But I don't know what is more efficient.

---
In reply to: [275573483](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275573483) [](ancestors = 275573483)

targets/XPlatCppSdk/source/build/Android/app/src/main/java/com/microsoft/xplatcppsdk/unittest/MainActivity.java
```diff
@@ -0,0 +1,57 @@
...
+            public void run() {
+
+                //use a handler to run a toast that shows the current timestamp
```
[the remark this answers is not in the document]
v-hoccha: We have to discuss for this whether to implement some UI or not. And responsiblity to implement if we choose to implement.

---
In reply to: [275571748](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275571748) [](ancestors = 275571748)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
```diff
@@ -0,0 +1,78 @@
...
+    protected:
+        struct RequestTask;
+
+        void WorkerThreadEntry();
+        void WorkerThread(void* jniEnv);
+        bool ExecuteRequest(RequestTask& requestTask);
+        bool CheckResponse(RequestTask& requestTask);
+        void SetResponceAsBadRequest(RequestTask& requestTask);
+
+        virtual std::string GetUrl(RequestTask& requestTask) const;
```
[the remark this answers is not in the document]
v-hoccha: This all patterns come from lead platform. Need to think we have to change those all.

---
In reply to: [275577271](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275577271) [](ancestors = 275577271)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
```diff
@@ -0,0 +1,78 @@
...
+            }
+#ifndef DISABLE_ONEDS_API
+            OneDSCallRequestContainer& OneDSRequestContainer()
+            {
+                return *dynamic_cast<OneDSCallRequestContainer*>(requestContainer.get());
+            }
+#endif
+            std::atomic<State> state;
+            std::unique_ptr<CallRequestContainerBase> requestContainer;
+            RequestImpl* impl;
```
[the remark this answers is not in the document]
v-hoccha: Some compiler required whole implement of class when the class is used with template. I will check this and change.

---
In reply to: [275583807](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275583807) [](ancestors = 275583807)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
v-hoccha: It uses shared_ptr, I think it doesn't need to use std::move. Can you explain why I have to use std::move on shared_ptr?

---
In reply to: [275601921](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275601921) [](ancestors = 275601921)

MichaelMackendMS: Sure.  It's purely a performance question.  When you copy a shared_ptr the reference count is adjusted, which is an atomic (and therefore more expensive) operation.  Since you know you're not using the shared_ptr after this, you can avoid the atomic reference adjustements by moving, which will essentially perform a shallow copy the control block and object pointers under the hood.

targets/XPlatCppSdk/source/build/Android/app/src/main/java/com/microsoft/xplatcppsdk/unittest/MainActivity.java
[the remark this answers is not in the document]
MichaelMackendMS: I'm not sure about requirements for UX.  If there is no expectation to have UX, it's perfectly acceptable to simply remove the comment as it is misleading.

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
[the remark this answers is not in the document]
MichaelMackendMS: I don't think you need to worry about the other platforms here. IPlayFabHttpPlugin doesn't declare these virtual methods.  I'm not sure what the other platforms are doing with respect to their interfaces, but strictly from a compiler perspective there's no reason we can't make these const.   There are only two apparent implementations of this method: one for PlayFabAndroidHttpPlugin and one in it's derived class, the OneDSAndroidHttpPlugin.

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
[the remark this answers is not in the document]
MichaelMackendMS: I'm not sure what the sate of your code was when you checked,, but you can run into that error if you don't define the destructor of the class owning the unique_ptr in the implementation file.  Avoiding a drawn out explanation, the solution here would simply be to declare the destructor for RequestTask as you have, but replace the body with "= default;"

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
v-hoccha: But shared_ptr already uses reference count as atomic, and any accessing atomic object or memory will use it. isn't it? 

---
In reply to: [275905830](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275905830) [](ancestors = 275905830)

targets/XPlatCppSdk/source/build/Android/app/src/main/java/com/microsoft/xplatcppsdk/unittest/MainActivity.java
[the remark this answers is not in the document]
MichaelMackendMS: Hah! Well, being candid, neither do I!  I just thought the Android Studio recommendation looked cleaner.  I am not familiar enough with Java to say whether it's better.  If you have a good (or really any) reason to leave it the way it was, that's fine with me.  Being C++ oriented probably just tilted my favor towards lambdas.

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
MichaelMackendMS: Yes, it's certainly a tradeoff question.  I'd be interested to hear from @debhaldarMS on this point, since he would have a better insight as to typical usage than I would.  Still, if we can typically expect a relatively high frequency of API calls from a client, the memory usage may be required often enough to be considered permanent either way.  And if the frequency is at all high, the repetitive allocation and deallocation will only increase the risk of fragmentation, which may actually increase the chance that a larger allocation elsewhere will fail.

MichaelMackendMS:
The test result?  The CloudScriptErorr test failed because it was looking for a keyword in the error message that was not present in the returned result.  But the returned result *was* well-formed and correct.  My suspicion is that the server-side code was altered, but I haven't looked into it any further yet.   I've made a note of the finding and will track it down internally. :)

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
MichaelMackendMS: In the case of a move the control block (which hosts the reference count) will not be dereferenced or accessed at all, thus there will be no need for any atomic operation to occur.

MichaelMackendMS: Here is a conversation on stack overflow discussing this matter: 
https://stackoverflow.com/questions/29643974/using-stdmove-with-stdshared-ptr

targets/XPlatCppSdk/source/build/Android/UnitTest/TestAppAndroid.cpp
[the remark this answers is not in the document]
alex-xbox: Actually I've removed this part for Android, leaving it empty and specifying TODO comment for future, if required. Looks like I somehow have not submitted it, so it's kind of my mistake :)
Raymond, I'd recommend you do the same, just leave it blank with TODO because we don't have proper scenario for packaging Android test app and test it right now.

targets/XPlatCppSdk/source/code/source/playfab/PlayFabAndroidHttpPlugin.cpp
v-hoccha: Ok. I understand what you talk. not using std::move causing two time of atomic operators increase and decrease. To use std::move, there is only one atomic operator.

---
In reply to: [275921176](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275921176) [](ancestors = 275921176)

MichaelMackendMS: Actually, if you use emplace_back instead of push_back you can cut that down to zero because emplace just forwards the rvalue reference and move constructs the new shared_ptr in place.

targets/XPlatCppSdk/source/build/Android/app/src/main/java/com/microsoft/xplatcppsdk/unittest/MainActivity.java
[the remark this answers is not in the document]
v-hoccha: the reason I asking is the lambda has more cost. it is depends on the implementation, when we using lambda it captured context around of its resource, and it also make backward reference. so even if you delete the object wihch using lambda, the object will not remove until the lambda is release it. So mis-using lambda can cause circular reference. But lambda also has good pros too.

---
In reply to: [275914854](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275914854) [](ancestors = 275914854)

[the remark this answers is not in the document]
MichaelMackendMS: Makes sense.  Feel free to leave your code as-is.

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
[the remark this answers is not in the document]
v-hoccha: There are Win, Linux, Xbox platforms.

---
In reply to: [275908206](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275908206) [](ancestors = 275908206)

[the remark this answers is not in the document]
MichaelMackendMS: Those platforms do not inherit from the PlayFabAndroidHttpPlugin.

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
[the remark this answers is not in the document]
v-hoccha: I think I don't completely understand what you are talking, but to use PIMPL to hide the implementation, and this pattern comes from iOS which has an Objective-C object pointer in RequestImpl. So if the compiler make an error this , I will keep using this pattern.

---
In reply to: [275912625](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275912625) [](ancestors = 275912625)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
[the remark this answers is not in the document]
v-hoccha: Oh, I mean consistent with other platforms.

---
In reply to: [275998381](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275998381) [](ancestors = 275998381)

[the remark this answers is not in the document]
MichaelMackendMS: Ah, yes.  I see what you mean.   To that end I would rather see us do it the right way here and let the rest of the platforms play catch-up.  There's no reason our Android SDK shouldn't be our best!

---
In reply to: [275625458](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275625458) [](ancestors = 275625458,275577271)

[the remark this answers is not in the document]
MichaelMackendMS: Ah, yes.  I see what you mean.   To that end I would rather see us do it the right way here and let the rest of the platforms play catch-up.  There's no reason our Android SDK shouldn't be our best!

---
In reply to: [275999701](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275999701) [](ancestors = 275999701,275998381)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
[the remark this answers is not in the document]
MichaelMackendMS: Suggestion #1 does duplicate memory.  
Suggestion #2 manages the memory. 
Suggestion #3 just basically leaves us in the present state but puts the accessor on the object we are reaching into.

---
In reply to: [275613540](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275613540) [](ancestors = 275613540,275580593)

debhaldarMS:
You guys can probably ignore cloudscript errors for now - we're revisiting that design

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
v-hoccha: to change this, it reuires many type changes. I added new backlog, and it will be better to fix once with iOS. [https://microsoft.visualstudio.com/Xbox/_workitems/edit/21252997/](https://microsoft.visualstudio.com/Xbox/_workitems/edit/21252997/)

---
In reply to: [275560116](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275560116) [](ancestors = 275560116,273685101)

targets/XPlatCppSdk/source/code/source/playfab/OneDSAndroidHttpPlugin.cpp
v-hoccha: Need to change all platform to be consistency. Added a backlog. [https://microsoft.visualstudio.com/Xbox/_workitems/edit/21253089](https://microsoft.visualstudio.com/Xbox/_workitems/edit/21253089)

---
In reply to: [273691547](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r273691547) [](ancestors = 273691547)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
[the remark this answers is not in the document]
v-hoccha: Add a backlog for this change on boardl. [https://microsoft.visualstudio.com/Xbox/_workitems/edit/21256316](https://microsoft.visualstudio.com/Xbox/_workitems/edit/21256316)


---
In reply to: [276013227](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r276013227) [](ancestors = 276013227,275999701,275998381)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
```diff
@@ -0,0 +1,78 @@
...
+        void WorkerThread(void* jniEnv);
+        bool ExecuteRequest(RequestTask& requestTask);
+        bool CheckResponse(RequestTask& requestTask);
+        void SetResponceAsBadRequest(RequestTask& requestTask);
+
+        virtual std::string GetUrl(RequestTask& requestTask) const;
+        virtual void SetPredefinedHeaders(RequestTask& requestTask);
+        virtual void SetHeader(RequestTask& requestTask, const char* name, const char* value);
+        virtual bool GetBinaryPayload(RequestTask& requestTask, void*& payload, size_t& payloadSize) const;
+        virtual void ProcessResponse(RequestTask& requestTask, const int httpCode);
```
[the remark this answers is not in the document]
v-hoccha: Add a backlog for this change on boardl. [https://microsoft.visualstudio.com/Xbox/_workitems/edit/21256316](https://microsoft.visualstudio.com/Xbox/_workitems/edit/21256316)

---
In reply to: [275584289](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275584289) [](ancestors = 275584289)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
```diff
@@ -0,0 +1,78 @@
...
+
+        void WorkerThreadEntry();
+        void WorkerThread(void* jniEnv);
+        bool ExecuteRequest(RequestTask& requestTask);
+        bool CheckResponse(RequestTask& requestTask);
+        void SetResponceAsBadRequest(RequestTask& requestTask);
+
+        virtual std::string GetUrl(RequestTask& requestTask) const;
+        virtual void SetPredefinedHeaders(RequestTask& requestTask);
+        virtual void SetHeader(RequestTask& requestTask, const char* name, const char* value);
```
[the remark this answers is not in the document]
v-hoccha: Add a backlog for this change on boardl. [https://microsoft.visualstudio.com/Xbox/_workitems/edit/21256316](https://microsoft.visualstudio.com/Xbox/_workitems/edit/21256316)

---
In reply to: [275578271](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275578271) [](ancestors = 275578271)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
```diff
@@ -0,0 +1,78 @@
...
+        struct RequestTask;
+
+        void WorkerThreadEntry();
+        void WorkerThread(void* jniEnv);
+        bool ExecuteRequest(RequestTask& requestTask);
+        bool CheckResponse(RequestTask& requestTask);
+        void SetResponceAsBadRequest(RequestTask& requestTask);
+
+        virtual std::string GetUrl(RequestTask& requestTask) const;
+        virtual void SetPredefinedHeaders(RequestTask& requestTask);
```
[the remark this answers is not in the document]
v-hoccha: Add a backlog for this change on boardl. [https://microsoft.visualstudio.com/Xbox/_workitems/edit/21256316](https://microsoft.visualstudio.com/Xbox/_workitems/edit/21256316)

---
In reply to: [275578105](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275578105) [](ancestors = 275578105)

targets/XPlatCppSdk/source/build/Android/UnitTest/TestAppAndroid.cpp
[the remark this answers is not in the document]
v-hoccha: Added a backlog [https://microsoft.visualstudio.com/Xbox/_workitems/edit/21256721](https://microsoft.visualstudio.com/Xbox/_workitems/edit/21256721)

---
In reply to: [275923271](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r275923271) [](ancestors = 275923271)

targets/XPlatCppSdk/source/code/include/playfab/PlayFabAndroidHttpPlugin.h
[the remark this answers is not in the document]
MichaelMackendMS: Still not comfortable with this function design, but I'll resolve it for the moment.


---
In reply to: [276013635](https://github.com/PlayFab/SDKGenerator/pull/470#discussion_r276013635) [](ancestors = 276013635,275613540,275580593)

MichaelMackendMS:
I have completed my review.  @debhaldarMS  should approve if he is satisfied.

PR #471, opened 2019-04-09
Overhauling the Unity Context and Settings for multi-instance and multi-login

pgilmorepf:
... so that the tests will pass again, and the test corruption is significantly reduced.
Removing ancient unused files.
Fixing some template spacing abuse
Converting Unity Client tests to instance tests (Found a million errors)
Slight tweaks to all tests and their data loading cleanliness.
Removing redundant step in LightweightEventsTests
Clarifying failure in LightweightEventsTests
ServerInstance tests got way simpler when you don't have 12 independent places to set dev secret key.
UUnitIncrementalTestRunner now uses a client api instance for isolation of the final result upload.
Better request/result checking.
Generic CloudScript method works again.
Removed a ton of super terrible redundant cross-cross-cross-checking for input credentials
JS syntax fixes
Removed unnecessary spammy log
Removed a few unused usings
PlayFabDeviceUtil does the correct thing when triggered from an instance (totally wrong before)
Fixed some tabbing and spacing on #ifs and #regions
Removed some obsolete stuff we said we'd remove 2 years ago
Removed IPlayFabTransportPlugin and its yet-another-redundant way to save credentials
Making VerticalName stuff internal because it never worked, and should never have been used by customers
Adding a new PlayFabBaseModel and a for-convienience ToJson() Method
Removing some comments about how IPlayFabTransportPlugin is deprecated and will be removed
LightweightEvents stuff will call APIs correctly with the correct credentials
Cleaned up the 4xRedirects in PlayFabSettings (Not better, just easier to follow)
A bunch of stuff to make lots of nasty code look smaller and nicer
VS default spacing and tabbing fixes
Most settings are stored in one master place, and redirects from old legacy variables work in a more code-traceable way
Added the Assembly stripping thing requested by customer
Static and Instance APIs have the correct flow of credentials, and don't leak, and don't spill


targets/unity-v2/Testing/Tests/Client/ClientApiTests.cs
```diff
@@ -11,34 +12,45 @@ public class ClientApiTests : UUnitTestCase
     {
         private Action _tickAction = null;
         private TestTitleDataLoader.TestTitleData testTitleData;
+        private static readonly PlayFabApiSettings clientSettings = new PlayFabApiSettings();
+        private static readonly PlayFabAuthenticationContext clientContext = new PlayFabAuthenticationContext();
+        //private static readonly PlayFabClientInstanceAPI clientInstance = new PlayFabClientInstanceAPI(PlayFabSettings.staticSettings, PlayFabSettings.staticPlayer);
```
toddbellMSFT: should we remove commented out code? Or is this meant for toggleable tests?

PR #481, opened 2019-04-12
bypass cloudscript listener in jenkins

toddbellMSFT:
Jenkins machines have been held up due to cloudscript limitations in the past. This commit will just output an xml file to be picked up by the cloudscript listener instead of attempting to route around an ExecuteCloudscript.

This is essentially a temporary solution. The correct solution (which will take time) would be to a.) chunk tests into separate files so ExecuteCloudscript doesn't error out with a size limitation of 10KB (but how would we get the JenkinsUtitlity to read these new chunks?), b.) upgrade our test app to allow this kind of limit (but would we just see this issue later on?).

When we start adding more tests, this problem will become much worse. This current workaround will incurr a test run cost (it takes time to output a "large" 10KB xml file. We will be adding more tests, so this file will take longer to output.

### DIFF
--- a/JenkinsConsoleUtility/jcuSrc/Commands/CloudScriptListener.cs
+++ b/JenkinsConsoleUtility/jcuSrc/Commands/CloudScriptListener.cs
@@ -50,6 +50,12 @@ namespace JenkinsConsoleUtility.Commands
             verbose = bool.Parse(JenkinsConsoleUtility.GetArgVar(argsLc, "verbose", "false"));
             _getRequest = new CsGetRequest { customId = buildIdentifier };
 
+            // shortcut if the file already got output
+            if (File.Exists(workspacePath + "\\test.xml"))
+            {
+                return 0;
+            }
+
             JenkinsConsoleUtility.FancyWriteToConsole("Begin CloudScriptListener", null, ConsoleColor.Gray);
             var returnCode = Login(testTitleData.titleId, buildIdentifier, testTitleData);
             if (returnCode != 0)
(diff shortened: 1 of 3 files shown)


JenkinsConsoleUtility/jcuSrc/Commands/CloudScriptListener.cs
```diff
@@ -50,6 +50,12 @@ public int Execute(Dictionary<string, string> argsLc, Dictionary<string, string>
             verbose = bool.Parse(JenkinsConsoleUtility.GetArgVar(argsLc, "verbose", "false"));
             _getRequest = new CsGetRequest { customId = buildIdentifier };
 
+            // shortcut if the file already got output
+            if (File.Exists(workspacePath + "\\test.xml"))
```
MichaelMackendMS:
```suggestion
            if (File.Exists(Path.Combine(workspacePath, "test.xml")))
```

targets/unity-v2/Testing/Tests/Shared/Uunit/UUnitIncrementalTestRunner.cs
```diff
@@ -97,9 +97,17 @@ private void OnLoginSuccess(LoginResult result)
...
 
+            SaveReportToLocalWorkspace(suite.GetInternalReport());
+
             clientInstance.ExecuteCloudScript(request, OnCloudScriptSubmit, OnPostTestResultsError, null, testTitleData.extraHeaders);
         }
 
+        private void SaveReportToLocalWorkspace(TestSuiteReport testSuiteReport)
+        {
+            // TODO: this should put the file under our workspace directory
+            JUnitXml.WriteXmlFile(Directory.GetCurrentDirectory() + "..\\..\\..\\test.xml", new List<TestSuiteReport>() { testSuiteReport }, true);
```
MichaelMackendMS: Not sure what to do about this path combination.  It looks super fragile.  Maybe consider defining a higher level root as an anchor like a workspace environment variable.

toddbellMSFT: We will address this issue on Monday.

PR #482, opened 2019-04-15
Add Unreal CPP & BP SDK support for Local Debugging Azure Functions (#4)

nima-ap:
Support for local debugging has been added in both the C++ and Blueprints SDKs in the Unreal Marketplace Plugin. The feature is summarized as follows:
C++:
* Following the same pattern as the Unity and C# SDKs, we have created a series of utility functions in `PlayFabUtilities.cpp` that allow us to extract the location of the TEMP dir in a user's device (mac and pc), find the local settings file, extract the `LocalApiServer` field, and attempt to send the `ExecuteFunctionRequest` to that address instead of the remote PlayFab address.
BP:
* The logic here is separated out into two parts, and does not follow the same exact approach as the other SDK work we've done for this feature before. This is mainly due to the fact that the process of actually sending over the HTTP request in the BP SDK is decoupled from the API method. As a result, the API method only takes care of assigning a value to a field that represents a full url being present, and the `Activator` method then looks at that field and if it's assigned decides to override the API url with it.


targets/UnrealMarketplacePlugin/makeBp.js
```diff
@@ -296,3 +298,29 @@ function getAuthBools(tabbing, apiCall) {
 
     return output;
 }
+
+function getCustomApiAssignmentLogic(tabbing, apiCall) {
+    if (apiCall.name === "ExecuteFunction")
```
pgilmorepf: I don't think this is a strict enough check. api.name and apiCall.name would be specific enough.

targets/UnrealMarketplacePlugin/makeBp.js
```diff
@@ -296,3 +298,29 @@ function getAuthBools(tabbing, apiCall) {
...
+    {
+        return tabbing + "// Check for local debugging\n"
+            + tabbing + "FString localApiServer = PlayFabCommon::PlayFabCommonUtils::GetLocalSettingsFileProperty(TEXT(\"LocalApiServer\"));\n"
+            + tabbing + "if (!localApiServer.IsEmpty())\n"
+            + tabbing + "{\n"
+            + tabbing + "    FString endpoint = TEXT(\"" + apiCall.url + "\");\n"
+            + tabbing + "    endpoint.RemoveFromStart(TEXT(\"/\"));\n"
```
pgilmorepf: This is a line that shouldn't be necessary if we're doing the url string construction consistently in all places.

targets/UnrealMarketplacePlugin/makeBp.js
```diff
@@ -296,3 +298,29 @@ function getAuthBools(tabbing, apiCall) {
...
+            + tabbing + "    FString url = localApiServer + endpoint;\n"
+            + tabbing + "    manager->PlayFabRequestFullURL = url;\n"
+            + tabbing + "}\n"
+    }
+}
+
+function getCustomApiActivationUrlLogic(api) {
+    if (api.name === "CloudScript")
```
pgilmorepf: I don't think this is a strict enough check. api.name and apiCall.name would be specific enough.

targets/UnrealMarketplacePlugin/makeCpp.js
```diff
@@ -660,6 +661,20 @@ function getRequestActions(tabbing, apiCall, isInstanceApi) {
     return "";
 }
 
+function getCustomApiLogic(tabbing, api, apiCall, isInstanceApi) {
+    if (apiCall.name === "ExecuteFunction")
```
pgilmorepf: I don't think this is a strict enough check. api.name and apiCall.name would be specific enough.

targets/UnrealMarketplacePlugin/makeCpp.js
```diff
@@ -660,6 +661,20 @@ function getRequestActions(tabbing, apiCall, isInstanceApi) {
...
+        return tabbing + "FString localApiServer = PlayFabSettings::GetLocalApiServer();\n"
+            + tabbing + "if (!localApiServer.IsEmpty())\n"
+            + tabbing + "{\n"
+            + tabbing + "    FString endpoint = TEXT(\"" + apiCall.url + "\");\n"
+            + tabbing + "    endpoint.RemoveFromStart(TEXT(\"/\"));\n"
```
pgilmorepf: Url string construction - ditto.

targets/UnrealMarketplacePlugin/source/PlayFab/Source/PlayFabCommon/Private/PlayFabCommmonUtils.cpp.ejs
```diff
@@ -0,0 +1,46 @@
+﻿<%- copyright %>
```
pgilmorepf: My concerns with this file are that it may not compile for any platform except windows.
Have you tested this at all with other platforms?
It doesn't have to work, but it must still compile.

targets/UnrealMarketplacePlugin/source/PlayFab/Source/PlayFabCommon/Public/PlayFabCommonSettings.h.ejs
```diff
@@ -38,7 +38,7 @@ namespace PlayFabCommon
         {
             if (serverURL.Len() == 0)
             {
-                serverURL = TEXT("https://") + (!verticalName.IsEmpty() ? verticalName : titleId) + productionEnvironmentURL;
+                serverURL = TEXT("https://") + titleId + (verticalName.IsEmpty() ? "" : "." + verticalName) + productionEnvironmentURL;
```
pgilmorepf: This isn't correct.
Revert this line, because the correct fix is more complex, and if it's not the correct change, then we shouldn't bother with it.

targets/UnrealMarketplacePlugin/templates/PlayFab/PlayFab_API.h.ejs
```diff
@@ -93,12 +93,16 @@ public:
...
+<% if (api.name === "CloudScript") {
+%>    FString PlayFabRequestFullURL;
+<% } %>
     bool useEntityToken = false;
     bool useSecretKey = false;
     bool useSessionTicket = false;
     bool returnsSessionTicket = false;
     bool returnsEntityToken = false;
 
+
```
pgilmorepf: Please don't add double-line-breaks.

targets/unity-v2/source/Shared/Internal/Util.cs
```diff
@@ -140,10 +140,8 @@ internal static string GetLocalSettingsFileProperty(string propertyKey)
                     object result;
```
pgilmorepf: Please put Unity changes into a separate Pull Request.

targets/UnrealMarketplacePlugin/source/PlayFab/Source/PlayFabCommon/Public/PlayFabCommonSettings.h.ejs
nima-ap: Can you please explain why this isn't correct?
Based on my understanding, the URL construction we have is title.vertical.playfabapi.com.
With the current URL construction achieving title.vertical is pretty much impossible and so it results in us not being able to use this code for testing the feature on our own private verticals.

targets/UnrealMarketplacePlugin/makeBp.js
nima-ap: I believe that everywhere else you suggested this it works but not in this case.
The context is the API here not the call. There is no `Activate` method per endpoint, but rather the `CloudScript` API as a whole. As a result, what you are requesting cannot be done in this scenario specifically since there is no `apiCall` but rather just an `api`.

targets/UnrealMarketplacePlugin/makeBp.js
nima-ap: Resolving this as per our conversation in person.

targets/UnrealMarketplacePlugin/makeCpp.js
nima-ap: Resolving this as per our conversation in person.

targets/UnrealMarketplacePlugin/source/PlayFab/Source/PlayFabCommon/Public/PlayFabCommonSettings.h.ejs
nima-ap: Furthermore, I'd like to remind you about the fact that a similar change was checked into the unity-v2 target a little earlier as part of our [work for local-debugging in Unity](https://github.com/PlayFab/SDKGenerator/commit/7ed2127df3ccf74a7a7441c57baa43587a2dcc30).

pgilmorepf: Let's talk in person about the url construction.
In short: it's way more complicated.

targets/UnrealMarketplacePlugin/makeBp.js
pgilmorepf: Got it, thanks. I got overzealous with the copy paste. This specific one is resolved, but I think I'm right about the others.

targets/UnrealMarketplacePlugin/source/PlayFab/Source/PlayFabCommon/Public/PlayFabCommonSettings.h.ejs
nima-ap: Resolving as per our email thread conversation regarding keeping this code the way it is and renaming `vertical` to `cloud` in a subsequent PR either by our team or yours.

nima-ap:
@pgilmorepf @toddbellMSFT There was an issue with the branching structure I was adapting on my fork. In short, we were merging changes from our target specific branches into our master and then sending PRs from our master branch to your master branch. This lead to issues such as inter-target changes bleeding across PRs as has happened here (with unity). As a result, I am closing this PR and sending out another one shortly. Please note that all your comments here have been resolved in the new PR except the fact that we still haven't tested the compilation on other platforms, something I will hopefully get to this week.

PR #483, opened 2019-04-15
Jstatz/macos fix

kf-jstatz:
This shuffles files around within the modules, with the goal of eliminating the circular dependency error when loading the project in Unreal Editor specifically on MacOS.  The "right" thing to do would probably be to put the blueprint stuff into the test module, but I'm told we can't do that, which pretty much means the test module has to come to the blueprints instead.

Ultimately almost everything in Test ends up dependent on other things in Test, and ultimately I eliminated that module as it was basically empty by the time all the circularity was gone.


targets/UnrealMarketplacePlugin/source/PlayFab/Source/PlayFab/Private/PlayFab.cpp.ejs
```diff
@@ -35,3 +35,4 @@ class FPlayFab : public IPlayFab
 IMPLEMENT_MODULE(FPlayFab, PlayFab)
 
 DEFINE_LOG_CATEGORY(LogPlayFab);
+DEFINE_LOG_CATEGORY(LogPlayFabTests);
```
pgilmorepf: This is slightly confusing to me.
You are adding LogPlayFabTests here, but you're removing multiple uses of it up above.
Can you clarify that?

targets/UnrealMarketplacePlugin/examplesource/ExampleProject/Plugins/PlayFab/Source/PlayFabCommon/Public/TestFramework/PlayFabTestContext.h.ejs
```diff
@@ -116,7 +115,7 @@ public:
     UFUNCTION()
     void EndTest(PlayFabApiTestFinishState InFinishState = PlayFabApiTestFinishState::PASSED, FString InMessage = "")
     {
-        UE_LOG(LogPlayFabTests, Log, TEXT("Ending Test: %s."), *testName);
+        UE_LOG(LogPlayFabCommon, Log, TEXT("Ending Test: %s."), *testName);
```
pgilmorepf: This is slightly confusing to me.
You are adding LogPlayFabTests below, but you're removing multiple uses of it up here.
Can you clarify that?

targets/UnrealMarketplacePlugin/examplesource/ExampleProject/Plugins/PlayFab/Source/PlayFabCommon/Private/TestFramework/PlayFabTestRunner.cpp.ejs
```diff
@@ -94,7 +94,7 @@ void IPlayFabTestRunner::Run(const float InDeltaTime)
         {
             numberOfTests++;
 
-            UE_LOG(LogPlayFabTests, Log, TEXT("Starting test: %s"), *CurrentTestData->testName);
+            UE_LOG(LogPlayFabCommon, Log, TEXT("Starting test: %s"), *CurrentTestData->testName);
```
pgilmorepf: This is slightly confusing to me.
You are adding LogPlayFabTests below, but you're removing multiple uses of it up here.
Can you clarify that?

kf-jstatz:
I'm pretty sure it ended up that way, because if I want to use
LogPlayFabTests inside those files, they would have to include PlayFab.h,
and that means the Common module needs a dependency on the PlayFab module,
and we're back to a circular dependency again.  It looked to me like Common
is where a lot of the working logic is meant to go, so I was trying to put
the guts of the tests there.  I put the CPP test implementations into the
PlayFab module so that they were alongside the Blueprint tests, since as an
outside observer I'd expect those things to be side by side.

So in order to do this, everything would have go to into the PlayFab
module, and none of the test stuff could be in Common.

PR #484, opened 2019-04-15
Turn off UnitTests and only log strings for failures

toddbellMSFT:


targets/unity-v2/Testing/Tests/Shared/Uunit/UUnitTestContext.cs
```diff
@@ -45,7 +45,11 @@ public UUnitTestContext(UUnitTestCase testInstance, Action<UUnitTestContext> tes
         public void EndTest(UUnitFinishState finishState, string resultMsg)
         {
             EndTime = DateTime.UtcNow;
-            TestResultMsg = resultMsg;
```
pgilmorepf: Plz don't change this. Even pass messages can be vital for diagnosing failures of tests in the same suite.

PR #485, opened 2019-04-15
Update SDKBuildScripts/unreal.sh

toddbellMSFT:
new script that can be called immediatley when pulled to build Unreal on a mac.


.gitignore
```diff
@@ -6,6 +6,9 @@
 **/.idea/
 **/.vs/
 **/node_modules/
+**/beta_*.bat
```
pgilmorepf: Build scripts has its own .gitignore:
https://github.com/PlayFab/SDKGenerator/blob/master/SDKBuildScripts/.gitignore

SDKBuildScripts/Linux/NewTarget.sh
```diff
@@ -0,0 +1,6 @@
+SdkName="NewTarget"
+targetSrc="newTarget"
+delSrc=false
```
pgilmorepf: ```delSrc=false``` is equivalent to <default> so I'd just delete those lines in the sh files.

SDKBuildScripts/Linux/actionscript_build.sh
```diff
@@ -0,0 +1,7 @@
+SdkName="ActionScriptSDK"
+targetSrc="actionscript"
+delSrc=true
```
pgilmorepf: Obviously keep it when it's true.

PR #488, opened 2019-04-22
Add CSharp SDK support for Local Debugging Azure Functions

nima-ap:
Support for local debugging has been added in the CSharp SDK. 

The feature follows the same patterns as the previous work on the Unity SDK with the exception that there are certain #ifdefine tags added to allow local debugging only if the target framework is .NET 4.5 or NETCore 2.0 as providing such capabilities in .NETPortable and others is neither feasible nor possible with our implementation.

### DIFF
--- a/targets/csharp/templates/API.cs.ejs
+++ b/targets/csharp/templates/API.cs.ejs
@@ -18,6 +18,7 @@ namespace PlayFab
 %>        public static async Task<PlayFabResult<<%- apiCall.result %>>> <%- apiCall.name %>Async(<%- apiCall.request %> request, object customData = null, Dictionary<string, string> extraHeaders = null)
         {
 <%- getRequestActions("            ", apiCall, false) %>
+<%- getCustomApiLogic("            ", apiCall) %>
             var httpResult = await PlayFabHttp.DoPost("<%- apiCall.url %>", request, <%- getAuthParams(apiCall) %>, extraHeaders);
             if (httpResult is PlayFabError)
             {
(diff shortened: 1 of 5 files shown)


nima-ap:
@pgilmorepf Could you please add yourself as a reviewer on this? This is the CSharp SDK changes that were put on hold last minute before GDC in favor of lack of time for the tests as you specified.

nima-ap:
@toddbellMSFT @MichaelMackendMS 
Hey guys :) Paul was interested in getting another pair of eyes on this too. Could one of you please review it as well so that I can proceed with merging it? Thanks!

targets/csharp/templates/API.cs.ejs
```diff
@@ -18,6 +18,7 @@ namespace PlayFab
 %>        public static async Task<PlayFabResult<<%- apiCall.result %>>> <%- apiCall.name %>Async(<%- apiCall.request %> request, object customData = null, Dictionary<string, string> extraHeaders = null)
         {
 <%- getRequestActions("            ", apiCall, false) %>
+<%- getCustomApiLogic("            ", apiCall) %>
```
toddbellMSFT: nitpick: I know this line follows current conventions, but it seems difficult to read exactly how many spaces we expect to tab in. 

Could we convert this parameter to an int? Or have a util function that can convert from an int to output that tabbing as a string? Would be a bit easier to read.

nima-ap: I'd be more than glad to vote in favor of this! However, one thing I'm concerned about in regards to this PR is that since it's a bigger scale change to (all) make.js files, maybe we should address that separately later?

PR #494, opened 2019-04-25
#define away members that are only used per api

toddbellMSFT:
dev secret key is only used in either server or admin api's; advert members are only needed in client side code.

### DIFF
--- a/targets/csharp/source/source/PlayFabSettings.cs.ejs
+++ b/targets/csharp/source/source/PlayFabSettings.cs.ejs
@@ -38,8 +38,7 @@ namespace PlayFab
         [Obsolete("Moved to PlayFabSettings.staticSettings.TitleId")]
         public static string TitleId { get { return staticSettings.TitleId; } set { staticSettings.TitleId = value; } }
 
-#if !DISABLE_PLAYFABCLIENT_API
-<% if (hasClientOptions) { %>
+<% if (hasClientOptions) { %>#if !DISABLE_PLAYFABCLIENT_API
         [Obsolete("Moved to PlayFabSettings.staticSettings.AdvertisingIdType")]
         public static string AdvertisingIdType { get { return staticSettings.AdvertisingIdType; } set { staticSettings.AdvertisingIdType = value; } }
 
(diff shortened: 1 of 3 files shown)


targets/csharp/source/source/PlayFabSettings.cs.ejs
```diff
@@ -30,11 +30,15 @@ namespace PlayFab
...
 
+#if ENABLE_PLAYFABSERVER_API || ENABLE_PLAYFABADMIN_API
         [Obsolete("Moved to PlayFabSettings.staticSettings.DeveloperSecretKey")]
         public static string DeveloperSecretKey { get { return staticSettings.DeveloperSecretKey; } set { staticSettings.DeveloperSecretKey = value; } }
+#endif
 
         [Obsolete("Moved to PlayFabSettings.staticSettings.TitleId")]
         public static string TitleId { get { return staticSettings.TitleId; } set { staticSettings.TitleId = value; } }
+
+#if !DISABLE_PLAYFABCLIENT_API
```
pgilmorepf: I'd put the #ifs inside of the <% if (hasClientOptions) { %> ejs, because otherwise, you'll have:

```
#endif
```
In the server output source.
Generate this SDK and make sure you like all the changes.

targets/csharp/templates/PlayFabApiSettings.cs.ejs
```diff
@@ -8,23 +8,31 @@ namespace PlayFab
...
+
+#if ENABLE_PLAYFABSERVER_API || ENABLE_PLAYFABADMIN_API
         /// <summary> You must set this value for PlayFabSdk to work properly (Found in the Game Manager for your title, at the PlayFab Website) </summary>
         public string DeveloperSecretKey = null;
+#endif
+
         /// <summary> You must set this value for PlayFabSdk to work properly (Found in the Game Manager for your title, at the PlayFab Website) </summary>
         public string TitleId;
+
+#if !DISABLE_PLAYFABCLIENT_API
```
pgilmorepf: Same here.
Please look at the output diff in the repo itself and make sure it looks appropriate.

toddbellMSFT:
retest

PR #500, opened 2019-05-03
Adding a sleep and retry to clone operations

pgilmorepf:


MichaelMackendMS:
Do we understand the underlying cause as to why our first attempt might fail?

pgilmorepf:
Yes. Multiple Jenkins jobs are trying to touch the same locations at potentially the same times. Every once in a great while they try to lock a repo at exactly the same time.

PR #502, opened 2019-05-06
Add SDK generator powershell script

veleek:
Adding a script to enable quick and easy (and documented) generation of the various SDKs from Powershell.

### DIFF
--- a/SDKBuildScripts/Update-PlayFabSdk.ps1
+++ b/SDKBuildScripts/Update-PlayFabSdk.ps1
@@ -53,32 +53,32 @@ Indicates that we should generate the SDK for a platform that does not support n
 Indicates whether or not to include any APIs tagged with as beta.
 
 .EXAMPLE
+Update-PlayFabSdk -SdkNames CSharpSDK
+
 Run the PlayFab API generation for the CSharp SDK using the API specification documents from
 the PlayFab git repository.
 
-    PS> Update-PlayFabSdk -SdkNames CSharpSDK
-
 .EXAMPLE
+Update-PlayFabSdk CSharpSDK,JavaScriptSDK -Vertical example
+
 Run the PlayFab API generation for the CSharp and JavaScript SDKs using the API specification
 documents from the example vertical and include all of the beta SDKs.
-
-    PS> Update-PlayFabSdk CSharpSDK,JavaScriptSDK -Vertical example
 
 This is functionally identical to using -ApiSpecUrl "https://example.playfabapi.com/apispec"
 
 .EXAMPLE
+Update-PlayFabSdk CSharpSDK,JavaScriptSDK -Vertical example -Cloud sample
+
 Run the PlayFab API generation for the CSharp and JavaScript SDKs using the API specification
 documents from the example vertical and the sample cloud.
-
-    PS> Update-PlayFabSdk CSharpSDK,JavaScriptSDK -Vertical example -Cloud sample
 
 This is functionally identical to using -ApiSpecUrl "https://example.sample.playfabapi.com/apispec"
 
 .EXAMPLE
+Update-PlayFabSdk UnitySDK -Beta
+
 Run the PlayFab API generation for the CSharp and JavaScript SDKs using the API specification
 documents from the example vertical and include all of the beta SDKs.
-
-    PS> Update-PlayFabSdk UnitySDK -Beta
 
 #>
 [CmdletBinding(SupportsShouldProcess = $true, DefaultParameterSetName = "ApiSpecGitUrl")]
--- a/SDKBuildScripts/Update-PlayFabSdk.ps1
+++ b/SDKBuildScripts/Update-PlayFabSdk.ps1
@@ -1,0 +1,160 @@
+[CmdletBinding(SupportsShouldProcess = $true, DefaultParameterSetName = "ApiSpecGitUrl")]
+param(
+    [Parameter(Position = 1)]
+    [ValidateSet(
+        "ActionScriptSDK",
+        "Cocos2d-xSDK",
+        "CSharpSDK",
+        "JavaSDK",
+        "JavaScriptSDK",
+        "LuaSDK",
+        "NodeSDK",
+        "Objective_C_SDK",
+        "PhpSDK",
+        "PostmanCollection",
+        "PythonSDK",
+        "SdkTestingCloudScript",
+        "UnrealMarketplacePlugin",
+        "UnitySDK",
+        "WindowsSDK",
+        "XPlatCppSDK")]
+    [string[]]$SdkNames,
+    [Parameter(ParameterSetName="ApiSpecPath")]
+    [string]$ApiSpecPath,
+    [Parameter(ParameterSetName="ApiSpecPfUrl")]
+    [string]$ApiSpecPfUrl,
+    [Parameter(ParameterSetName="ApiSpecPfUrlBranch")]
+    [string]$ApiSpecBranch,
+    [Parameter(ParameterSetName="ApiSpecGitUrl")]
+    [switch]$ApiSpecGitUrl,
+    [switch]$KeepSource,
+    [switch]$NonNullable,
+    [switch]$Beta
+)
+
+begin
+{
+    if(!(Get-Command node))
+    {
+        throw "You must have Node.js installed to generate the SDK"
+    }
+
+    $sdkTargetSrcMap = @{
+        "ActionScriptSDK" = "actionscript";
+        "Cocos2d-xSDK" = "cpp-cocos2dx";
+        "CSharpSDK" = "csharp";
+        "JavaSDK" = "java";
+        "JavaScriptSDK" = "javascript";
+        "LuaSDK" = "LuaSdk";
+        "NodeSDK" = "js-node";
+        "Objective_C_SDK" = "objc";
+        "PhpSDK" = "PhpSdk";
+        "PostmanCollection" = "postman";
+        "PythonSDK" = "PythonSdk";
+        "SdkTestingCloudScript" = "SdkTestingCloudScript";
+        "UnrealMarketplacePlugin" = "UnrealMarketplacePlugin";
+        "UnitySDK" = "unity-v2";
+        "WindowsSDK" = "windowssdk";
+        "XPlatCppSDK" = "xplatcppsdk";
+    }
+
+    $sdksPath = Resolve-Path (Join-Path $PSScriptRoot "..\..\sdks")
+    if(!(Test-Path $sdksPath))
+    {
+        mkdir $sdksPath
+    }
+
+    if($ApiSpecBranch)
+    {
+        $ApiSpecPfUrl = "https://$($ApiSpecBranch).playfabapi.com/apispec"
+    }
+
+    if($ApiSpecPath)
+    {
+        if(!(Test-Path $ApiSpecPath))
+        {
+            throw "Unable to find ApiSpecPath '$ApiSpecPath'."
+        }
+
+        $apiSpecSource = "-apiSpecPath $ApiSpecPath"
+    }
+    elseif($ApiSpecPfUrl)
+    {
+        if(![Uri]::IsWellFormedUriString($ApiSpecPfUrl, "Absolute"))
+        {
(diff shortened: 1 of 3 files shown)


SDKBuildScripts/Update-PlayFabSdk.ps1
```diff
@@ -0,0 +1,160 @@
...
+)
+
+begin
+{
+    if(!(Get-Command node))
+    {
+        throw "You must have Node.js installed to generate the SDK"
+    }
+
+    $sdkTargetSrcMap = @{
```
pgilmorepf: I might come back to you later and ask that you read this information from a different place. But it's certainly useful now.

SDKBuildScripts/Update-PlayFabSdk.ps1
```diff
@@ -0,0 +1,160 @@
...
+
+    $sdksPath = Resolve-Path (Join-Path $PSScriptRoot "..\..\sdks")
+    if(!(Test-Path $sdksPath))
+    {
+        mkdir $sdksPath
+    }
+
+    if($ApiSpecBranch)
+    {
+        $ApiSpecPfUrl = "https://$($ApiSpecBranch).playfabapi.com/apispec"
```
pgilmorepf: This URL is incomplete and mis-labeled.
It should properly be:
https://($verticalName).$(clusterName).playfabapi.com/apispec"
There's also alternate formats that apply when one or the other is undefined.

I'd go so far as to say this is at minimum, misleading.
(I also understand that you're probably reading from other public code, which is also misleading in the same way, and we're trying to fix that).

SDKBuildScripts/Update-PlayFabSdk.ps1
```diff
@@ -0,0 +1,160 @@
...
+        "XPlatCppSDK" = "xplatcppsdk";
+    }
+
+    $sdksPath = Resolve-Path (Join-Path $PSScriptRoot "..\..\sdks")
+    if(!(Test-Path $sdksPath))
+    {
+        mkdir $sdksPath
+    }
+
+    if($ApiSpecBranch)
```
pgilmorepf: The better condition might be: If ApiSpecPfUrl is not already defined...

SDKBuildScripts/Update-PlayFabSdk.ps1
```diff
@@ -0,0 +1,160 @@
...
+    {
+        $apiSpecSource = "-apiSpecGitUrl"
+    }
+}
+
+process
+{
+    foreach($sdkName in $SdkNames)
+    {
+        $targetSrc = $sdkTargetSrcMap[$sdkName]
```
pgilmorepf: For non-public repos, it's important to be able to accept this information as a parameter, if it's not present in the public dictionary.

generate.js
```diff
@@ -95,7 +95,7 @@ function parseCommandInputs(args, argsByName, errorMessages, targetOutputPathLis
     // Output an error if no targets are defined
```
pgilmorepf: It looks like submitting this file was probably an accident, as it's a generated file, and not a source file.
Can you revert it?

SDKBuildScripts/Update-PlayFabSdk.ps1
veleek: Sounds good to me!

SDKBuildScripts/Update-PlayFabSdk.ps1
veleek: @pgilmorepf I'm fine with changing this.  This was literally intended to be almost identical to the existing scripts, but I'm renaming stuff to make it much more clear so I'm happy to make improvements.

However I think you got one of the cloud/vertical things backwards.  If you specify https://foo.playfabapi.com you're looking at the foo vertical in the main cloud.  You can verify this by looking at `https://spi-economy.spi.playfabapi.com/healthstatus` vs `https://spi-economy.playfabapi.com/healthstatus`.  In both of these cases, it returns the vertical `spi-economy`.

SDKBuildScripts/Update-PlayFabSdk.ps1
veleek: ApiSpecUrl and ApiSpecCloud/Vertical are defined in separate parameter sets, so it's actually impossible to pass BOTH parameters at the same time so there's no need to check that.

SDKBuildScripts/Update-PlayFabSdk.ps1
veleek: I don't quite understand how this is used.  Will the user still need to pass in `$sdkName` CSharpSDK and then some other local path?  Or does it make sense to just use sdkName as targetSrc if it doesn't exist in the map?

This is passed into generate.js as:
```
node generate.js "$targetSrc=$destPath" ...
```

What does this represent when the user provides a value?

generate.js
veleek: My change to this file was to remove the 5 second delay at the end.  I don't think it's necessary is it?  And if not where should I go to remove it?

pgilmorepf: Two things:
1. It's there for my personal debugging purposes, based on nuances of my environment (node and VS don't always cooperate as much as I'd like).
I get that it's not useful to most anybody else though.
2. Changing it in the .js doesn't work. You have to change the .ts or it'll just come back the next time I load the solution.

SDKBuildScripts/Update-PlayFabSdk.ps1
pgilmorepf: Replied privately.

SDKBuildScripts/Update-PlayFabSdk.ps1
```diff
@@ -0,0 +1,262 @@
...
+                "Would you like to clone the $sdkName repository from $($repoPath) into '$destPath'?",
+                "Unable to find $sdkName repository")
+              )
+            {
+                git clone $repoPath $destPath
+            }
+        }
+
+        # Add any overrides for a specific SDK
+        if($sdkName -eq "UE4")
```
pgilmorepf: I think the comparison is "UnrealMarketplacePlugin", not "UE4"

pgilmorepf: And also, you don't need an input for NonNullable at the top.
UE4 is explicitly the only non-nullable SDK right now.

SDKBuildScripts/Update-PlayFabSdk.ps1
```diff
@@ -0,0 +1,293 @@
...
+
+        $sdkGenArgs = ""
+        if($NonNullable)
+        {
+            $sdkGenArgs += "-flags nonnullable "
+        }
+
+        if($Beta)
+        {
+            $sdkGenArgs += "-flags beta "
```
pgilmorepf: We need a condition where if it's both, the sdkGenArgs are
-flags beta nonnullable
(It won't properly handle -flags beta -flags nonnulable, the new nonnullable will clobber the earlier beta declaration.

SDKBuildScripts/Update-PlayFabSdk.ps1
```diff
@@ -0,0 +1,293 @@
...
+    [AllowEmptyString()]
+    [string]$ApiSpecGitUrl,
+    [Parameter(ValueFromPipelineByPropertyName = $true)]
+    [string]$OutputPath = "..\..\sdks",
+    [Parameter(ValueFromPipelineByPropertyName = $true)]
+    [string]$TargetSource,
+    [Parameter(ValueFromPipelineByPropertyName = $true)]
+    [switch]$KeepSource,
+    [Parameter(ValueFromPipelineByPropertyName = $true)]
+    [switch]$NonNullable,
```
pgilmorepf: remove

SDKBuildScripts/Update-PlayFabSdk.ps1
```diff
@@ -0,0 +1,293 @@
...
+folder under this path will be generated for each provided SDK platform.  The default path
+points to a sibling directory to the SDKGeneratory repository called "sdks".
+
+.PARAMETER TargetSource
+Override the target source parameter name passed to the generation script.
+
+.PARAMETER KeepSource
+Indicites whether or not to remove all generated source files from the destination location.
+
+.PARAMETER NonNullable
```
pgilmorepf: remove

SDKBuildScripts/Update-PlayFabSdk.ps1
```diff
@@ -0,0 +1,293 @@
...
+        mkdir $sdksPath | Out-Null
+    }
+
+    if($PSCmdlet.ParameterSetName -eq "ApiSpecPfUrlCloudVertical")
+    {
+        if($Vertical -or $Cloud)
+        {
+            # If either cloud or vertical is not provided, we want to remove the leading or trailing dot.
+            $cloudVertical = "$Vertical.$Cloud".TrimEnd(".")
+            $ApiSpecPfUrl = "https://$($cloudVertical).playfabapi.com/apispec"
```
pgilmorepf: is "$($" correct here? That seems wrong.

SDKBuildScripts/Update-PlayFabSdk.ps1
```diff
@@ -0,0 +1,293 @@
...
+        }
+    }
+
+    if($PSCmdlet.ParameterSetName -eq "ApiSpecPath")
+    {
+        $apiSpecSource = "-apiSpecPath"
+
+        if($ApiSpecPath -and $ApiSpecPath -ne "default")
```
pgilmorepf: Is the OOP with the -and and -ne operations going to be right?

SDKBuildScripts/Update-PlayFabSdk.ps1
```diff
@@ -0,0 +1,293 @@
...
+            $apiSpecSource += " $ApiSpecPfUrl"
+        }
+    }
+    else
+    {
+        $apiSpecSource = "-apiSpecGitUrl"
+
+        if($ApiSpecGitUrl -and $ApiSpecGitUrl -ne "default")
+        {
+            $apiSpecSource += " $ApiSpecGitUrl"
```
pgilmorepf: You could check:
if(![Uri]::IsWellFormedUriString($ApiSpecGitUrl, "Absolute"))
here as well.

SDKBuildScripts/Update-PlayFabSdk.ps1
veleek: Yup, it follows the regular order for these parameter types.  I added parens to make it obvious though.

SDKBuildScripts/Update-PlayFabSdk.ps1
veleek: This is correct but not strictly necessary., `$(...)` just says evaluate what's in here and put it in the string.  Sometimes you can have issues if you're doing stuff like `"One_$Value_Two"` which will look for a variable called $Value_Two, so you have to surround it in $(...) to prevent that.

I'm removing it.

veleek: You can see the _ scenario I talked about at the end of this file.

SDKBuildScripts/Update-PlayFabSdk.ps1
```diff
@@ -0,0 +1,289 @@
...
+    [string]$ApiSpecPfUrl,
+    [Parameter(ParameterSetName="ApiSpecPfUrlCloudVertical", ValueFromPipelineByPropertyName = $true)]
+    [string]$Vertical = "master",
+    [Parameter(ParameterSetName="ApiSpecPfUrlCloudVertical", ValueFromPipelineByPropertyName = $true)]
+    [string]$Cloud,
+    [Parameter(ParameterSetName="ApiSpecGitUrl", ValueFromPipelineByPropertyName = $true)]
+    [AllowEmptyString()]
+    [string]$ApiSpecGitUrl,
+    [Parameter(ValueFromPipelineByPropertyName = $true)]
+    [string]$OutputPath = "..\..\sdks",
```
pgilmorepf: Traditionally this relative path has been measured from the "Generate.js" file, rather than this script file.
Line 283 makes me I believe that it should be "../sdks" here too.
Please confirm.

PR #506, opened 2019-05-09
Add Unreal CPP & BP SDK support for Local Debugging Azure Functions

nima-ap:
Support for local debugging has been added in both the C++ and Blueprints SDKs in the Unreal Marketplace Plugin. The feature is summarized as follows:

C++:

* Following the same pattern as the Unity and C# SDKs, we have created a series of utility functions in `PlayFabUtilities.cpp` that allow us to extract the location of the TEMP dir in a user's device (mac and pc), find the local settings file, extract the `LocalApiServer` field, and attempt to send the `ExecuteFunctionRequest` to that address instead of the remote PlayFab address.

BP:

* The logic here is separated out into two parts, and does not follow the same exact approach as the other SDK work we've done for this feature before. This is mainly due to the fact that the process of actually sending over the HTTP request in the BP SDK is decoupled from the API method. As a result, the API method only takes care of assigning a value to a field that represents a full url being present, and the `Activator` method then looks at that field and if it's assigned decides to override the API url with it.

Test results:
Windows:
Windows Unreal Tests
Android:
Android Unreal Tests


nima-ap:
@pgilmorepf I'm not authorized to merge this request, could you please merge?

PR #508, opened 2019-05-13
Remove SecretKey from TitleAuthenticationContext

nima-ap:
We have decided to ask developers to store their title secret key in the app settings of their Azure Functions App and as a result removed this field from `TitleAuthenticationContext`


nima-ap:
As I do not have merge rights, please merge when appropriate :)

PR #510, opened 2019-05-14
Add Task and Player PlayStream based Function Contexts

nima-ap:
The following PR extends the previous `FunctionExecutionContext` (now renamed to `FunctionContext` to account for creating a similar object for both Player PlayStream and Tasks as they are both freshly added invocation mechanisms for Azure Functions which have existed for the classic version of CloudScript. 

As before, these classes are _convenience_ classes which simplify the process of payload de-serialization and Api and Authentication settings assignment for users.

### DIFF
--- a/targets/csharp/source/source/Extensions/Models/PlayFabCloudScriptModelsExtensions.cs.ejs
+++ b/targets/csharp/source/source/Extensions/Models/PlayFabCloudScriptModelsExtensions.cs.ejs
@@ -146,7 +146,7 @@ namespace PlayFab.CloudScriptModels
 
         public PlayFabAuthenticationContext AuthenticationContext { get; set; }
 
-        public ClientModels.NameIdentifier ScheduledTaskNameId { get; set; }
+        public NameIdentifier ScheduledTaskNameId { get; set; }
 
         public Stack<PlayStreamEventHistory> EventHistory { get; set; }
 
@@ -187,7 +187,7 @@ namespace PlayFab.CloudScriptModels
 
         private class FunctionTaskContextInternal
         {
-            public ClientModels.NameIdentifier ScheduledTaskNameId { get; set; }
+            public NameIdentifier ScheduledTaskNameId { get; set; }
 
             public Stack<PlayStreamEventHistory> EventHistory { get; set; }
 
@@ -195,6 +195,13 @@ namespace PlayFab.CloudScriptModels
 
             public TFunctionArgument FunctionArgument { get; set; }
         }
+    }
+
+    public class NameIdentifier
+    {
+        public string Name { get; set; }
+
+        public string Id { get; set; }
     }
 
     public class PlayStreamEventEnvelope
--- a/targets/csharp/source/source/Extensions/Models/PlayFabCloudScriptModelsExtensions.cs.ejs
+++ b/targets/csharp/source/source/Extensions/Models/PlayFabCloudScriptModelsExtensions.cs.ejs
@@ -2,6 +2,7 @@
 namespace PlayFab.CloudScriptModels
 {
     using PlayFab.Json;
+    using System;
     using System.Collections.Generic;
     using System.Net.Http;
     using System.Threading.Tasks;
@@ -40,6 +41,9 @@ namespace PlayFab.CloudScriptModels
                 {
                     TitleId = contextInternal.TitleAuthenticationContext.Id
                 };
+
+                FunctionContextUtil.TrySetSecretKey(settings);
+
                 var authContext = new PlayFabAuthenticationContext
                 {
                     EntityToken = contextInternal.TitleAuthenticationContext.EntityToken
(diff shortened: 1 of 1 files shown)


nima-ap:
retest

nima-ap:
retest

targets/csharp/source/source/Extensions/Models/PlayFabCloudScriptModelsExtensions.cs.ejs
```diff
@@ -1,13 +1,20 @@
+#if NETCOREAPP2_0
```
pgilmorepf: It's possible this isn't the right #if condition.
I'm compiling a .NetCoreApp 2.0, but it's grey-ed out this entire file.

targets/csharp/source/source/Extensions/Models/PlayFabCloudScriptModelsExtensions.cs.ejs
```diff
@@ -17,55 +24,222 @@ namespace PlayFab.CloudScriptModels
...
-        /// <returns>A new populated <c>FunctionExecutionContext</c> instance</returns>
-        public static async Task<FunctionExecutionContext<TFunctionArgument>> Create(System.Net.Http.HttpRequestMessage request)
+        /// <returns>A new populated <c>FunctionContext</c> instance</returns>
+        public static async Task<FunctionContext<TFunctionArgument>> Create(HttpRequestMessage request)
         {
             using (var content = request.Content) 
             {
                 var body = await content.ReadAsStringAsync();
-                var contextInternal = Json.PlayFabSimpleJson.DeserializeObject<FunctionExecutionContextInternal>(body);
+                var contextInternal = Json.PlayFabSimpleJson.DeserializeObject<FunctionContextInternal>(body);
```
pgilmorepf: You are hard coding yourself to a fixed json serializer, and that's not what you should be doing.

                var json = PluginManager.GetPlugin<ISerializerPlugin>(PluginContract.PlayFab_Serializer);
                var contextInternal = json.DeserializeObject<FunctionContextInternal>(body);

targets/csharp/source/source/Extensions/Models/PlayFabCloudScriptModelsExtensions.cs.ejs
```diff
@@ -17,55 +24,222 @@ namespace PlayFab.CloudScriptModels
...
+        /// <param name="request">The request incoming to an Azure Function from the PlayFab server</param>
+        /// <returns>A new populated <c>FunctionPlayerPlayStreamContext</c> instance</returns>
+        public static async Task<FunctionPlayerPlayStreamContext<TFunctionArgument>> Create(HttpRequestMessage request)
         {
-            public string Id { get; set; }
-            public string EntityToken { get; set; }
+            using (var content = request.Content)
+            {
+                var body = await content.ReadAsStringAsync();
+                var contextInternal = PlayFabSimpleJson.DeserializeObject<FunctionPlayerPlayStreamContextInternal>(body);
```
pgilmorepf: You are hard coding yourself to a fixed json serializer, and that's not what you should be doing.

                var json = PluginManager.GetPlugin<ISerializerPlugin>(PluginContract.PlayFab_Serializer);
                var contextInternal = json.DeserializeObject<FunctionPlayerPlayStreamContextInternal>(body);

targets/csharp/source/source/Extensions/Models/PlayFabCloudScriptModelsExtensions.cs.ejs
```diff
@@ -17,55 +24,222 @@ namespace PlayFab.CloudScriptModels
...
+                    PlayerProfile = contextInternal.PlayerProfile,
+                    FunctionArgument = contextInternal.FunctionArgument,
+                    CurrentPlayerId = contextInternal.PlayerProfile.PlayerId
+                };
+            }
         }
+
+        private class FunctionPlayerPlayStreamContextInternal
+        {
+            public TFunctionArgument FunctionArgument { get; set; }
```
pgilmorepf: Regarding your usage of TFunctionArgument
You have utilized reflection and generics at the same time.
You should be aware that this will not work on all platforms.
In particular, this will fail spectacularly on Unity if you try to use the same code there. Some Json serializers will fail here as well, on some platforms.
(Generally iOS)

nima-ap: I'm not sure how I'm utilizing reflection here could you please explain further?
I simply added a generic type to the class and utilized that type in a field. The idea here is the customer must have a type for their `FunctionArguments` field that's in the payload to the AF, and hence can get a context that already deserializes their JSON into that type rather than deserializing it into object and them having to manually do it again. 
Also, as per our previous conversation, the platform issue you're bringing up is the reason why this class is #if'd with NETSTANDARD2_0 as it will only be applicable in Azure Functions runtime which runs on NET Standard 2+ and hence won't be compiled anywhere else. Furthermore, this is all moving to another project, as per our other conversation, and so hence we don't need to resolve this comment by fixing the code in this PR, but I'd still like to get some feedback from you regarding how I'm using reflection here.

PR #513, opened 2019-05-15
Fix missing usage of authenticationContext for InstanceAPIs

nima-ap:
This PR addresses a hot-fix for InstanceAPIs that use EntityToken for authentication. 

I'm not sure if this needs to be addressed in other SDK targets as well.


nima-ap:
retest

nima-ap:
retest

PR #514, opened 2019-05-15
Mimacken/appcenter android

MichaelMackendMS:
Adds AppCenter build and test for the Android and iOS  builds of the Unity test app.


SDKBuildScripts/AppCenterUITestLauncher/AppCenterUITestLauncher/Tests.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        [SetUp]
+        public void BeforeEachTest()
+        {
+            app = AppInitializer.StartApp(platform);
+
+        }
+
+        //[Test]
```
toddbellMSFT: Is there a particular reason to leave this commented out code in?

targets/unity-v2/Testing/Editor/PlayFabPackager.cs
```diff
@@ -16,6 +56,8 @@ public static class PlayFabPackager
             "assets/Testing/scenes/testscene.unity"
         };
 
+    private static readonly string TestPackageName = "com.playfab.service";
```
toddbellMSFT: nitpick: Can we place members at the top of the class? 

I can see the use of keeping variables closer to where they are used, but it helps for general code navigation to organize members at the top of the file for use anywhere else in the file. 

Thoughts?

targets/unity-v2/Testing/Editor/PlayFabPackager.cs
```diff
@@ -3,9 +3,49 @@
...
+using UnityEditor.iOS;
+using UnityEditor.iOS.Xcode;
+using UnityEditor.Build;
+using UnityEditor.Build.Reporting;
+
+using BuildPipeline = UnityEditor.BuildPipeline;
 
 namespace PlayFab.Internal
 {
+    class MyCustomBuildProcessor : IPostprocessBuildWithReport
```
toddbellMSFT: Should we change it to something like PlayFabTestBuildProcessor?

SDKBuildScripts/unity_buildAppCenterTestIOS.sh
```diff
@@ -0,0 +1,117 @@
+#!/bin/bash
+#USAGE: unity_buildAppCenterTestIOS.sh
```
toddbellMSFT: Oo should we standardize this across scripts? This looks like it would be useful

SDKBuildScripts/AppCenterUITestLauncher/AppCenterUITestLauncher/Tests.cs
```diff
@@ -0,0 +1,56 @@
...
+        //public void TestRepl()
+        //{
+        //    app.Repl();
+        //}
```
pgilmorepf: Let's not commit commented code.

pgilmorepf:
retest

targets/unity-v2/Testing/Editor/PlayFabPackager.cs
MichaelMackendMS: Yes.  I'll rename it to something other than the generated name.

targets/unity-v2/Testing/Editor/PlayFabPackager.cs
MichaelMackendMS: Isn't it at the top already?  The only thing ahead of it is another static readonly.

MichaelMackendMS:
retest

PR #515, opened 2019-05-15
Add ifdef on private localApiServer field

nima-ap:
The `_localApiServer` field is only used if the target framework is not net45 and not core20, and hence can itself also be placed under such an `#if` block.


nima-ap:
retest

nima-ap:
retest

nima-ap:
Closing in favor of #520

PR #517, opened 2019-05-17
Creating a new Unity RunAutoTests script in bash

pgilmorepf:
Essentially identical to the previous one.
Added PS4, Switch, and Xbox build test flags, and connected them to appropriate build methods.
Refactored a bit for cleanliness, since Bash is easier to do that with.

I'm not removing the other one yet, since Jenkins relies on the old one. So compare against the old one, and once committed, I'll change the prod Jenkins job to the new script.


pgilmorepf:
retest

PR #518, opened 2019-05-17
Remove Client and Server API from CSharp SDK

nima-ap:
As requested by paul, this clean up of the SDK is part of a rather bigger structural change in SDK Gen coming in with the introduction of "plugins" and the first being the `CloudScriptPlugin` that i'll be PRing soon.

PR #520, opened 2019-05-20
Reorganize CSharp SDK

nima-ap:
The following PR reorganizes the structure of the files in the CSharp SDK generation files for the following purposes:

1. Allow for other _plugin_ projects to be added side by side in the future
1.1. With `PlayFabCloudScriptPlugin` being the first plugin added, which includes the `FunctionContext`, `FunctionPlayerPlayStreamContext` and `FunctionTaskContext` classes that were added previously to the core SDK, but better moved outside of it as they are only useful for those wishing to use the SDK inside Azure Functions.
2. Remove the Client and Server SDK as per Paul's request.
3. Apply a folder structure to generation source code that mimics the output folder structure, to ease the process of finding source and template files corresponding to generated SDK source files.


nima-ap:
Please note that this PR extends off the branch of [this other PR, #518 ](https://github.com/PlayFab/SDKGenerator/pull/518) and so I'm closing 518 in favor of this cause that work (Client and Server SDK removal) is also a part of reorganizing the CSharp SDK.

nima-ap:
retest

targets/csharp/templates/PlayFabUtil.cs.ejs
```diff
@@ -111,7 +111,7 @@ namespace PlayFab
             if (!string.IsNullOrEmpty(envFileContent))
             {
                 var jsonPlugin = PluginManager.GetPlugin<ISerializerPlugin>(PluginContract.PlayFab_Serializer);
-                dynamic envJson = jsonPlugin.DeserializeObject(envFileContent);
+                JsonObject envJson = jsonPlugin.DeserializeObject<JsonObject>(envFileContent);
```
nima-ap: Please note that I'm running out of options here.
1. `JsonObject` is a class in PlayFab.Json, I'm not sure if any serializer can deserialize to it.
2. Creating the object type to deserialize into would require to use reflection if we wish to get the fields by property name rather than by constraining to fixed type. _Making the whole method not string property based isn't a bad idea but is not consistent with what we've done elsewhere, however I'm open to this option_
3. Using `dynamic` requires dependencies like `Microsoft.CSharp` and `System.Dynamic.Runtime` which we do not have on the SDK.

targets/csharp/make.js
```diff
@@ -70,6 +33,9 @@ exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
     const xamarinOutputDir = path.join(rootOutputDir, "XamarinTestRunner");
     templatizeTree(locals, path.resolve(sourceDir, "XamarinTestRunner"), xamarinOutputDir);
     templatizeTree(locals, path.join(apiOutputDir, "source"), path.join(xamarinOutputDir, "XamarinTestRunner", "PlayFabSDK"));
+    5
```
pgilmorepf: Probably a typo...

targets/csharp/make.js
```diff
@@ -100,9 +66,9 @@ function getBaseTypeSyntax(datatype) {
 }
 
 function makeDatatypes(apis, sourceDir, apiOutputDir) {
-    var modelTemplate = getCompiledTemplate(path.resolve(sourceDir, "templates/Model.cs.ejs"));
-    var modelsTemplate = getCompiledTemplate(path.resolve(sourceDir, "templates/Models.cs.ejs"));
-    var enumTemplate = getCompiledTemplate(path.resolve(sourceDir, "templates/Enum.cs.ejs"));
+    var modelTemplate = getCompiledTemplate(path.resolve(sourceDir, "templates", "Model.cs.ejs"));
```
pgilmorepf: While I applaud your sense of correctness, I think that here today, this is "changing a lot of lines for no purpose".
This is a big PR, and I'd rather not have no-op line changes for a PR that is otherwise already complicated.

targets/csharp/templates/PlayFabUtil.cs.ejs
pgilmorepf: 1. It'll only work with one specific serializer. It'll likely outright fail if the customer swaps out the serializer plugin.
3. I agree, dynamic would be bad.
2. Could you not make a strong class definition of the structure of this file? For example, we have a strong and complete definition of what's in TestTitleData.json. Even if you change it, this version of this SDK should know what it's looking for, and deserialize into the static type that it expects. Why does it need to be dynamic?

targets/csharp/source/Plugins/CloudScript/source/PlayFabFunctionContexts.cs
```diff
@@ -0,0 +1,245 @@
...
+        public string ParentTriggerId { get; set; }
+
+        public string ParentEventId { get; set; }
+
+        public bool TriggeredEvents { get; set; }
+    }
+
+    internal class TitleAuthenticationContext
+    {
+        public string Id { get; set; }
```
pgilmorepf: What is this ID? TitleId?

targets/csharp/source/Plugins/CloudScript/source/PlayFabFunctionContexts.cs
```diff
@@ -0,0 +1,245 @@
...
+
+            public TFunctionArgument FunctionArgument { get; set; }
+        }
+    }
+
+    public class NameIdentifier
+    {
+        public string Name { get; set; }
+
+        public string Id { get; set; }
```
pgilmorepf: What is this Id? EntityId?

nima-ap: I'm not sure. It's part of PlayStream types and I created a 1:1 of that type here as it's needed for my deserialization until we get type generation going later.

targets/csharp/source/Plugins/CloudScript/source/PlayFabFunctionContexts.cs
```diff
@@ -0,0 +1,245 @@
...
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// Used for extracting the data fields incoming as payload to Azure Functions into their respective objects. 
+    /// Meant to be used with basic HTTP based Azure Functions.
+    /// Provides  fields through ApiSettings and AuthenticationContext, the profile of the caller, as well as the 
+    /// arguments passed to the execution request.
+    /// </summary>
+    /// <typeparam name="TFunctionArgument">Type of FunctionArgument</typeparam>
+    public class FunctionContext<TFunctionArgument>
```
pgilmorepf: I'd like to make sure my earlier comments about TFunctionArgument weren't lost or hidden.
This solution will work for the CSharpSDK, but not Unity.
And two solutions with totally different implementations will have an increased maintenance cost.

targets/csharp/source/Plugins/CloudScript/source/PlayFabFunctionContexts.cs
nima-ap: Yep. It was general consensus from the MainServer side to not call this TitleId and TitleX TitleY TitleZ because the class itself has Title in its name.

targets/csharp/source/Plugins/CloudScript/source/PlayFabFunctionContexts.cs
nima-ap: So this piece of code is meant to work only and purely from Azure Functions. In other words, thinking about Unity when you write an Azure Function in C# .Net Standard 2.0 is not a concern imho. Things like this and the ability to focus on one target environment when only one environment is applicable to a use case is the reason why we decided to isolate this plugin from the rest of the SDK.

nima-ap: Tl;dr, think of the PlayFabCloudScriptPlugin as a library you install on your Azure Functions App. As long as it works in the context of an Azure Functions App, it's serving its purpose well.

targets/csharp/templates/PlayFabUtil.cs.ejs
nima-ap: Strong type it is.

targets/csharp/source/Plugins/CloudScript/source/PlayFabFunctionContexts.cs
pgilmorepf: Since it's unclear, it would be nice if you'd find out and add a summary comment for it.

targets/csharp/source/Plugins/CloudScript/source/PlayFabFunctionContexts.cs
pgilmorepf: I'm on the fence for this one.
Let's add a summary for it, I think that'll ease my hesitation.

nima-ap: Sure I can try to do that but this is an internal class that's not being exposed to the public interface of the library so I believe adding docs for it is rather unnecessary?

nima-ap: It's simply a structure for typing what goes through the wire from mainserver to azure function

pgilmorepf: Yes and no... I mean...
Our team has to maintain this code, and when things are unclear, we can help out our future selves with a bit of clarity.
I will not remember this conversation in 2 months (or in 2 weeks even...)

PR #521, opened 2019-05-21
Edit QoS SDK for xplat c++

yzhao174:
Edit QoS SDK for c++. Now it returns a sorted list based on latency and send metrics to PlayStreamv2


targets/XPlatCppSdk/source/code/include/playfab/QoS/QoS.h
```diff
@@ -23,5 +23,14 @@ namespace PlayFab
 
         // Number of times a server is pinged
         constexpr int NUM_OF_PING_ITERATIONS = 3;
+
+        enum QoSErrorCode
```
Vicodex: >QoSErrorCode [](start = 13, length = 12)

Can we please add a comment for this enum explaining that these are possible error code values for QoSResult.erroCode?

targets/XPlatCppSdk/source/code/source/playfab/QoS/PlayFabQoSApi.cpp
```diff
@@ -30,6 +94,12 @@ namespace PlayFab
         QoSResult PlayFabQoSApi::GetQoSResult(unsigned int numThreads, unsigned int timeoutMs)
         {
             QoSResult result(move(GetResult(numThreads, timeoutMs)));
+
+            if (result.errorCode != QoSErrorCode::Success)
+            {
+                return result;
```
Vicodex: >return result; [](start = 16, length = 14)

Is it intentional that we don't want to send failures to PlayFab?

yzhao174: Yeah, this is caused by either player not logged in or PlayFab doesn't reply with QosServers list. We shouldn't spam the event stream with those failures.

targets/XPlatCppSdk/source/code/include/playfab/QoS/QoS.h
yzhao174: I will add commnents

pgilmorepf:
Landing this will be delayed until the XPlat automated testing is brought up to date, in line with the other SDKs. Apologies, we're working on it now.

Vicodex:
retest

PR #522, opened 2019-05-21
This checks the TitleId and DeveloperKey settings in case that we don…

JTonoJimenez:
…'t have any of those when we start making the request.

The settings are obtain as soon as we start the program, but if the user change them we have no way on updating unless we restart UE4


pgilmorepf:
retest

pgilmorepf:
I broke the test run, sorry, my bad.

PR #523, opened 2019-05-21
Update PlayFabCommmonUtils.cpp.ejs

toddbellMSFT:
adding in a deprecated unreal call to just 4.20 logic paths.

### DIFF
--- a/targets/UnrealMarketplacePlugin/source/PlayFab/Source/PlayFabCommon/Private/PlayFabCommmonUtils.cpp.ejs
+++ b/targets/UnrealMarketplacePlugin/source/PlayFab/Source/PlayFabCommon/Private/PlayFabCommmonUtils.cpp.ejs
@@ -1,9 +1,27 @@
-﻿<%- copyright %>
+<%- copyright %>
 
 #include "PlayFabCommonUtils.h"
 
 using namespace PlayFabCommon;
 
+// see {your unreal root folder}\Engine\Source\Runtime\Launch\Resources\Version.h for more preprocessor defines
+#if ENGINE_MINOR_VERSION==20
+FString PlayFabCommonUtils::GetTempDir()
+{
+    FString vars[] = { TEXT("TEMPDIR"), TEXT("TEMP"), TEXT("TMP") };
+    TCHAR FileSystemPath[256];
+    for (FString& envVar : vars)
+    {
+       FileSystemPath[0] = 0;
+        FPlatformMisc::GetEnvironmentVariable(*envVar, FileSystemPath, 256);
+        if (FileSystemPath[0])
+        {
+            return FString(256, FileSystemPath);
+        }
+    }
+    return FString();
+}
+#else
 FString PlayFabCommonUtils::GetTempDir()
 {
     FString vars[] = { TEXT("TEMPDIR"), TEXT("TEMP"), TEXT("TMP") };
@@ -17,6 +35,7 @@ FString PlayFabCommonUtils::GetTempDir()
     }
     return FString();
 }
+#endif
 
 FString PlayFabCommonUtils::GetLocalSettingsFileContent()
 {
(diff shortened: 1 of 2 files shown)


targets/UnrealMarketplacePlugin/source/PlayFab/Source/PlayFabCommon/Private/PlayFabCommmonUtils.cpp.ejs
```diff
@@ -1,9 +1,27 @@
-﻿<%- copyright %>
+<%- copyright %>
 
 #include "PlayFabCommonUtils.h"
 
 using namespace PlayFabCommon;
 
+// see {your unreal root folder}\Engine\Source\Runtime\Launch\Resources\Version.h for more preprocessor defines
+#if ENGINE_MINOR_VERSION==20
```
nima-ap: Do higher version qualify for `==20` or should this be `>=20`?

nima-ap: Nvm, I see that your goal was to add logic specifically for 4.20.

toddbellMSFT: Nope, this version of GetEnvironmentVariable with 3 parameters is the only way to access the system variables in 4.20. For later versions (21 and 22) there's the single parameter version which is considered the "correct" version (3 params is labeled as deprecated http://api.unrealengine.com/INT/API/Runtime/Core/GenericPlatform/FGenericPlatformMisc/GetEnvironmentVariable/2/index.html)

pgilmorepf: It's actually a fix for less than 4.20.
The current code works for 4.22, but went uncaught until we tried to verify on 4.20.

nima-ap: Thanks for the fix

targets/UnrealMarketplacePlugin/source/PlayFab/Source/PlayFabCommon/Private/PlayFabCommmonUtils.cpp.ejs
```diff
@@ -1,9 +1,27 @@
...
+FString PlayFabCommonUtils::GetTempDir()
+{
+    FString vars[] = { TEXT("TEMPDIR"), TEXT("TEMP"), TEXT("TMP") };
+    TCHAR FileSystemPath[256];
+    for (FString& envVar : vars)
+    {
+       FileSystemPath[0] = 0;
```
debhaldarMS: Something weird about this formatting

targets/UnrealMarketplacePlugin/source/PlayFab/Source/PlayFabCommon/Private/PlayFabCommmonUtils.cpp.ejs
```diff
@@ -1,9 +1,27 @@
...
 #include "PlayFabCommonUtils.h"
 
 using namespace PlayFabCommon;
 
+// see {your unreal root folder}\Engine\Source\Runtime\Launch\Resources\Version.h for more preprocessor defines
+#if ENGINE_MINOR_VERSION==20
+FString PlayFabCommonUtils::GetTempDir()
+{
+    FString vars[] = { TEXT("TEMPDIR"), TEXT("TEMP"), TEXT("TMP") };
+    TCHAR FileSystemPath[256];
```
debhaldarMS: This is used in 3 places. Please define a local const and use it to futureproff out of sync issues

targets/UnrealMarketplacePlugin/source/PlayFab/Source/PlayFabCommon/Private/PlayFabCommmonUtils.cpp.ejs
debhaldarMS: Add some verbiage around why this is necessary and a code link to the Unreal Docs which shows this is deprecated

toddbellMSFT:
retest

PR #524, opened 2019-05-22
XPlat Timestamps parsed incorrectly

JTonoJimenez:


JTonoJimenez:
Here is the difference in the before/after the proposed changes.

Timestamp before change
Timestamp after change

pgilmorepf:
The screenshots you have provided are demonstrating a difference, but that doesn't count as a unit test.

Can you investigate the existing test that verifies timestamps?:
https://github.com/PlayFab/SDKGenerator/blob/865c83f6b98db85de1a121455a9e56129a73b74b/targets/XPlatCppSdk/source/test/TestApp/PlayFabApiTest.cpp#L220
Is that test correct? If not, why?

I imagine that somehow, we have to update/fix that test above, or we have to determine why that test is insufficient, and add another one.

The diff value you provided demonstrates that your change actually has an affect, but it doesn't demonstrate that the change is actually correct.

JTonoJimenez:
> The diff value you provided demonstrates that your change actually has an affect, but it doesn't demonstrate that the change is actually correct.

The issue is with the TokenExpiration amount of time that is on the call GetEntityToken.
The change is to modify the amount of time from a 32 hours to 24 hours.

Since that piece of data is obtain through a service we might need to add a new test to verify that specific function, or define the amount of time that we want the session to remain.

pgilmorepf:
Admin bypass:
The only job that failed was Unity+Switch, and this change doesn't affect that job.
The failure was related to yesterday's attempt to add App Center credentials to the standard credentials check.

JTonoJimenez:
Link of Reference:
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/mkgmtime-mkgmtime32-mkgmtime64?view=vs-2019

To convert from a local time to UTC time the correct functions are: mktime, _mktime32, and _mktime64
But to converts a UTC time represented by a struct to a UTC time represented by a time_t type: _mkgmtime, _mkgmtime32, _mkgmtime64

PR #530, opened 2019-05-24
XPlat: removed wrong header include in LibUnitTest project that was not needed but required a specific dependency

Vicodex:
- Removed wrong header include which "worked" when building LibUnitTest on a local PC but failed when building on a Jenkins slave which happened not to have a specific dependency lib


Vicodex:
retest

Vicodex:
Retest

PR #534, opened 2019-05-28
nullptr exception in PlayFabEventPipeline

JTonoJimenez:


pgilmorepf:
> Seems weird that we would be dropping modern C++ syntax. Do we not have a modern implementation of CreateEventPacket?

Can you fix your username so that you're using your own account, and not jenkinsbot?
Also,
I think the idea is that it's a shared pointer, not a unique pointer. Therefore it shouldn't be moved, because it's just one more pointer to the same object?

targets/XPlatCppSdk/source/code/source/playfab/PlayFabEventBuffer.cpp
```diff
@@ -98,7 +98,7 @@ namespace PlayFab
         // create an event packet, set it to the tail->next and move the tail
 
         const auto currentEventIndex = eventIndex->fetch_add(1, std::memory_order_relaxed);
-        PlayFabEventPacket* event = CreateEventPacket(reinterpret_cast<uint8_t*>(eventStart), currentEventIndex, std::move(request));
```
Vicodex: > std::move(request)); [](start = 112, length = 21)

Please don't remove this move

targets/XPlatCppSdk/source/code/source/playfab/PlayFabEventBuffer.cpp
```diff
@@ -122,7 +122,7 @@ namespace PlayFab
         }
 
         // event is available; return its values
-        request = std::move(event->eventRequest);
```
Vicodex: >std::move(event->eventRequest); [](start = 18, length = 31)

Please don't remove this move

targets/XPlatCppSdk/source/code/source/playfab/PlayFabEventBuffer.cpp
```diff
@@ -136,7 +136,7 @@ namespace PlayFab
     PlayFabEventPacket* PlayFabEventBuffer::CreateEventPacket(uint8_t *location, const uint64_t index, std::shared_ptr<const IPlayFabEmitEventRequest> request)
     {
         // Use placement new to allocate an event packet in the buffer
-        return new(location)PlayFabEventPacket(index, std::move(request));
```
Vicodex: >std::move(request)) [](start = 54, length = 19)

Please don't remove this move

Vicodex:
Paul, those moves were added for optimization. A shared pointer can be moved when, for example, it needs to be passed as an argument into other function, and not needed in the outer function anymore. It saves some time and memory because a full copy is not created. But we have a bug in buffer.TryPut because we attempted to use the pointer after moving. In that particular case, we should not use move.

---
In reply to: [496755838](https://github.com/PlayFab/SDKGenerator/pull/534#issuecomment-496755838) [](ancestors = 496755838)

Vicodex:
Jorge, please write a test for buffer overflow

JTonoJimenez:
> Jorge, please write a test for buffer overflow

I'm writing the test for buffer overflow

targets/XPlatCppSdk/source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -91,32 +91,35 @@ namespace PlayFabUnit
...
-            eventFailLog += pfEvent->GetName() + " received an error back " +
-                "in the batch #" + std::to_string(pfResponse->batchNumber) + " "
-                "of " + std::to_string(pfResponse->batch->size()) + " events. "
-                "HTTP code: " + std::to_string(pfResponse->playFabError->HttpCode) +
-                ", app error code: " + std::to_string(pfResponse->playFabError->ErrorCode) +
-                ", HTTP status: " + pfResponse->playFabError->HttpStatus +
-                ", Message: " + pfResponse->playFabError->ErrorMessage +
-                "\n";
-        }
+		if (pfResponse->playFabError != nullptr)
```
pgilmorepf: Please change your Visual Studio/Text Editor to use 4 spaces instead of tabs.
And then replace all tab characters in this file with 4 spaces.

targets/XPlatCppSdk/source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -91,32 +91,35 @@ namespace PlayFabUnit
...
+		{
+			// Handle successful event delivery.
+			if (PlayFab::PlayFabErrorCode::PlayFabErrorSuccess == pfResponse->playFabError->ErrorCode)
+			{
+				++eventPassCount;
```
pgilmorepf: If playFabError == null, then I'd expect that situation to be a success...
And this change will completely disable the success block (and everything else) in that situation.
That doesn't seem right.

targets/XPlatCppSdk/source/test/TestApp/PlayFabEventTest.cpp
```diff
@@ -91,32 +91,35 @@ namespace PlayFabUnit
...
+				eventFailLog += pfEvent->GetName() + " received an error back " +
+					"in the batch #" + std::to_string(pfResponse->batchNumber) + " "
+					"of " + std::to_string(pfResponse->batch->size()) + " events. "
+					"HTTP code: " + std::to_string(pfResponse->playFabError->HttpCode) +
+					", app error code: " + std::to_string(pfResponse->playFabError->ErrorCode) +
+					", HTTP status: " + pfResponse->playFabError->HttpStatus +
+					", Message: " + pfResponse->playFabError->ErrorMessage +
+					"\n";
+			}
+		}
```
pgilmorepf: We need to log that something failed here, because of null obj.

Vicodex: You may call testContext->Fail("EmitEventCallback received an error"); when pfResponse->playFabError is null

---
In reply to: [289120073](https://github.com/PlayFab/SDKGenerator/pull/534#discussion_r289120073) [](ancestors = 289120073)

Vicodex:
retest

PR #535, opened 2019-05-29
Add Qos SDK in CSharp

yzhao174:


pgilmorepf:
Linked with:
https://github.com/PlayFab/CSharpSDK/pull/104
(Note to self, land them together)

targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabQosApiTest.cs
```diff
@@ -0,0 +1,71 @@
...
+            };
+
+            try
+            {
+                PlayFabSettings.staticSettings.TitleId = testTitleData.titleId;
+                PlayFabSettings.staticSettings.DeveloperSecretKey = testTitleData.developerSecretKey;
+                PlayFabResult<ServerLoginResult> serverLoginResult = PlayFabServerAPI.LoginWithServerCustomIdAsync(loginWithServerCustomIdRequest, null, testTitleData.extraHeaders).Result;
+                PlayFabSettings.staticPlayer.EntityToken = serverLoginResult.Result.EntityToken.EntityToken;
+                PlayFabSettings.staticPlayer.PlayFabId = serverLoginResult.Result.PlayFabId;
+                PlayFabSettings.staticPlayer.ClientSessionTicket = serverLoginResult.Result.SessionTicket;
```
pgilmorepf: Can you separate these three PlayFabSettings.staticPlayer lines into a separate helper method with a summary that says that it's a bug that these aren't set for the LoginWithServerCustomId method,
and that the method should be deleted when proper SDK support for LoginWithServerCustomId is added?

targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabQosApiTest.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+            catch (AggregateException aggregateException) when (aggregateException.InnerException != null)
+            {
+                throw aggregateException.InnerException;
+            }
+
+            PlayFabQosApi playFabQosApi = new PlayFabQosApi();
+            QosResult qoSResult = playFabQosApi.GetQosResultAsync().Result;
+
+            testContext.IntEquals(0, qoSResult.ErrorCode, $"Qos error: expected ErrorCode 0, actual ErrorCode {qoSResult.ErrorCode}");
```
pgilmorepf: If there is any kind of error string in the qoSResult when it's in a bad state, you should probably display that string too. For maintenance, it'll be easier that trying to figure out what the integer codes mean.

targets/csharp/source/PlayFabSDK/source/Qos/PlayFabQosApi.cs
```diff
@@ -0,0 +1,163 @@
...
+                response = await WriteTelemetryAsync(writeEventsRequest);
+            }
+            catch (Exception exception)
+            {
+                LogQos(exception.ToString());
+            }
+
+            if (response == null || response.Error != null)
+            {
+                LogQos(response?.Error.GenerateErrorReport());
```
pgilmorepf: This line will fail if response is null.

targets/csharp/source/PlayFabSDK/source/Qos/QosRegionResult.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    using MultiplayerModels;
+
+    public class QosRegionResult
+    {
+        public AzureRegion Region { get; set; }
+
+        public int LatencyMs { get; set; }
+
+        public int ErrorCode { get; set; }
```
pgilmorepf: This ErrorCode is always cast from an enum QosErrorCode value, cast to an integer. Why isn't this just a QosErrorCode?

targets/csharp/source/PlayFabSDK/source/Qos/QosResult.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace PlayFab.QoS
+{
+    using System.Collections.Generic;
+
+    public class QosResult
+    {
+        public List<QosRegionResult> RegionResults { get; set; }
+
+        public int ErrorCode { get; set; }
```
pgilmorepf: This ErrorCode is always compared against an enum QosErrorCode value, cast to an integer. Why isn't this just a QosErrorCode?

targets/csharp/source/PlayFabSDK/source/Qos/PlayFabQosApi.cs
```diff
@@ -0,0 +1,163 @@
...
+
+        private async Task<QosResult> GetResultAsync(int timeoutMs)
+        {
+            var result = new QosResult();
+
+            if (!PlayFabClientAPI.IsClientLoggedIn())
+            {
+                LogQos("Client is not logged in");
+                result.ErrorCode = (int)QosErrorCode.NotLoggedIn;
+                return result;
```
pgilmorepf: I feel that you're making it unnecessarily harder for customers to determine what a QosResult actually means.
In this case, the information they need is scattered in multiple places. The int result means nothing unless they know it's actually a QosErrorCode. And the string is sent off to an entirely different place (logAction), almost always totally disconnected from this result.

You could add an ErrorMessage string field to QosResult, which is where "Client is not logged in" would fit perfectly, and it would be sent to the caller of this method, instead of somewhere else entirely. And changing QosResult.ErrorCode to QosErrorCode would reinforce the meaning of those values, and make it very easy for the caller to resolve the error.

targets/csharp/source/PlayFabSDK/source/Qos/PlayFabQosApi.cs
```diff
@@ -0,0 +1,163 @@
...
+            }
+
+            // ping servers
+            result.RegionResults = await GetSortedRegionLatencies(timeoutMs);
+            result.ErrorCode = (int)QosErrorCode.Success;
+            return result;
+        }
+
+        private async Task GetQoSServerList()
```
pgilmorepf: The name GetQoSServerList makes me think that it should return the list once retrieved. Having a null return value is weird for a function named "Get".

targets/csharp/source/PlayFabSDK/source/Qos/PlayFabQosApi.cs
```diff
@@ -0,0 +1,163 @@
...
+                // If the dataCenterMap is already initialized, return
+                return;
+            }
+
+            var request = new ListQosServersRequest();
+            PlayFabResult<ListQosServersResponse> response = await PlayFabMultiplayerAPI.ListQosServersAsync(request);
+            
+            if (response == null || response.Error != null)
+            {
+                LogQos("Could not get the server list from thunderhead\n Error : " + response?.Error.GenerateErrorReport());
```
pgilmorepf: You should do a global replace for:
"response?.Error."
to
"response?.Error?."

targets/csharp/source/PlayFabSDK/source/Qos/PlayFabQosApi.cs
```diff
@@ -0,0 +1,163 @@
+﻿namespace PlayFab.QoS
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
```
pgilmorepf: I strongly discourage people from using Linq in code that the customer will use.
It invokes the garbage collector VERY heavily.
I know it makes "pretty code" and will often produce a result in fewer cpu cycles, but the cost is a huge GC hit later, resulting in an overall lower performance.
In my experience, ANY lazy by-hand-implementation can outperform Linq when you include the GC cost.

targets/csharp/source/PlayFabSDK/source/Qos/PlayFabQosApi.cs
```diff
@@ -0,0 +1,163 @@
...
+    using System.Threading.Tasks;
+    using EventsModels;
+    using Internal;
+    using MultiplayerModels;
+
+    public class PlayFabQosApi
+    {
+        private const int DefaultTimeoutMs = 250;
+        private readonly Dictionary<AzureRegion, string> _dataCenterMap = new Dictionary<AzureRegion, string>();
+        private readonly Action<string> _logAction;
```
pgilmorepf: Given the way that you're using this, I think you'd be better off removing _logAction entirely, and always returning the logged string as a field on the result instead.

targets/csharp/source/PlayFabSDK/source/Qos/QosRegionResult.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace PlayFab.QoS
+{
+    using MultiplayerModels;
+
+    public class QosRegionResult
+    {
+        public AzureRegion Region { get; set; }
```
pgilmorepf: Unless you have an important reason to use properties, please use fields in customer facing code.
Fields are 1000x more performant than properties.

targets/csharp/source/PlayFabSDK/source/Qos/QosResult.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace PlayFab.QoS
+{
+    using System.Collections.Generic;
+
+    public class QosResult
+    {
+        public List<QosRegionResult> RegionResults { get; set; }
```
pgilmorepf: Unless you have an important reason to use properties, please use fields in customer facing code.
Fields are 1000x more performant than properties.

targets/csharp/source/PlayFabSDK/source/Qos/RegionPinger.cs
```diff
@@ -0,0 +1,94 @@
+﻿namespace PlayFab.QoS
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
```
pgilmorepf: See my disclaimer about Linq in the other file above.

targets/csharp/source/PlayFabSDK/source/Qos/PlayFabQosApi.cs
```diff
@@ -0,0 +1,163 @@
...
+
+        private static async Task<PlayFabResult<WriteEventsResponse>> WriteTelemetryAsync(WriteEventsRequest request, object customData = null, Dictionary<string, string> extraHeaders = null)
+        {
+            if ((request?.AuthenticationContext?.EntityToken ?? PlayFabSettings.staticPlayer.EntityToken) == null) throw new PlayFabException(PlayFabExceptionCode.EntityTokenNotSet, "Must call GetEntityToken before calling this method");
+
+            var httpResult = await PlayFabHttp.DoPost("/Event/WriteTelemetryEvents", request, "X-EntityToken", PlayFabSettings.staticPlayer.EntityToken, extraHeaders);
+            if (httpResult is PlayFabError)
+            {
+                var error = (PlayFabError)httpResult;
+                PlayFabSettings.GlobalErrorHandler?.Invoke(error);
```
pgilmorepf: Please remove this. There's a high chance we will remove GlobalErrorHandler later.

targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabQosApiTest.cs
pgilmorepf: Scratch that, I was totally wrong.
You should not be using LoginWithServerCustomId at all. The usage of that method here is not correct, and it took both of us down a rabbit hole of differing assumptions, all of which were incorrect.
Please use PlayFabClientApi.LoginWithCustomId.
That will assign the staticPlayer correctly, and you won't need these 3 lines.

targets/csharp/source/PlayFabSDK/source/Qos/QosRegionResult.cs
yzhao174: Would you point me any reference about this performance comparison. I have read several posts about this, and they all say that auto property performs the same as field. [https://stackoverflow.com/questions/632831/why-are-public-fields-faster-than-properties](https://stackoverflow.com/questions/632831/why-are-public-fields-faster-than-properties) 

---
In reply to: [288687997](https://github.com/PlayFab/SDKGenerator/pull/535#discussion_r288687997) [](ancestors = 288687997)

yzhao174: Change them to properties.

---
In reply to: [289988178](https://github.com/PlayFab/SDKGenerator/pull/535#discussion_r289988178) [](ancestors = 289988178,288687997)

targets/csharp/source/PlayFabSDK/source/Qos/QosRegionResult.cs
yzhao174: Pf-main only expects errorcode and not string message for size concern. I am creating a Facade object here because JsonIgnore is not available here. 

---
In reply to: [288679856](https://github.com/PlayFab/SDKGenerator/pull/535#discussion_r288679856) [](ancestors = 288679856)

targets/csharp/source/PlayFabSDK/source/Qos/QosResult.cs
yzhao174: Creating an error message instead. 

---
In reply to: [288680173](https://github.com/PlayFab/SDKGenerator/pull/535#discussion_r288680173) [](ancestors = 288680173)

targets/csharp/source/PlayFabSDK/source/Qos/PlayFabQosApi.cs
yzhao174: I created an ErrorMessage field. However I am still keeping the ErrorCode as int as pf-main is expecting an integer error code. 

---
In reply to: [288682801](https://github.com/PlayFab/SDKGenerator/pull/535#discussion_r288682801) [](ancestors = 288682801)

targets/csharp/source/PlayFabSDK/source/Qos/PlayFabQosApi.cs
yzhao174: Yes, good suggestion. I have removed it. 

---
In reply to: [288687512](https://github.com/PlayFab/SDKGenerator/pull/535#discussion_r288687512) [](ancestors = 288687512)

targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabQosApiTest.cs
```diff
@@ -0,0 +1,66 @@
+﻿namespace PlayFab.UUnit
```
pgilmorepf: This file needs to be wrapped in:
#if !DISABLE_PLAYFABCLIENT_API && !DISABLE_PLAYFABENTITY_API
#endif

targets/csharp/source/PlayFabSDK/source/Qos/PlayFabQosApi.cs
```diff
@@ -0,0 +1,139 @@
...
+            };
+
+            await WriteTelemetryAsync(writeEventsRequest);
+        }
+
+        private static async Task<PlayFabResult<WriteEventsResponse>> WriteTelemetryAsync(WriteEventsRequest request, object customData = null, Dictionary<string, string> extraHeaders = null)
+        {
+            if ((request?.AuthenticationContext?.EntityToken ?? PlayFabSettings.staticPlayer.EntityToken) == null) throw new PlayFabException(PlayFabExceptionCode.EntityTokenNotSet, "Must call GetEntityToken before calling this method");
+
+            var httpResult = await PlayFabHttp.DoPost("/Event/WriteTelemetryEvents", request, "X-EntityToken", PlayFabSettings.staticPlayer.EntityToken, extraHeaders);
```
pgilmorepf: It's unusual to call obscured APIs directly with PlayFabHttp.
Is this not a public API method?
Note, that by definition, releasing open source client code that calls it, makes it public anyway.

targets/csharp/source/PlayFabSDK/source/Qos/PlayFabQosApi.cs
```diff
@@ -0,0 +1,139 @@
+﻿namespace PlayFab.QoS
```
pgilmorepf: This file needs to be wrapped in:
#if !DISABLE_PLAYFABCLIENT_API && !DISABLE_PLAYFABENTITY_API
#endif

(And as I think about it, every new file you have added here.)

targets/csharp/source/PlayFabSDK/source/Qos/PlayFabQosApi.cs
yzhao174: As we chatted. This is already in Beta. We can change it to call the real API later once it is in prod. 

---
In reply to: [291755067](https://github.com/PlayFab/SDKGenerator/pull/535#discussion_r291755067) [](ancestors = 291755067)

targets/csharp/source/PlayFabSDK/source/Uunit/UUnitIncrementalTestRunner.cs
```diff
@@ -69,6 +69,7 @@ public static TestTitleData VerifyTestTitleData()
 
 #if !DISABLE_PLAYFABCLIENT_API
             PlayFabApiTest.SetTitleInfo(TestTitleData);
+            PlayFabQosApiTest.SetTitleInfo(TestTitleData);
```
pgilmorepf: This **line** has to be additionally #if'd to also !DISABLE_PLAYFABENTITY_API

yzhao174:
retest

Vicodex:
retest

PR #541, opened 2019-06-01
It is now possible to read a genConfig file from the destination repo…

pgilmorepf:
… to determine how an SDK should be read.

genConfig.json will override a large number of inputs if defined, and lack of those inputs will no longer throw early errors, because they may be retrieved very late, from genConfig.json


pgilmorepf:
We need a few more changes before this can be landed safely.

PR #543, opened 2019-06-03
Modify CSharp .csproj file locations and generation process

nima-ap:
The following PR updates the way .csproj files are copied to the output directory, as well as address some additional comments needed on the `PlayFabCloudScriptPlugin` file, and add the AssemblyInfo.cs file in the generation pipeline.


nima-ap:
retest

nima-ap:
@pgilmorepf @toddbellMSFT It seems like the tests are failing due to some instability with regards to the UEM Plugin. Since that's completely irrelevant to the code that was changed here, and since the C# and C# Beta succeeded, can we proceed with merging?

toddbellMSFT:
retest

targets/csharp/source/Plugins/CloudScript/source/PlayFabFunctionContexts.cs
```diff
@@ -1,4 +1,9 @@
-﻿#if NETSTANDARD2_0
+/////////////////////////////////////////////////////////////////////////////////////////////////
+/// THIS PLUGIN IS CURRENTLY IN A BETA STATE MEANT TO ONLY BE USED BY PRIVATE PREVIEW CUSTOMERS.
```
toddbellMSFT: This sounds like this file would be a good candidate to be in a separate beta repo instead of our public repo. The comment is fairly easy to understand, but it seems like anything we put in the public repo should be expected to be scrutinized and potentially used.

nima-ap: The thing with this file is that it's only really useful to create an object for you out of a payload sent to your Azure Function. Non private-preview customers do not have anything to do with this file for now cause they won't have the required input coming in to this function, even if they try to use it.

nima-ap:
Adding Gudge on this so that he can take over if I don't get to make the required changes before I take off on vacation

targets/csharp/source/Plugins/CloudScript/source/PlayFabFunctionContexts.cs
pgilmorepf: As long as you're fine with the fact that some customers WILL find the file, and the comments, I don't strongly oppose its existence.
We will have the ability to keep features in private repos soon, so we can revisit this after that.

pgilmorepf:
retest

Vicodex:
Please make sure that Nuget package generation still works after this

pgilmorepf:
retest

pgilmorepf:
retest

pgilmorepf:
retest

pgilmorepf:
retest

PR #544, opened 2019-06-04
Pg/gen config

pgilmorepf:


pgilmorepf:
I am relaunching this code review, this time as a refactor, with the genConfig.json line disabled for now.
A lot of this refactoring is similar to what I want to do for the target & target2 feature (private inputs), but I don't want to have to resolve it later.

generate.js
```diff
@@ -647,14 +678,17 @@ function getCompiledTemplate(templatePath) {
...
+    }
+    catch (error) {
+        console.error(error);
+        setTimeout(doNothing, 30000);
+        // throw(error);
+    }
 }
+// Kick everything off
+catchAndReport(parseAndLoadApis);
 setTimeout(doNothing, 5000);
```
Vicodex: Btw i don't think we need these timeouts anymore

generate.js
```diff
@@ -647,14 +678,17 @@ function getCompiledTemplate(templatePath) {
...
-    setTimeout(doNothing, 30000);
-    throw error;
+function catchAndReport(method) {
+    try {
+        method();
+    }
+    catch (error) {
+        console.error(error);
+        setTimeout(doNothing, 30000);
+        // throw(error);
```
Vicodex: >// throw(error); [](start = 8, length = 16)

Is this comment needed?

generate.ts
```diff
@@ -187,24 +189,36 @@ function extractArgs(args, argsByName, targetOutputPathList, errorMessages) {
...
-interface ITargetOutput {
-    name: string,
-    dest: string,
+interface IBuildTarget {
+    buildFlags: string[],
+    destPath: string,
+    srcFolder: string,
+    versionKey: string,
+    versionString: string,
 }
```
Vicodex: Let's move interface to the top of the file for easier find?

generate.ts
```diff
@@ -42,8 +41,13 @@ var sdkGeneratorGlobals: SdkGenGlobals = {
     // Internal note: We lowercase the argsByName-keys, targetNames, buildIdentifier, and the flags.  Case is maintained for all other argsByName-values, and targets
     argsByName: {}, // Command line args compiled into KVP's
     errorMessages: [], // String list of errors during parsing and loading steps
-    targetOutputPathList: [], // A list of objects that describe sdk targets to build
-    buildFlags: [], // The sdkBuildFlags which modify the list of APIs available in this build of the SDK
+    buildTarget: {
```
Vicodex: >buildTarget [](start = 4, length = 11)

This can be simply of a type IBuildTarget, right? (instead of repeating the same field declarations that you already declared in that interface)

generate.ts
```diff
@@ -42,8 +41,13 @@ var sdkGeneratorGlobals: SdkGenGlobals = {
...
+        buildFlags: [], // The flags applied to this build
+        destPath: null, // The path to the destination (usually a git repo)
+        srcFolder: null, // The SdkGenerator/targets/subfolder
+        versionKey: null, // The key in the API_Specs/SdkManualNotes.json file that has the version for this SDK
+        versionString: null, // The actual version string, from SdkManualNotes, or from another appropriate input
+    }, // Describes where and how to build the target
```
Vicodex: Put field comments in the interface please

generate.ts
```diff
@@ -56,10 +60,10 @@ let specializationContent;
 
 /////////////////////////////////// The main build sequence for this program ///////////////////////////////////
 function parseAndLoadApis() {
-    console.log("My args:" + process.argv.join(" "));
+    // console.log("My args:" + process.argv.join(" "));
```
Vicodex: >// console.log("My args:" + process.argv.join(" ")); [](start = 4, length = 52)

Why removing this?

generate.js
pgilmorepf: I'm willing to discuss them separately, but I like/need them, and I don't have a good workaround.
(VS and NodeJS 10 don't get along properly)

generate.js
```diff
@@ -341,40 +350,62 @@ function downloadFromUrl(srcUrl, appendUrl, apiCache, cacheKey, onEachComplete,
...
-    console.log("Generating PlayFab APIs from specs: " + apiSrcDescription);
+function generateApis(buildIdentifier, target) {
+    console.log("Generating PlayFab APIs from specs: " + sdkGeneratorGlobals.apiSrcDescription);
+    var genConfig = null;
+    var genConfigPath = path.resolve(target.destPath, "genConfig.json");
+    try {
+        genConfig = require(genConfigPath);
+        console.log("Loaded genConfig at: " + genConfigPath);
+    }
+    catch (_) {
```
toddbellMSFT: >_ [](start = 11, length = 1)

Clarification: is this naming the err variable as _ ? or is this a special JavaScript feature?

generate.ts
```diff
@@ -42,8 +41,13 @@ var sdkGeneratorGlobals: SdkGenGlobals = {
...
     argsByName: {}, // Command line args compiled into KVP's
     errorMessages: [], // String list of errors during parsing and loading steps
-    targetOutputPathList: [], // A list of objects that describe sdk targets to build
-    buildFlags: [], // The sdkBuildFlags which modify the list of APIs available in this build of the SDK
+    buildTarget: {
+        buildFlags: [], // The flags applied to this build
+        destPath: null, // The path to the destination (usually a git repo)
+        srcFolder: null, // The SdkGenerator/targets/subfolder
+        versionKey: null, // The key in the API_Specs/SdkManualNotes.json file that has the version for this SDK
+        versionString: null, // The actual version string, from SdkManualNotes, or from another appropriate input
```
Vicodex: >versionString [](start = 8, length = 13)

Do we need both a version string and a version key? seems like redundancy. Can we keep just the version value?

generate.js
```diff
@@ -341,40 +350,62 @@ function downloadFromUrl(srcUrl, appendUrl, apiCache, cacheKey, onEachComplete,
...
+        console.log("Did not find: " + genConfigPath);
```
toddbellMSFT: >console.log("Did not find: " + genConfigPath); [](start = 8, length = 46)

Should we allow the generation to continue to the code below if something went wrong here?

generate.ts
pgilmorepf: No, because I'm using interfaces, not objects.
There's a long answer, but the short answer is JS and TS are weird.

generate.ts
```diff
@@ -81,7 +85,7 @@ function reportErrorsAndExit(errorMessages) {
 
     // Else, report all errors and exit the program
     console.log("Syntax: node generate.js\n" +
-        "\t\t<targetName>=<targetOutputPath>\n" +
+        "\t\t(<targetName>=<targetOutputPath>|-destPath <destFolderPath>)\n" +
```
Vicodex: >-destPath <destFolderPath> [](start = 46, length = 26)

When using -destPath argument, can we consider using an additional parameter -getConfigPath=<path to getConfig.json file>? Sure, in most cases it will point to a file located at the root of destPath, but.. not necessarily, for advanced scenarios? It would be cool to keep door open for additional flexibility

generate.ts
```diff
@@ -135,10 +139,10 @@ function parseCommandInputs(args, argsByName, errorMessages, targetOutputPathLis
         sdkGeneratorGlobals.unitySubfolder = argsByName.unityDestinationSubfolder;
 
     // Output an error if no targets are defined
-    if (targetOutputPathList.length === 0)
-        errorMessages.push("No targets defined, you must define at least one.");
+    if (!buildTarget.destPath)
+        errorMessages.push("Build target not defined, nothing to build.");
```
Vicodex: >Build target [](start = 28, length = 12)

"Build target's destPath not defined", it's probably better to be more precise with error messages

generate.ts
```diff
@@ -187,24 +189,36 @@ function extractArgs(args, argsByName, targetOutputPathList, errorMessages) {
     }
 
     // Pull from environment variables if there's no console-defined targets
-    if (targetOutputPathList.length === 0 && process.env.hasOwnProperty("SdkSource") && process.env.hasOwnProperty("SdkName")) {
-        checkTarget(process.env.hasOwnProperty("SdkSource"), process.env.hasOwnProperty("SdkName"), targetOutputPathList, errorMessages);
+    if (!buildTarget.destPath) {
+
+        console.log("argsByName: " + JSON.stringify(argsByName) + " " + argsByName["destpath"]);
```
Vicodex: >  [](start = 67, length = 1)

Please insert text ", destPath: " here instead of just a space

generate.ts
```diff
@@ -187,24 +189,36 @@ function extractArgs(args, argsByName, targetOutputPathList, errorMessages) {
...
+
+        if (argsByName["destpath"]) {
+            checkTarget(argsByName["srcfolder"], argsByName["destpath"], buildTarget, errorMessages);
+        } else if (process.env.hasOwnProperty("SdkName")) {
+            checkTarget(process.env["SdkSource"], process.env["SdkName"], buildTarget, errorMessages);
+        }
```
Vicodex: > } [](start = 7, length = 2)

else { log error? }

generate.js
pgilmorepf: In JS, I believe it allocates a variable named "_" and assigns it to the value of the caught exception.
By convention in some scripting languages, a variable "_" is just an extremely temporary trash variable. Multiple scripting languages use it as convention, and one uses it as explicit "ignore this var" and it optimizes out that assignment. (But I can't remember which is which)

generate.ts
```diff
@@ -187,24 +189,36 @@ function extractArgs(args, argsByName, targetOutputPathList, errorMessages) {
...
-function checkTarget(sdkSource, sdkDestination, targetOutputPathList, errorMessages) {
-    var targetOutput: ITargetOutput = {
-        name: sdkSource,
-        dest: path.normalize(sdkDestination)
-    };
-    if (fs.existsSync(targetOutput.dest) && !fs.lstatSync(targetOutput.dest).isDirectory())
-        errorMessages.push("Invalid target output path: " + targetOutput.dest);
-    else
-        targetOutputPathList.push(targetOutput);
+function checkTarget(sdkSrcFolder, sdkDestination, buildTarget: IBuildTarget, errorMessages) {
```
Vicodex: >checkTarget [](start = 9, length = 11)

This sounds more like "setBuildTarget" rather than "checkTarget". Well, we are setting it there, not just checking something.

generate.ts
pgilmorepf: I did this on purpose.
Customers will not be able to put their version numbers into our SdkManualNodes.json file, and for now, that file is extremely hard coded.
Customers will have to specifiy versionString in their genConfig.json.
However, I for now, I didn't want to break our SdkManualNodes.json flow.
I'm just thinking about more general usage and long term planning.

generate.js
pgilmorepf: This file is explicitly designed to be optional (for now). Basically, we want a certain set of inputs, either from command line, or environment variables, or from genConfig.json.
If we get the right inputs, we don't care where they came from.

generate.ts
```diff
@@ -406,45 +420,77 @@ function downloadFromUrl(srcUrl: string, appendUrl: string, apiCache, cacheKey:
...
-    console.log("Generating PlayFab APIs from specs: " + apiSrcDescription);
+interface IGenConfig {
+    branchSpecMap: { [key: string]: string; },
+    delSrc: boolean,
+    buildFlags: string,
+    outputDirs: string[],
+    srcFolder: string,
+    versionKey: string,
+    versionString: string,
+}
```
Vicodex: Same here, let's move all interfaces up top please?

generate.ts
pgilmorepf: That's an interesting idea. I think we could add it later if we so decided.
I don't think it needs to come in with this pass of this feature.

generate.ts
```diff
@@ -406,45 +420,77 @@ function downloadFromUrl(srcUrl: string, appendUrl: string, apiCache, cacheKey:
 }
 
 /////////////////////////////////// Major step 3 - Generate the indicated ouptut files ///////////////////////////////////
-function generateApis(buildIdentifier, targetOutputPathList, buildFlags, apiSrcDescription) {
-    console.log("Generating PlayFab APIs from specs: " + apiSrcDescription);
+interface IGenConfig {
```
Vicodex: >IGenConfig [](start = 10, length = 10)

Please add a comment what this interface defines, what it is for

generate.ts
```diff
@@ -406,45 +420,77 @@ function downloadFromUrl(srcUrl: string, appendUrl: string, apiCache, cacheKey:
...
+
+    if (genConfig) {
+        if (genConfig.buildFlags) target.buildFlags = genConfig.buildFlags.split(" ");
+        if (genConfig.srcFolder) target.srcFolder = genConfig.srcFolder;
+        if (genConfig.versionKey) target.versionKey = genConfig.versionKey;
+        if (genConfig.versionString) target.versionString = genConfig.versionString;
+    }
+
+    if (!target.srcFolder) {
+        throw Error("SdkGenerator/target subfolder not defined: " + target.srcFolder);
```
Vicodex: >Error [](start = 14, length = 5)

Will this still display error as we expect it, in Jenkins logs?

Vicodex: >SdkGenerator/target subfolder not defined [](start = 21, length = 41)

"Build target's srcFolder is not defined"

generate.ts
Vicodex: I meant, you could type it as:

buildTarget: IBuildTarget {
 buildFlags: [] ...
 . . .
}
for additional type safety? no?

---
In reply to: [290496574](https://github.com/PlayFab/SDKGenerator/pull/544#discussion_r290496574) [](ancestors = 290496574)

generate.ts
```diff
@@ -16,17 +29,22 @@ interface SdkDoc {
...
+    errorMessages: string[]; // String list of errors during parsing and loading steps
+    buildTarget: IBuildTarget; // Describes where and how to build the target
+    apiSrcDescription: string; // Assigned if/when the api-spec source is fetched properly
+    apiCache: { [key: string]: any; } // We have to pre-cache the api-spec files, because latter steps (like ejs) can't run asynchronously
+    sdkDocsByMethodName: { [key: string]: SdkDoc; } // When loading TOC, match documents to the SdkGen function that should be called for those docs
     specialization: string;
     unitySubfolder: string;
 }
 
+interface SpecializationTocRef {
```
MichaelMackendMS: Shouldn't all interfaces adhere to the "I<interface name>" scheme?

generate.ts
pgilmorepf: No, the type safety is already strictly enforced, and can't be violated. Let's talk in person if you need.

generate.ts
pgilmorepf: Changing this from
```throw "string"```
to
```throw Error("string")```
Will significantly improve the stacktrace info that is displayed, and will propagate the failure to jenkins, as expected.

pgilmorepf:
retest

pgilmorepf:
The only failure was UE Beta, and it failed in the UE Compile/Build step, after the SDK Generation.
Thus, the failure seen in the pipeline was unrelated to this change, and all the other tests related to this change passed.
I'm gonna admin-bypass this change, pending a separate change that's already in PR/CR to fix the last failure.

PR #546, opened 2019-06-05
SdkGenerator now searches both targets and privateTemplates sub-directories for template files

pgilmorepf:
Reducing the meaning of "target" to just the container that does the build action.
The previous name clarification of "source" was changed to templates.
And the templates rename was aggressively re-applied to all internal variables (from source and target)
It should now be consistently: template to destination

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,8 @@
 **/beta_*.sh
 **/.DS_Store*
 
+/privateTemplates/
+
 .project
 .classpath
 *.suo
--- a/generate.ts
+++ b/generate.ts
@@ -107,15 +107,21 @@ function reportErrorsAndExit(errorMessages) {
         return; // No errors to report, so continue
 
     // Else, report all errors and exit the program
-    console.log("Syntax: node generate.js\n" +
-        "\t\t(<templateName>=<outputPath>|-destPath <destFolderPath>)\n" +
-        "\t\t-(apiSpecPath|apiSpecGitUrl|apiSpecPfUrl)[ (<apiSpecPath>|<apiSpecGitUrl>|<apiSpecPfUrl>)]\n" +
-        "\t\t[ -flags <flag>[ <flag> ...]]\n\n" +
-        "\tExample: node generate.js unity-v2=../sdks/UnitySDK -apiSpecPath ../API_Specs -flags xbox playstation\n" +
-        "\t\tThis builds the UnitySDK, from Specs at relative path ../API_Specs, with console APIs included\n" +
-        "\t<apiSpecPath> : Directory or url containing the *.api.json files\n" +
-        "\tYou must list exactly one <templateName>=<outputPath> arguments.\n" +
-        "\tWarning, there can be no spaces in the template to output specification\n");
+    console.log("Syntax: node generate.js\n\n" +
+        "\tCLI Options:\n" +
+        "\t(<templateName>=<outputPath>|-destPath <outputPath>)\n" +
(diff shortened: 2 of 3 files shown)


generate.ts
```diff
@@ -223,34 +225,61 @@ function extractArgs(args, argsByName: { [key: string]: string; }, buildTarget:
...
-    buildTarget.srcFolder = sdkSrcFolder;
-    buildTarget.versionKey = sdkSrcFolder;
+    buildTarget.templateFolder = sdktemplateFolder;
+    buildTarget.versionKey = sdktemplateFolder;
     buildTarget.versionString = null;
 }
 
-function getTargetsList() {
-    var targetList = [];
+function getMakeScriptForTemplate(buildTarget: IBuildTarget) {
```
pgilmorepf: Note to Reviewers:
The bulk of the change starts here.

generate.ts
```diff
@@ -223,34 +225,61 @@ function extractArgs(args, argsByName: { [key: string]: string; }, buildTarget:
...
+            var eachTemplate = templatesInRoot[i];
+            if (eachTemplate[0] === ".")
+                continue;
+
+            var eachTemplateDir = path.resolve(templateRootDir, eachTemplate);
+            var targetMain = path.resolve(eachTemplateDir, "make.js"); // search for make.js in each subdirectory within "targets"
+            if (fs.existsSync(targetMain))
+                targetList.push(eachTemplate);
+        }
     }
```
pgilmorepf: Note to Reviewers:
The bulk of the change wraps up here.

generate.ts
```diff
@@ -451,20 +480,13 @@ function generateApis(buildIdentifier, target: IBuildTarget) {
...
-    if (!target.srcFolder) {
-        throw Error("SDKGenerator/target/<srcFolder> not defined, for srcFolder: " + target.srcFolder);
-    }
-
-    var targetsDir = path.resolve(__dirname, "targets");
-    var targetSourceDir = path.resolve(targetsDir, target.srcFolder);
-    var targetMain = path.resolve(targetSourceDir, "make.js");
-    console.log("Making target from: " + targetMain + "\n - to: " + target.destPath);
-    var targetMaker = require(targetMain);
+    getMakeScriptForTemplate(target);
```
pgilmorepf: This calls the code I alerted to above.

pgilmorepf:
retest

.gitignore
```diff
@@ -10,6 +10,8 @@
 **/beta_*.sh
 **/.DS_Store*
 
+/privateTemplates/
```
Vicodex: >privateTemplates [](start = 1, length = 16)

How about just "templates". I actually hope to see that in the future any SDK will be generated this same way

generate.ts
```diff
@@ -106,22 +108,22 @@ function reportErrorsAndExit(errorMessages) {
 
     // Else, report all errors and exit the program
     console.log("Syntax: node generate.js\n" +
-        "\t\t(<targetName>=<targetOutputPath>|-destPath <destFolderPath>)\n" +
+        "\t\t(<templateName>=<outputPath>|-destPath <destFolderPath>)\n" +
```
Vicodex: >outputPath [](start = 30, length = 10)

Is OutputPath functionally different from destFolderPath? Are they essentially the same?

Vicodex: On that note, what do you think if we have this:

{ -templateName=<old-target-name> | -templatePath=<folder-containing-templates-for-some-private-repo>}

---
In reply to: [291384880](https://github.com/PlayFab/SDKGenerator/pull/546#discussion_r291384880) [](ancestors = 291384880)

.gitignore
pgilmorepf: We could add/change/rename/remove folders quite easily now.

pgilmorepf: I'm going to keep it named after our current intent for now.

PR #550, opened 2019-06-07
Bug fix: GenerateErrorReport can provide incomplete and misleading information…

JTonoJimenez:
… (C#)

### DIFF
--- a/targets/csharp/templates/Errors.cs.ejs
+++ b/targets/csharp/templates/Errors.cs.ejs
@@ -27,11 +27,23 @@ namespace PlayFab
         public Dictionary<string, string[]> ErrorDetails;
 
         private static readonly StringBuilder Sb = new StringBuilder();
+        /// <summary>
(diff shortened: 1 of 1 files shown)


targets/csharp/templates/Errors.cs.ejs
```diff
@@ -47,7 +47,10 @@ namespace PlayFab
     public class PlayFabBaseResult
     {
         public object CustomData;
-        public PlayFabError Error;
```
MichaelMackendMS: Why was it problematic to have the PlayFabError type directly in the base result? What does this mean for the future of the PlayFabError type?

Duplicating the contents of PlayFabError into here seems to introduce some ambiguity as to the purpose of each as well as increases code maintenance.

JTonoJimenez:
Their are cases where we can't extract data from the error, and we need to return whatever information might be useful to the client.

pgilmorepf:
retest

PR #552, opened 2019-06-10
spawn uploader instead of instantiating so as to give it world access

MichaelMackendMS:
The uploader was not spawned in the scene so it didn't have access to the world instance, hence it would never terminate.  This replaces the instantiation with spawning, fixing that issue.   A consequence of this is that the test actor is no longer the owner of the uploader- the world is. Hence, the uploader now calls destroy on itself when it's work is complete and lets the garbage collector do the rest .


MichaelMackendMS:
retest

pgilmorepf:
The failures are all XPlat, and this is an Unreal change.
Therefore it's safe to admin-land this.
Hopefully V will get the test filtering done this week.

Vicodex:
retest

Vicodex:
retest

pgilmorepf:
retest

PR #553, opened 2019-06-10
Bug fix: Curl Http plugin is disabling CURLOPT_SSL_VERIFYPEER

JTonoJimenez:
Tested and it works

### DIFF
--- a/targets/XPlatCppSdk/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
+++ b/targets/XPlatCppSdk/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
@@ -156,7 +156,7 @@ namespace PlayFab
         curl_easy_setopt(curlHandle, CURLOPT_WRITEFUNCTION, CurlReceiveData);
 
         // Send
-        curl_easy_setopt(curlHandle, CURLOPT_SSL_VERIFYPEER, true); // TODO: Replace this with a ca-bundle ref???
+        curl_easy_setopt(curlHandle, CURLOPT_SSL_VERIFYPEER, true);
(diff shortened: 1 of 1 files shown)


targets/XPlatCppSdk/source/code/source/playfab/PlayFabCurlHttpPlugin.cpp
```diff
@@ -156,7 +156,7 @@ namespace PlayFab
         curl_easy_setopt(curlHandle, CURLOPT_WRITEFUNCTION, CurlReceiveData);
 
         // Send
-        curl_easy_setopt(curlHandle, CURLOPT_SSL_VERIFYPEER, false); // TODO: Replace this with a ca-bundle ref???
+        curl_easy_setopt(curlHandle, CURLOPT_SSL_VERIFYPEER, true); // TODO: Replace this with a ca-bundle ref???
```
Vicodex: >// TODO: Replace this with a ca-bundle ref??? [](start = 68, length = 45)

I think we may remove this comment. It's really meant to work on Linux distributions that already have necessary public server cert storages set up

pgilmorepf: I don't want to land this until we have XPlat testing working on all platforms.
I agree it's the right change, but it was necessary at one point, so we need to verify first that it works on all platforms.

Vicodex: Paul, this particular file is used only and only when building XPlat on Linux using makefile. It is not used on any other platform

pgilmorepf:
Ok, sounds good.

PR #554, opened 2019-06-10
Bug fix: PubSub tests "overrun" their timers (Unity)

JTonoJimenez:


targets/unity-v2/Testing/Tests/Shared/Uunit/UUnitTestContext.cs
```diff
@@ -44,6 +44,10 @@ public UUnitTestContext(UUnitTestCase testInstance, Action<UUnitTestContext> tes
 
         public void EndTest(UUnitFinishState finishState, string resultMsg)
         {
+            if (finishState != UUnitFinishState.PENDING)
+            {
+                return;
```
pgilmorepf: This is not sufficient.
This just hides the problem.

There are state transitions that are normal/acceptable:
PENDING->PASSED
PENDING->FAILED
PASSED->FAILED

There are state transitions that should be explicitly disallowed, and attempting them should result in failure:
FAILED->PASSED (Keep original fail message, append "You can't mark a test as finished twice")
PASSED->PASSED (Fail with a new message that says "You can't mark a test as finished twice")
FAILED->FAILED (Keep/merge both messages)

MichaelMackendMS:
Can we get a bit of description as to how these changes addresses the problem?  Like: 
1) What caused the problem
2) How this addresses it

It would save a lot of time in trying to reverse engineer what the reasoning was and why we should feel like this is good to go. 

Thanks!

Vicodex:
retest

Vicodex:
retest

targets/unity-v2/Testing/Tests/Shared/Uunit/UUnitTestContext.cs
```diff
@@ -44,6 +44,29 @@ public UUnitTestContext(UUnitTestCase testInstance, Action<UUnitTestContext> tes
...
+                else
+                {
+                    if (finishState != UUnitFinishState.FAILED)
+                    {
+                        resultMsg += "You can't mark a test as finished twice.\n";
+                    }
+                    TestResultMsg = resultMsg;
+                    return;
+                }
+            }
```
pgilmorepf: It's really hard to follow this logic.

I would prefer:

if (condition && condition)
// logic
// logic
// logic
else if (condition
// logic
else
// logic

pgilmorepf:
retest

pgilmorepf:
retest

PR #555, opened 2019-06-11
Bug fix: PubSub tests "overrun" their timers (XPlattCpp)

JTonoJimenez:


targets/XPlatCppSdk/source/test/TestApp/TestContext.cpp
```diff
@@ -8,6 +8,29 @@ namespace PlayFabUnit
 {
     void TestContext::EndTest(TestFinishState state, std::string resultMsg)
     {
+        if (finishState != TestFinishState::PENDING)
+        {
+            resultMsg += testResultMsg;
```
JTonoJimenez: Saves the previous Message saved, just to know if we pass or fail before.

targets/XPlatCppSdk/source/test/TestApp/TestContext.cpp
```diff
@@ -8,6 +8,29 @@ namespace PlayFabUnit
 {
     void TestContext::EndTest(TestFinishState state, std::string resultMsg)
     {
+        if (finishState != TestFinishState::PENDING)
```
JTonoJimenez: If the status is not Pending that means that we already finish the test at least once, and we are running it again  for a different reason.

targets/XPlatCppSdk/source/test/TestApp/TestContext.cpp
```diff
@@ -8,6 +8,29 @@ namespace PlayFabUnit
...
+        {
+            resultMsg += testResultMsg;
+            resultMsg += "You can't mark a test as finished twice.\n";
+            if (finishState == TestFinishState::PASSED)
```
JTonoJimenez: This are the cases where we want to handle of sending a new message or adding times.

targets/XPlatCppSdk/source/test/TestApp/TestContext.cpp
```diff
@@ -8,6 +8,29 @@ namespace PlayFabUnit
...
+            else
+            {
+                if (state != TestFinishState::FAILED)
+                {
+                    resultMsg += "You can't mark a test as finished twice.\n";
+                }
+                testResultMsg += resultMsg;
+                return;
+            }
+        }
```
pgilmorepf: It's really hard to follow this logic.

I would prefer:

if (condition && condition)
// logic
// logic
// logic
else if (condition
// logic
else
// logic

Vicodex:
retest

Vicodex:
retest

Vicodex:
Jorge, this PR fails on Xbox (test application times out). Without these changes, Xbox succeeds. It means something in this PR is not working as intended.

pgilmorepf:
retest

pgilmorepf:
retest

PR #556, opened 2019-06-11
Bug fix: PubSub tests "overrun" their timers (C#)

JTonoJimenez:


targets/csharp/source/PlayFabSDK/source/Uunit/UUnitTestContext.cs
```diff
@@ -44,6 +44,29 @@ public UUnitTestContext(UUnitTestCase testInstance, Action<UUnitTestContext> tes
 
         public void EndTest(UUnitFinishState finishState, string resultMsg)
         {
+            if (FinishState != UUnitFinishState.PENDING)
```
JTonoJimenez: Same changes as in Unity.
https://github.com/PlayFab/SDKGenerator/pull/554

targets/csharp/source/PlayFabSDK/source/Uunit/UUnitTestContext.cs
```diff
@@ -44,6 +44,29 @@ public UUnitTestContext(UUnitTestCase testInstance, Action<UUnitTestContext> tes
...
+                else
+                {
+                    if (finishState != UUnitFinishState.FAILED)
+                    {
+                        resultMsg += "You can't mark a test as finished twice.\n";
+                    }
+                    TestResultMsg = resultMsg;
+                    return;
+                }
+            }
```
pgilmorepf: It's really hard to follow this logic.

I would prefer:

if (condition && condition)
    // logic
    // logic
    // logic
else if (condition
    // logic
else
    // logic

Vicodex:
retest

targets/csharp/source/PlayFabSDK/source/Uunit/UUnitTestContext.cs
```diff
@@ -44,10 +44,97 @@ public UUnitTestContext(UUnitTestCase testInstance, Action<UUnitTestContext> tes
...
+            else if (FinishState == UUnitFinishState.FAILED)
+            {
+                switch (finishState)
+                {
+                    case UUnitFinishState.PASSED:
+                        TestResultMsg += "Test try to Pass after Failing.\n";
+                        break;
+                    case UUnitFinishState.FAILED:
+                        TestResultMsg += "Test try to Fail twice.\n";
```
pgilmorepf: As I discussed earlier, we have declared that this test has failed twice. So we need to merge the old fail message and the new one, because the test-debugger will need to fix both issues.

targets/csharp/source/PlayFabSDK/source/Uunit/UUnitTestContext.cs
```diff
@@ -44,10 +44,97 @@ public UUnitTestContext(UUnitTestCase testInstance, Action<UUnitTestContext> tes
...
+            if (FinishState == UUnitFinishState.PENDING)
+            {
+                EndTime = DateTime.UtcNow;
+                TestResultMsg = resultMsg;
+                FinishState = finishState;
+                ActiveState = UUnitActiveState.READY;
+            }
+            else if (FinishState == UUnitFinishState.PASSED)
+            {
+                switch (finishState)
```
pgilmorepf:
```suggestion
                switch (finishState)
                FinishState = UUnitFinishState.FAILED; // Any attempt to end the test multiple times causes the test to be counted as failed.
```

Vicodex:
retest

pgilmorepf:
retest

targets/csharp/source/PlayFabSDK/source/Uunit/UUnitTestContext.cs
```diff
@@ -44,10 +44,97 @@ public UUnitTestContext(UUnitTestCase testInstance, Action<UUnitTestContext> tes
...
+                        TestResultMsg += "Test try to Fail twice.\n";
+                        break;
+                    case UUnitFinishState.SKIPPED:
+                        TestResultMsg += "Test try to be Skipped after Failing.\n";
+                        break;
+                    case UUnitFinishState.TIMEDOUT:
+                        TestResultMsg += "Test try to Timeout after Failing.\n";
+                        break;
+                    default:
+                        TestResultMsg += "How are you switching back to a Pending state from Failing.\n";
```
toddbellMSFT: nitpick: if this is a question, can it end in a question mark? Otherwise, as a statement, we could fit more info like:

"It should not be possible to switch back to a pending state from failing".

The statement sounds like it shouldn't be possible to reach this case. Should we be throwing an exception?

pgilmorepf:
retest

PR #571, opened 2019-06-20
Mimacken/xplat android

MichaelMackendMS:
XPlat Android in a working state (but not yet fully integrated with Jenkins).  Consists of 2 major parts:

1) testTitleData.json loading
2) renaming package to com.playfab.service

This is somewhat of a WIP.  It's a stable commit, but there are some code quality issues like script parameter usage checks and so forth that I have not gotten to yet.   Thanks!


SDKBuildScripts/xplat_BuildAndTestAndroid.sh
```diff
@@ -0,0 +1,35 @@
...
+
+BuildAPK() {
+    pushd "$AndroidProjectPath"
+    ./gradlew assembleDebug
+    popd
+}
+
+TestAPK() {
+    appcenter test run uitest --app "PlayFabSDKTeam/PlayFabXPlatAndroid" \
+    --devices ce2f4064 \
```
pgilmorepf: I feel like this should be an environment variable, either in the job itself, or in the Random Env Vars setup script (depending on how often it may change).
Not this commit, but just something to note.

targets/XPlatCppSdk/source/build/Android/XPlatCppSdk/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,5 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.microsoft.xplatcppsdk" >
+    package="com.playfab.xplatcppsdk" >
```
pgilmorepf: Yes, I agree. We don't want to confuse the issue right now.
PlayFab SDK's should stay com.playfab.etc until we eventually make a single big move to make it universally com.microsoft.playfab.etc or whatever.
But until then, we should avoid the confusion and stay com.playfab.

MichaelMackendMS: There's a more pragmatic reason too- the test assemblies all look for "com.playfab.service" when running on AppCenter. :)  All our apps need the same package name otherwise I would have to do some uggo matching code in the test assemblies.  yuck!

MichaelMackendMS:
retest

MichaelMackendMS:
retest

pgilmorepf:
retest

targets/XPlatCppSdk/source/build/Android/XPlatCppSdk/src/main/AndroidManifest.xml
pgilmorepf: Sounds awesome.

MichaelMackendMS:
I am closing this PR as it was really only generated for reference for @pgilmorepf.

MichaelMackendMS:
PSA: Please refrain from deleting this branch until I am ready!

PR #573, opened 2019-06-21
Fix parameter name collision

veleek:
As part of one of the last changes I modified the same of the parameter used in this foreach loop and it's now conflicting with the input parameter name which is causing problems.

### DIFF
--- a/SDKBuildScripts/Update-PlayFabSdk.ps1
+++ b/SDKBuildScripts/Update-PlayFabSdk.ps1
@@ -230,29 +230,29 @@ begin
 
 process
 {
-    foreach($sdkName in $SdkName)
+    foreach($targetSdkName in $SdkName)
     {
         $sdkTargetSource = $TargetSource
         if(!$sdkTargetSource)
         {
-            $sdkTargetSource = $sdkTargetSrcMap[$sdkName]
-            Write-Verbose "Setting Targetsource to $sdkTargetSource for $sdkName"
+            $sdkTargetSource = $sdkTargetSrcMap[$targetSdkName]
+            Write-Verbose "Setting Targetsource to $sdkTargetSource for $targetSdkName"
             if(!$sdkTargetSource)
             {
-                throw "Unable to determine TargetSource for '$sdkName'.  You must explicitly provide a value."
+                throw "Unable to determine TargetSource for '$targetSdkName'.  You must explicitly provide a value."
             }
         }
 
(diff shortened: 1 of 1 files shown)


SDKBuildScripts/Update-PlayFabSdk.ps1
```diff
@@ -230,29 +230,29 @@ begin
 
 process
 {
-    foreach($sdkName in $SdkName)
+    foreach($targetSdkName in $SdkName)
```
julioct: >$SdkName [](start = 30, length = 8)

Perhaps we could rename the $SdkName variable to $SdkNames to reflect it's a collection of string?

pgilmorepf: I think I agree with @julioct.
I mildly prefer to rename SdkName to indicate plurality (SdkNames, SdkNameList, etc)

veleek: I actually had it as $SdkNames before but I changed it to $SdkName because that seemed to be the standard that Powershell uses for most of it's APIs.

For example `Get-Item` has a `-Path` parameter which is a `string[]`, so you can either pass in one value or multiple values.  This prevents you from having to use `Update-PlayFabSdk -SdkNames CSharp ...` which can feel a little odd.

I'm happy to change it if you prefer, I just wanted to explain my reasoning first. :)

julioct: Fair enough, no issues then. Maybe just add a private $sdkNames variable to use in this loop while keeping the original as the public parameter

---
In reply to: [296388986](https://github.com/PlayFab/SDKGenerator/pull/573#discussion_r296388986) [](ancestors = 296388986)

veleek: Are you saying you just want to have something this?

```powershell
$sdkNames = $SdkName;
foreach($targetSdkName in $sdkNames)
{
  ...
}
```

julioct: Yea, but not feeling super strong there. Not a big deal

---
In reply to: [296408628](https://github.com/PlayFab/SDKGenerator/pull/573#discussion_r296408628) [](ancestors = 296408628)

veleek: Okay, gonna leave this as is for now.  @pgilmorepf, can you merge whenever?

PR #574, opened 2019-06-21
Pg/unity instance polish

pgilmorepf:


pgilmorepf:
Example Output:
https://github.com/PlayFab/UnitySDK/pull/215

targets/unity-v2/source/Shared/Public/PlayFabAuthenticationContext.cs
```diff
@@ -2,33 +2,59 @@ namespace PlayFab
...
     public sealed class PlayFabAuthenticationContext
     {
-        public string PlayFabId;
-
         public PlayFabAuthenticationContext()
         {
         }
 
-        public PlayFabAuthenticationContext(string clientSessionTicket, string entityToken, string playFabId) : this()
+        public PlayFabAuthenticationContext(string clientSessionTicket, string entityToken, string playFabId, string entityId, string entityType) : this()
```
Vicodex: >PlayFabAuthenticationContext [](start = 15, length = 28)

You are breaking people here. Make two last parameters null by default

pgilmorepf: It's my opinion that this specific constructor is meant for internal use.
Customers won't ever provide those parameters, because they can't know them in advance.

pgilmorepf: (Maybe that indicates needing to change that constructor to internal...)

pgilmorepf:
retest

PR #580, opened 2019-06-25
Add AuthenticationContext to make api call in AF local debugging

nima-ap:
This is a quick fix, there was a bug with the authentication context not being passed to `MakeApiCallWithFullUri` even though it was available to be passed. Was causing issues with missing `X-EntityToken` header as that's where it's being fetched from later down the line.


nima-ap:
retest

PR #585, opened 2019-06-28
Required mac building scripts

toddbellMSFT:
This check-in is required for master to work on the Mac builds for Unity.

### DIFF
--- a/SDKBuildScripts/mac_unity_RunAutoTests.sh
+++ b/SDKBuildScripts/mac_unity_RunAutoTests.sh
@@ -1,0 +1,267 @@
+#!/bin/bash
+
+# USAGE: unity_RunAutoTests.bat [<UNITY_VERSION>] [<SdkName>] [<ProjRootPath>] [<BuildIdentifier>]
+# Make folder links from the UnitySdk to this test project
+# Requires mklink which may require administrator
+# Requires the following environment variables:
+#   TestWin32Build - (Optional - Default false if unset) set to "true" to test building Win32, executing, and geting Jenker! results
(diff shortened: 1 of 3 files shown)


SDKBuildScripts/mac_unity_RunAutoTests.sh
```diff
@@ -0,0 +1,155 @@
...
+else
+    ProjRootPath="$3"
+fi
+if [ -z $4 ]; then
+    BuildIdentifier=JBuild_${SdkName}_${EXECUTOR_NUMBER}
+else
+    BuildIdentifier=$4
+fi
+
+CheckVars() {
```
MichaelMackendMS: Probably we can get rid of all these platform flags, since this script is specific to the mac build.

SDKBuildScripts/mac_unity_RunAutoTests.sh
```diff
@@ -0,0 +1,155 @@
...
+
+#boil error codes down to either 1 or 0.
+EC() {
+    if [ $1 -ne 0 ]; then return 1; else return 0; fi
+}
+
+DoWork() {
+    CheckVars
+    SetProjDefines
+    RunClientJenkernaught
```
MichaelMackendMS: Are we actually checking for cloudscript results in this case?  I thought the intention was to simply check compilation?  If we are meaning to check the test results, I didn't see any of the normal output I'd expect to see in the jenkins console output here: https://pf-jenkins.playfab.com/view/SDK/job/SDK_Work_Unity_MacOS/5/console

PR #587, opened 2019-07-12
Update PlayFab_InstanceAPI.cs.ejs

toddbellMSFT:
Quick revert. We were supposedly making this public and removing the getter.


toddbellMSFT:
retest

PR #592, opened 2019-08-22
Change AzureRegions to strings

toddbellMSFT:
AzureRegions must be treated as strings in our QoS C# code


debhaldarMS:
Are these static files ?

PR #598, opened 2019-09-05
Add the ability to add more assemblies into the UUnitIncrementalTestR…

pgilmorepf:
…unner.cs

### DIFF
--- a/targets/csharp/source/PlayFabSDK/source/Uunit/UUnitIncrementalTestRunner.cs
+++ b/targets/csharp/source/PlayFabSDK/source/Uunit/UUnitIncrementalTestRunner.cs
@@ -2,6 +2,7 @@ using PlayFab.Json;
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Reflection;
 using System.Threading.Tasks;
 
 #if !DISABLE_PLAYFABCLIENT_API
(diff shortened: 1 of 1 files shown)


targets/csharp/source/PlayFabSDK/source/Uunit/UUnitIncrementalTestRunner.cs
```diff
@@ -88,6 +88,11 @@ public static TestTitleData VerifyTestTitleData()
 #endif
         }
 
+        public static void AddTestAssembly(Assembly assembly, string filter = null)
```
toddbellMSFT: I see a weird build failure from our tests: 

" Uunit\UUnitIncrementalTestRunner.cs(91,44): error CS0246: The type or namespace name 'Assembly' could not be found (are you missing a using directive or an assembly reference?) [C:\program files\jenkins\workspace\SDK_Work_CSharp\sdks\CSharpBetaSdk\PlayFabSDK\source\PlayFabSDK.csproj]"

pgilmorepf: Good catch, thanks. I added the correct using when I was testing locally, but forgot to copy it over when I moved to SdkGenerator.

PR #610, opened 2019-10-21
Fulfilling customer request for ToString return information about errorDetails

pgilmorepf:
GenerateErrorReport returns all information about the error.
ToString was providing a streamlined set of information, but that was confusing to customers, so now ToString just calls GetErrorReport.
Ancillary changes:
Separating GetServerTime from UserDataApi, so that each test is distinct and less complicated
Fixing a few typos and misspellings.

### DIFF
--- a/targets/unity-v2/Testing/Tests/Client/ClientApiTests.cs
+++ b/targets/unity-v2/Testing/Tests/Client/ClientApiTests.cs
@@ -25,7 +25,7 @@ namespace PlayFab.UUnit
         private string _userEmail;
         private string PlayFabId;
 
-        // This test operates multi-threaded, so keep some thread-transfer varaibles
+        // This test operates multi-threaded, so keep some thread-transfer variables
         private int _testInteger;
 
         public override void ClassSetUp()
@@ -103,6 +103,9 @@ namespace PlayFab.UUnit
             var testContext = (UUnitTestContext)error.CustomData;
             testContext.False(errorReport.Contains("successful"), errorReport);
             testContext.True(errorReport.Contains("password"), errorReport);
+
+            testContext.StringEquals(error.GenerateErrorReport(), error.ToString());
+
             testContext.EndTest(UUnitFinishState.PASSED, null);
         }
 
@@ -210,7 +213,7 @@ namespace PlayFab.UUnit
         /// Test a sequence of calls that modifies saved data,
         ///   and verifies that the next sequential API call contains updated data.
         /// Verify that the data is correctly modified on the next call.
-        /// Parameter types tested: string, Dictionary&lt;string, string>, DateTime
+        /// Parameter types tested: string, Dictionary&lt;string, string>
         /// </summary>
         [UUnitTest]
         public void UserDataApi(UUnitTestContext testContext)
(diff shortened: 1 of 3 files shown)


toddbellMSFT:
Did you explicitly want to edit the meta png file? Seems like excess edits. 

Based on the title of the PR, I assumed I was just going to look at the last file. It took me a sec when I was reading tests for Timestamps which seems to be similar amount of work. I'll sign off on both, but I'd prefer to see them as separate micro PR's for cleaner git commit histories

targets/unity-v2/Testing/Tests/Client/ClientApiTests.cs
```diff
@@ -254,14 +257,31 @@ private void GetUserDataCallback2(GetUserDataResult result)
                 int.TryParse(userDataRecord.Value, out actualValue);
             testContext.IntEquals(_testInteger, actualValue);
             testContext.NotNull(userDataRecord, "UserData record not found");
+            testContext.EndTest(UUnitFinishState.PASSED, actualValue.ToString());
+        }
+
+        /// <summary>
+        /// CLIENT API
+        /// Test that the DateTime fields deserialize correctly, and accurately represent UTC time
+        /// Relies on the local machine time and time zone being set correctly.
```
toddbellMSFT: What does it mean to set the machine correctly? How will this work on our VMs? (I'm assuming we expect them to default to the time zone based on the data center?)

targets/unity-v2/Testing/Tests/Client/ClientApiTests.cs
```diff
@@ -254,14 +257,31 @@ private void GetUserDataCallback2(GetUserDataResult result)
...
-            var timeUpdated = userDataRecord.LastUpdated;
+            var serverTime = result.Time;
             var minTest = DateTime.UtcNow - TimeSpan.FromMinutes(5);
             var maxTest = DateTime.UtcNow + TimeSpan.FromMinutes(5);
-            testContext.True(minTest <= timeUpdated && timeUpdated <= maxTest);
+            testContext.True(minTest <= serverTime && serverTime <= maxTest);
 
-            testContext.True(Math.Abs((DateTime.UtcNow - timeUpdated).TotalMinutes) < 5); // Make sure that this timestamp is recent - This must also account for the difference between local machine time and server time
-            testContext.EndTest(UUnitFinishState.PASSED, actualValue.ToString());
+            testContext.True(Math.Abs((DateTime.UtcNow - serverTime).TotalMinutes) < 5); // Make sure that this timestamp is recent - This must also account for the difference between local machine time and server time
```
toddbellMSFT: Our documentation for [GetTime](https://api.playfab.com/documentation/Client/method/GetTime) seems to indicate we can have a potential variance of up to 5 seconds. Should this test be in minutes? or Should we expect/test stricter time limits?

pgilmorepf: The Server may have a variance of 5 minutes, but by experimentation I can tell you that cloud instances themselves have upto a ~2 minute variance in time, and that most local machines have the same variance. In fact, there's still a rare but plausible case where 5 minutes isn't even enough, but in practice it seems to be >99% stable.

targets/unity-v2/Testing/Tests/Client/ClientApiTests.cs
pgilmorepf: Basically to have a clock that is accurately synced to internet-time, within the given threshold, of 5 minutes. In my experience, if you look at a dozen or more machines, one machine can deviate from an authoritative clock by upto ~2 minutes, even if it's set "properly".

pgilmorepf:
I'm gonna admin land this.
The tests for Unity Windows passed, and some of the console specific tests are still offline, and therefore can't finish. These aren't changes that would affect consoles.

PR #612, opened 2019-10-23
Update playfab.js.ejs

toddbellMSFT:
This would enable users to test around the 443 SSL port. The customer reporting this said it helps with fiddler or mitmproxy


targets/js-node/source/Scripts/PlayFab/playfab.js.ejs
```diff
@@ -72,7 +72,7 @@ exports.MakeRequest = function (urlStr, request, authType, authValue, callback)
     if (options.protocol !== "https:")
         throw "Unsupported protocol: " + options.protocol;
     options.method = "POST";
-    options.port = 443;
+    options.port = options.port || 443;
```
pgilmorepf: This is probably fine.
For extra credits, you could additionally make this:
options.port || exports.settings.port;

But that would also require defining the port field in the settings, at the top, setting its default to 443, and updating the typescript definition of the settings to also include port as an int.

I'm happy with just this, or the extra credit too.

PR #618, opened 2019-11-01
Convert from HttpRequestMessage to HttpRequest

MGudgin:
This PR converts the various helpers in the CloudScript PlugIn that took
a System.Net.Http.HttpRequestMessage to use 
Microsoft.AspNetCore.Http.HttpRequest, which is the recommended type for
Azure Function running on the 2.0 runtime.

Details

Add PackageReference to Microsoft.AspNetCore.Http.

Amend the various Create methods in PlayFabFunctionContexts.cs to take
an HttpRequest. Add a helper, ReadBodyAsStringAsync to do the work
previously done by HttpContent.ReadAsStringAsync.

### DIFF
--- a/targets/csharp/source/Plugins/CloudScript/source/PlayFabCloudScriptPlugin.csproj.ejs
+++ b/targets/csharp/source/Plugins/CloudScript/source/PlayFabCloudScriptPlugin.csproj.ejs
@@ -45,6 +45,7 @@
   </PropertyGroup>
     
   <ItemGroup>
+      <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.2.2" />
      <PackageReference Include="PlayFabAllSDK" Version="<%- sdkVersion %>" />
   </ItemGroup>
 
--- a/targets/csharp/source/Plugins/CloudScript/source/PlayFabFunctionContexts.cs
+++ b/targets/csharp/source/Plugins/CloudScript/source/PlayFabFunctionContexts.cs
@@ -6,10 +6,11 @@
 #if NETCOREAPP2_0
 namespace PlayFab.Plugins.CloudScript
 {
+    using Microsoft.AspNetCore.Http;
     using PlayFab.Json;
     using System;
     using System.Collections.Generic;
-    using System.Net.Http;
+    using System.IO;
     using System.Threading.Tasks;
 
     /// <summary>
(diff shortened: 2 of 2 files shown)


targets/csharp/source/Plugins/CloudScript/source/PlayFabCloudScriptPlugin.csproj.ejs
```diff
@@ -45,6 +45,7 @@
   </PropertyGroup>
     
   <ItemGroup>
+      <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.2.2" />
```
pgilmorepf: Tabbing is inconsistent.

pgilmorepf: Tabbing is inconsistent.

targets/csharp/source/Plugins/CloudScript/source/PlayFabFunctionContexts.cs
```diff
@@ -108,33 +106,30 @@ public class FunctionPlayerPlayStreamContext<TFunctionArgument>
         /// </summary>
         /// <param name="request">The request incoming to an Azure Function from the PlayFab server</param>
         /// <returns>A new populated <c>FunctionPlayerPlayStreamContext</c> instance</returns>
-        public static async Task<FunctionPlayerPlayStreamContext<TFunctionArgument>> Create(HttpRequestMessage request)
+        public static async Task<FunctionPlayerPlayStreamContext<TFunctionArgument>> Create(HttpRequest request)
         {
-            using (var content = request.Content)
+            string body = await request.ReadBodyAsStringAsync();
+            var contextInternal = PlayFabSimpleJson.DeserializeObject<FunctionPlayerPlayStreamContextInternal>(body);
```
pgilmorepf: You should not be using PlayFabSimpleJson directly.
You should be fetching the serializer from PluginManager.
We JUST fixed a bunch of hard coded refs to PlayFabSimpleJson which were breaking the SDK when customers replaced the serializer.

targets/csharp/source/Plugins/CloudScript/source/PlayFabFunctionContexts.cs
```diff
@@ -38,32 +39,29 @@ public class FunctionContext<TFunctionArgument>
         /// </summary>
         /// <param name="request">The request incoming to an Azure Function from the PlayFab server</param>
         /// <returns>A new populated <c>FunctionContext</c> instance</returns>
-        public static async Task<FunctionContext<TFunctionArgument>> Create(HttpRequestMessage request)
+        public static async Task<FunctionContext<TFunctionArgument>> Create(HttpRequest request)
```
pgilmorepf: At a high level, I feel like you should be using the ITransport plugin, which performs HTTPS interactions for you in an implementation agnostic way.
You access it through PluginManager, and it allows the customer to replace the ITransport plugin with their own implementation.

pgilmorepf: I think that it's especially relevant since you seem to be replacing it here, which is exactly what ITransport is meant to avoid.
ALSO, ITransport will vary from one platform to another, and the default that comes with the SDK is guaranteed to work.

targets/csharp/source/Plugins/CloudScript/source/PlayFabFunctionContexts.cs
```diff
@@ -38,32 +39,29 @@ public class FunctionContext<TFunctionArgument>
...
         {
-            using (var content = request.Content) 
+            string body = await request.ReadBodyAsStringAsync();
+            var contextInternal = Json.PlayFabSimpleJson.DeserializeObject<FunctionContextInternal>(body);
```
pgilmorepf: Repeat comment: Don't use PlayFabSimpleJson directly.

targets/csharp/source/Plugins/CloudScript/source/PlayFabFunctionContexts.cs
```diff
@@ -108,33 +106,30 @@ public class FunctionPlayerPlayStreamContext<TFunctionArgument>
...
+
+            return new FunctionPlayerPlayStreamContext<TFunctionArgument>()
+            {
+                ApiSettings = settings,
+                AuthenticationContext = authContext,
+                PlayStreamEventEnvelope = contextInternal.PlayStreamEventEnvelope,
+                PlayerProfile = contextInternal.PlayerProfile,
+                FunctionArgument = contextInternal.FunctionArgument,
+                CurrentPlayerId = contextInternal.PlayerProfile.PlayerId
+            };
```
pgilmorepf: Can you help me understand why there's two seemingly identical copy paste copies of this code?

targets/csharp/source/Plugins/CloudScript/source/PlayFabFunctionContexts.cs
```diff
@@ -175,32 +170,29 @@ public class FunctionTaskContext<TFunctionArgument>
...
+
+            return new FunctionTaskContext<TFunctionArgument>()
+            {
+                ApiSettings = settings,
+                AuthenticationContext = authContext,
+                ScheduledTaskNameId = contextInternal.ScheduledTaskNameId,
+                EventHistory = contextInternal.EventHistory,
+                FunctionArgument = contextInternal.FunctionArgument
+            };
```
pgilmorepf: Or rather... three?

pgilmorepf:
At a high level: I think the best option is to implement the HTTP interface you want to use as an ITransport. Make that ITransport available to the system in PluginManager, and then make the CloudScript thingy utilize the ITransport plugin via PluginManager.GetPlugin.
In short, if you have a better http transport, we probably want to use it more than just here.

And the more things that integrate with the PluginManager properly, the fewer overall maintenance issues we'll have later.

As a bonus, you could add some unit tests to verify that the "active" ITransport plugin is capable of performing the calls that you need, with whatever signature is needed for your case.

What I don't want, is to have to support multiple simultaneous ITransports, because there's multiple hard coded ones in place. Same thing with the serializer refs, but that's a lot easier to fix.

pgilmorepf:
I have only just discovered that this is a separate solution from the SDK itself.
Maybe this is an issue for another time, but I think I'd prefer the project to exist at:
CSharpSDK\PlayFabSDK
and be part of the PlayFabSDK+Unit.sln
I'm not opposed to a separate solution as well, but the above solution is meant to be the manual-test-bed where we do all manual testing to ensure that the SDK is complete and working.

I didn't realize this plugin was totally separate, and so I missed making other changes in your plugin, which I had meant to make across the whole SDK.

pgilmorepf:
Aha, this PR hides the top lines:

/// THIS PLUGIN IS CURRENTLY IN A BETA STATE MEANT TO ONLY BE USED BY PRIVATE PREVIEW CUSTOMERS.
/// DO NOT RELY ON THE FOLLOWING CODE FOR USE ALONGSIDE PRODUCTION READY CODE.
/// THIS FILE IS SUBJECT TO CHANGE AT ANY TIME.

I guess, please accept my notes with the idea that I think they should be done for the "final version", but maybe aren't required for now.

PR #623, opened 2019-11-05
Toddbell/as fix

toddbellMSFT:
TreatmentAssignment appears in our ClientModels and our ServerModels. The error we see is "TreatmentAssignment is not a recognized class" BUT this class definitely exists in both Models packages. 

This PR is a quick fix that applies the specific package path when TreatmentAssignement is used. 

I'm unable to see why we can't just rely on what was working before (simple declaration without namespace). There are MANY classes that are "shared" between Client and Server, but ONLY TreatmentAssignment is having this compile problem? 

TreatmentAssignment is at least compiling standalone. So the main issue is not in the class itself, just the objects that use it as a member variable.

Potentially the cleanest approach would be to change the package namespace for every model to just be within the brackets of "package { }" and then import the Models namespace below? This would be a fundamental packaging change just for one particular class (but would be ideal for continued future maintenance as we wouldn't need to add any new objects to the make.js file).

### DIFF
--- a/targets/actionscript/make.js
+++ b/targets/actionscript/make.js
@@ -163,7 +163,7 @@ function getModelPropertyDef(property, datatype) {
     } else {
         if (property.optional && (basicType === "Boolean" || basicType === "int" || basicType === "uint" || basicType === "Number"))
             basicType = "*";
-        if (property.isclass) {
+        if (property.isclass && basicType === "TreatmentAssignment") {
             return property.name + ":" + getValidPlayFabActionScriptNamespacePrefix(datatype) + basicType;
         }
         else
@@ -217,7 +217,13 @@ function getModelPropertyInit(tabbing, property, datatype) {
             else
                 throw "Unknown collection type: " + property.collection + " for " + property.name + " in " + datatype.name;
         } else {
-            return tabbing + property.name + " = new " + getValidPlayFabActionScriptNamespacePrefix(datatype) + property.actualtype + "(data." + property.name + ");";
+            if (property.actualtype === "TreatmentAssignment") {
+                return tabbing + property.name + " = new " + getValidPlayFabActionScriptNamespacePrefix(datatype) + property.actualtype + "(data." + property.name + ");";
+            }
+            else {
+
+                return tabbing + property.name + " = new " + property.actualtype + "(data." + property.name + ");";
+            }
         }
     } else if (property.collection) {
         if (property.collection === "array") {
--- a/targets/actionscript/make.js
+++ b/targets/actionscript/make.js
@@ -123,7 +123,7 @@ function getVerticalNameDefault() {
     return "null";
 }
 
-function getValidActionscriptModelPath(datatype, basicType) {
+function getValidPlayFabActionScriptNamespacePrefix(datatype) {
 
     var asIndividualNamespaces = datatype.classNameSpace.split('.');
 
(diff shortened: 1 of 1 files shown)


targets/actionscript/make.js
```diff
@@ -123,6 +123,55 @@ function getVerticalNameDefault() {
...
+
+    return asNamespace;
+}
+
+function fullNamespaceRequired(propName)
+{
+    // NOTE: As of PlayFab version 191029, some objects are not properly getting recognized by ActionScript
+    // Due to the language getting deprecated within a year, we are going to add any breaking objects to this list
+    // and add the full namespace to any new object that breaks actionscript comiplation here.
+    var typesThatNeedFullNamespace = [ "TreatmentAssignment" ];
```
pgilmorepf: This line should be a global at the top of the file.

targets/actionscript/make.js
```diff
@@ -123,6 +123,55 @@ function getVerticalNameDefault() {
...
+
+    for(var i = 0; i < typesThatNeedFullNamespace.length; i++)
```
pgilmorepf: We have some list.contains() helpers in generate.ts
Search the project for an example usage and exact function name.

targets/actionscript/make.js
```diff
@@ -123,6 +123,55 @@ function getVerticalNameDefault() {
...
+            asNamespace += asIndividualNamespaces[i].toLowerCase();
+        }
+        else {
+            asNamespace += asIndividualNamespaces[i];
+        }
+
+        // we expect the input to be playfab.[SDK].[Name]
+        // but the actual datatype-callable prefix namespace in ActionScript is com.playfab.[SDK][Name].
+        if (i < asIndividualNamespaces.length - 2) {
+            asNamespace += ".";
```
pgilmorepf: String constructions like this in JS are very poor performance.
In JS, you should build a list of strings:
var collection = ["com.", "playfab", ".", "classnamespace", ".", "classname"]
and then do a join:
"".join(collection);

pgilmorepf: Repeated += on a string in scripting languages is an N^2 operation.
The array plus join is much closer to just an N operation.

targets/actionscript/make.js
```diff
@@ -184,7 +239,13 @@ function getModelPropertyInit(tabbing, property, datatype) {
             else
                 throw "Unknown collection type: " + property.collection + " for " + property.name + " in " + datatype.name;
         } else {
-            return tabbing + property.name + " = new " + property.actualtype + "(data." + property.name + ");";
+            if (fullNamespaceRequired(property.actualtype)) {
+                return tabbing + property.name + " = new " + getValidPlayFabActionScriptNamespacePrefix(datatype) + property.actualtype + "(data." + property.name + ");";
```
pgilmorepf: I'm almost thinking that
getValidPlayFabActionScriptNamespacePrefix
can just return empty string when !fullNamespaceRequired
That way you can simplify this block change.

targets/actionscript/make.js
```diff
@@ -136,7 +185,13 @@ function getModelPropertyDef(property, datatype) {
     } else {
         if (property.optional && (basicType === "Boolean" || basicType === "int" || basicType === "uint" || basicType === "Number"))
             basicType = "*";
-        return property.name + ":" + basicType;
+        if (property.isclass && fullNamespaceRequired(basicType)) {
+            return property.name + ":" + getValidPlayFabActionScriptNamespacePrefix(datatype) + basicType;
```
pgilmorepf: I'm almost thinking that
getValidPlayFabActionScriptNamespacePrefix
can just return empty string when !fullNamespaceRequired
That way you can simplify this block change.

PR #624, opened 2019-11-11
Don't run every `if` during enum conversion.

MSFT-Chris-Barrett:
This would prevent enum conversion functions from running the if-check
for every possible value during conversion.  There's no code in the
conversion functions after the if-checks, so there's no reason not to
leave when a matching value is found.  This is a tiny change that
_should_ be safe, but I haven't tested it.  See
PlayFabServerDataModels.h for country codes, currencies, generic error
codes, etc. for examples of giant functions running every if-check.

Better yet, make a bigger change to use a switch statement when
converting to json enums.  Result should be a condition-free conversion
that uses a jump table underneath for small enough enums, and falling
back on the behavior as currently written on larger enums.  Depending on
the compiler's "decision".

### DIFF
--- a/targets/WindowsSdk/templates/PlayFab_DataModels.h.ejs
+++ b/targets/WindowsSdk/templates/PlayFab_DataModels.h.ejs
@@ -19,14 +19,14 @@ for (var enumIdx = 0; enumIdx < enumtypes.length; enumIdx++) { var enumtype = en
         inline void ToJsonEnum(const <%- enumtype.name %> input, web::json::value& output)
         {
 <% for(var i=0; i<enumtype.enumvalues.length; i++) { var enumval = enumtype.enumvalues[i]
-%>            if (input == <%- enumtype.name %><%- enumval.name %>) output = web::json::value(L"<%- enumval.name %>");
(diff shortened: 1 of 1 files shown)


toddbellMSFT:
Thanks for the great feedback Chris! I will add this to our bug backlog and make the change over in the XplatCoreTemplate repo.

MSFT-Chris-Barrett:
Thanks for the quick response, Todd.  :)
What's the relation between the SDKGenerator and XplatCoreTemplate repos?  Is one better than the other for this sort of PR feedback?

toddbellMSFT:
SDKGenerator is the go to place for all SDKs with the exception of the C++ codebase. 

C++ sdk is now generated in the SDKGenerator with the XPlatCoreTemplates repo. In the SDKGenerator, check out the privateTemplates folder where it will have a script to auto clone the repo for you into that subfolder. 

This privateTemplates directory is meant for developers who may end up customizing our C++ SDK beyond our standardized XplatCoreTemplate repo (we have a lot of C++ 11 code that some devs would prefer to leave out, so they manage their own private template repo to pull into the SDKGenerator privateTemplates subfolder to generate their own version of PlayFab's C++ sdk).

toddbellMSFT:
Accepted this change in the core template repo: https://github.com/PlayFab/XPlatCoreTemplate/pull/49

PR #625, opened 2019-11-19
take request entitytoken into account before attempting to use static

toddbellMSFT:
and fix some warnings along the way


pgilmorepf:
Nevermind, a (thankfully) minor aspect of the data-isolation fix was missed, or somehow lost from the earlier fix.

targets/csharp/UnittestRunner/UUnitTestRunner.cs
```diff
@@ -5,12 +5,11 @@
 using PlayFab.ClientModels;
 using System.Threading.Tasks;
 
+#pragma warning disable 0649, 0414
```
toddbellMSFT: If you can remove the _onCompleted, then we can remove the warning supressions (they are both for unused fields/ unused assignments)

pgilmorepf: There's other unused in this file as well, so I left them.

pgilmorepf:
Manual test kick-off passed. Landing.

PR #626, opened 2019-11-22
Slight adjustment and comments for gitignore

pgilmorepf:


pgilmorepf:
Tests did not execute at all. *sigh*
Gitignore obviously can't break that, so I'm gonna admin land this one.

PR #627, opened 2019-11-22
Primary Fix: Add SynchronizationContextRemover, which fixes task deadlocks

pgilmorepf:
Added a new Forms app that tests the deadlocks in the most aggressive manner.
Simplified the CSharp template in the SDKGenerator a bit.
Added SynchronizationContextRemover in all the neccessary places
(Tested with the full unit test suite, so maybe not 100%, but it provides significant coverage)


SDKGenerator.njsproj
```diff
@@ -186,17 +188,50 @@
...
+    <Content Include="targets\csharp\source\PlayFabSDK\WindowsFormsApp1\bin\Debug\PlayFabAllSDK.dll" />
+    <Content Include="targets\csharp\source\PlayFabSDK\WindowsFormsApp1\bin\Debug\PlayFabAllSDK.pdb" />
+    <Content Include="targets\csharp\source\PlayFabSDK\WindowsFormsApp1\bin\Debug\WindowsFormsApp1.exe" />
+    <Content Include="targets\csharp\source\PlayFabSDK\WindowsFormsApp1\bin\Debug\WindowsFormsApp1.exe.config" />
+    <Content Include="targets\csharp\source\PlayFabSDK\WindowsFormsApp1\bin\Debug\WindowsFormsApp1.pdb" />
+    <Content Include="targets\csharp\source\PlayFabSDK\WindowsFormsApp1\Form1.cs" />
+    <Content Include="targets\csharp\source\PlayFabSDK\WindowsFormsApp1\Form1.Designer.cs" />
+    <Content Include="targets\csharp\source\PlayFabSDK\WindowsFormsApp1\Form1.resx" />
+    <Content Include="targets\csharp\source\PlayFabSDK\WindowsFormsApp1\obj\Debug\build.force" />
+    <Content Include="targets\csharp\source\PlayFabSDK\WindowsFormsApp1\obj\Debug\DesignTimeResolveAssemblyReferences.cache" />
```
toddbellMSFT: are all these obj directories necessary?

pgilmorepf: probably not...
(thanks)

targets/csharp/make.js
```diff
@@ -29,7 +29,6 @@ exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
     };
 
     templatizeTree(locals, path.resolve(sourceDir, "source"), rootOutputDir);
-    templatizeTree(locals, path.resolve(sourceDir, "UnittestRunner"), path.resolve(apiOutputDir, "UnittestRunner")); // Copy the actual unittest project in the CombinedAPI
```
toddbellMSFT: Will the source dir templatizeTree pick up tests for us now?

pgilmorepf: Yes, I just moved the same files into the "source" subfolder, one line up.

targets/csharp/source/PlayFabSDK/UnittestRunner/UUnitTestRunner.cs
```diff
@@ -18,15 +18,19 @@ public class CsSaveRequest
 
         public static int Main(string[] args)
         {
+            int result = 1;
+            bool taskFinished = false;
             try
             {
-                MainTask(args).Wait(60000);
+                var mainTask = MainTask(args);
+                taskFinished = mainTask.Wait(120000);
```
toddbellMSFT: I'm assuming this is a timeout? Can we make this a named variable to make sure we know time units (I'm assuming this is milliseconds?)

targets/csharp/source/PlayFabSDK/WindowsFormsApp1/Form1.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        public int Main2(string[] args)
+        {
+            int result = 1;
+            bool taskFinished = false;
+            try
+            {
+                var mainTask = MainTask(args);
+                taskFinished = mainTask.Wait(120000);
```
toddbellMSFT: Is it possible to make the previously mentioned named variable shared to be used in here? Or would that not make sense?

targets/csharp/source/PlayFabSDK/UnittestRunner/UUnitTestRunner.cs
pgilmorepf: Done.

targets/csharp/source/PlayFabSDK/WindowsFormsApp1/Form1.cs
pgilmorepf: They're actually separate projects, so I can't link them together. But I can make it an independent const.

targets/csharp/source/PlayFabSDK/WindowsFormsApp1/Form1.cs
```diff
@@ -0,0 +1,115 @@
...
+        private static void OnComplete(PlayFabResult<ExecuteCloudScriptResult> result)
+        {
+            WriteConsoleColor("Save to CloudScript result for: " + PlayFabSettings.BuildIdentifier + " => " + UUnitIncrementalTestRunner.PfClient.authenticationContext.PlayFabId, ConsoleColor.Gray);
+            if (result.Error != null)
+                WriteConsoleColor(result.Error.GenerateErrorReport(), ConsoleColor.Red);
+            else if (result.Result != null)
+                WriteConsoleColor("Successful!", ConsoleColor.Green);
+        }
+
+        private void button1_Click(object sender, EventArgs e)
```
toddbellMSFT: I'm assuming you have a button on your form then? At first glance it seems like an unneeded button callback

pgilmorepf: Actually, it's necessary. It kicks off the test suite.
To repro the bug, I needed the entire PlayFab test suite to be executed directly off of a button in the most "inappropriate" way, to demonstrate that even if a customer does it utterly wrong, our lib won't deadlock their app.

targets/csharp/source/PlayFabSDK/source/PlayFabUtil.cs.ejs
```diff
@@ -154,5 +155,39 @@ namespace PlayFab
             }
             return Sb.ToString();
         }
+
+        public struct SynchronizationContextRemover : INotifyCompletion
```
toddbellMSFT: Should this be a class?

pgilmorepf: This is a copy of known/good MS authored code, from an expert on this particular topic.
I avoided changing anything, such as struct/class.

PR #633, opened 2019-12-06
Toddbell/unreal ptr

toddbellMSFT:
A user has reported when attempting to package Unreal, SharedPtr's seem to not work across Instance based API's. Currently, PlayFab.h creates SharedPtr's in the Editor to work with a Standalone flag, but this won't work when the project is outside the editor. The user's got around the issue by wrapping the SharedPtr in a StrongRef. 

We can essentially use a flag as a replacement of the Standalone flag for the editor (which will treat the object as a strong ref on the first frame it exists, which seemed to work when I built a separate project).


targets/UnrealMarketplacePlugin/source/PlayFab/Source/PlayFabCpp/Public/PlayFab.h.ejs
```diff
@@ -20,6 +21,17 @@ namespace PlayFab
             Instance->ClearFlags(RF_Standalone);
         });
     }
+#else
+    template<typename T>
+    TSharedPtr<T> MakeSharedUObject()
+    {    
+        return TSharedPtr<T>(NewObject<T>((UObject*)GetTransientPackage(), NAME_None, RF_StrongRefOnFrame),
```
pgilmorepf: It seems like the RF_Standalone vs RF_StrongRefOnFrame is the only line-difference between the WITH_EDITOR vs not.
I would prefer to only #If/#else the one line that's different (for now).

PR #634, opened 2019-12-12
Prevent build flags from being overwritten

veleek:
Merge build flags from on args with ones loaded from genConfig.json.
- Currently if you specify the beta build flag it will be ignored for C#
  builds because the genConfig.json includes the CSharpOnly flag.
Update generate.js to support version override argument.
Fix some generate.js debug output.
Add -nowait parameter to enable skipping the delay at end of generate.
Updated date use in C# targets to be more Y3K compliant.
Updated Powershell script to support additional build flags and version.

### DIFF
--- a/SDKBuildScripts/Update-PlayFabSdk.ps1
+++ b/SDKBuildScripts/Update-PlayFabSdk.ps1
@@ -133,7 +133,7 @@ param(
     [AllowEmptyString()]
     [string]$ApiSpecGitUrl,
     [Parameter(ValueFromPipelineByPropertyName = $true)]
-    [string]$OutputPath = "..\..\sdks",
+    [string]$OutputPath,
     [Parameter(ValueFromPipelineByPropertyName = $true)]
     [string]$TargetSource,
     [Parameter(ValueFromPipelineByPropertyName = $true)]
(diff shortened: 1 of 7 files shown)


SDKBuildScripts/Update-PlayFabSdk.ps1
```diff
@@ -54,6 +54,15 @@ Indicites whether or not to remove all generated source files from the destinati
 .PARAMETER Beta
 Indicates whether or not to include any APIs tagged with as beta.
 
+.PARAMETER BuildFlags
```
pgilmorepf: As you're the primary user and more-or-less owner of this file, I'll approve any changes you want to make to this file, if they're made in isolation.
I can't approve this PR with the other changes though.
In particular, we should never make a PR with changes to multiple SDKs at once.

pgilmorepf: Oh wow, I missed a major problem with this PR.
You can't change the generate.js file directly, as it's an intermediate file generated from generate.ts.
You have to change the generate.ts file, which is the "source" file, and then commit both.
ALSO, that would have to be in a yet third PR.
As for the changes in generate.[ts], you are perhaps mis-understanding the purpose of genConfig.json.
Beta should not be generated to CSharpSDK at all, and we enforce the genConfig.json above the command line inputs on purpose.
If you want CSharp with Beta, you should be building to:
https://github.com/PlayFab/CSharpBetaSdk

toddbellMSFT:
Currently has conflicts and was not accepted in a timely manner. 

Closing the Pull Request. 

If it is still needed. Please make a new PR off the latest version of SDKGenerator for consideration.

PR #638, opened 2019-12-30
Rebuild the UnitySDK repo structure

pgilmorepf:
This will not be considered a breaking change, because the main entry point: the UnitySDK.unitypackage file, does not change in location or structure

Re-remove the Testing folder from the SDK package (should not have been added to it)
Fix the Testing folder to have two asmdef files, which properly identify the necessary compilation order
Add the EditorExtensions files to the templates and repo, and build the EditorExtensions package from the sdk repo.
Remove an unused template file.
Change some access levels for files that are needed across dll boundaries.
Move some variables and other bits around to fix errors with dll boundaries.
APIs now always provide explicit callSettings when calling MakeApiCall.
Server tests login with ServerCustomIdLogin before making server cloud script calls (fixes for new added validation on the PlayFab server - no more FakePlayFabId)
Fixed a warning in the latest version of Unity
Clarified staticSettings and staticPlayer in PlayFabSettings
All devSecretKey calls now validate the devSecretKey before making the call (Removed the related TODO comment)


targets/unity-v2/source/ExampleTestProject/Assets/PlayFabSDK/Shared/Internal/PlayFabHttp/PlayFabHTTP.cs
```diff
@@ -446,7 +446,7 @@ protected internal static void SendEvent(string apiEndpoint, PlayFabRequestCommo
             }
         }
 
-        protected internal static void ClearAllEvents()
+        public static void ClearAllEvents()
         {
```
toddbellMSFT: If this is no longer internal, should the file (or just this function) be moved to a more appropriate folder location name?

toddbellMSFT: Nope! This file is still under the PlayFab.Internal name space so this should be fine #resolved

targets/unity-v2/source/ExampleTestProject/Assets/PlayFabEditorExtensions/Editor/PlayFabEditorExtensions.asmdef
```diff
@@ -1,8 +1,6 @@
-{
-    "name": "PlayFabTestingEditor",
-    "references": [
-        "PlayFab"
-    ],
+﻿{
+    "name": "PlayFabEditorExtensions",
+    "references": [],
```
toddbellMSFT: So I know this was definitely needed when I was implementing the changes, but I do see there are some functions being made public rather than internal. I'd like to manually test this changeset myself to ensure this change is needed (because if so I think a lot of this file potentially becomes unneeded as the default file generated only has 1 or 2 values)

pgilmorepf: I think the "diff" here is misleading.
This is the new file copied as-is from Editor Extensions, and it should just appear as a pure add, not a diff.
The diff is a false comparison, and only seems to be here because the existing file is so close.
I just copied this straight from EdEx without any changes.

---
In reply to: [362102681](https://github.com/PlayFab/SDKGenerator/pull/638#discussion_r362102681) [](ancestors = 362102681)

toddbellMSFT: Ah sweet! I like that we now directly add the EditorExtension to SDKGenerator. There have been too many times testing the SDK where the editor extensions would be useful to have at the same time. 

Would we want to consider making these a git submodule in the UnitySDK repo? Or would we want to make this the final resting place of the editor extensions (and deprecate the separate public repo?)

---
In reply to: [362262535](https://github.com/PlayFab/SDKGenerator/pull/638#discussion_r362262535) [](ancestors = 362262535,362102681)

toddbellMSFT:
> #if ENABLE_PLAYFABPUBSUB_API

Is this file worth keeping around? I haven't heard an update on this which may make this dead code.

---
Refers to: targets/unity-v2/source/ExampleTestProject/Assets/Testing/Tests/PersistentSockets/PubSubTest.cs:1 in 0851ab3. [](commit_id = 0851ab35b23c1b69b80a48461a042cde498f6909, deletion_comment = False)

targets/unity-v2/source/ExampleTestProject/Assets/Testing/Editor/UUnitTestRunner.cs
```diff
@@ -27,16 +25,17 @@ private static void ClearDebugLog()
...
 
+        // From an empty scene, this will create the test scene
         [MenuItem("PlayFab/Testing/Play and Run Tests")]
         public static void TestImmediately()
         {
             ClearDebugLog();
 
-            _testers = UnityEngine.Object.FindObjectsOfType<UUnitIncrementalTestRunner>();
-            if (_testers.Length == 0)
+            var testComponents = UnityEngine.Object.FindObjectsOfType<UUnitIncrementalTestRunner>();
```
toddbellMSFT: >UUnitIncrementalTestRunner [](start = 70, length = 26)

I'm seeing an Error in Unity 2019.2 here stating "Assets\Testing\Editor\UUnitTestRunner.cs(34,71): error CS0246: The type or namespace name 'UUnitIncrementalTestRunner' could not be found (are you missing a using directive or an assembly reference?)"

[This also appears on line 37]

This file may need to be moved to the Assets/Testing/Tests/Shared/UUnit

Or PlayFab.UUnit classes may need to be in their own asmdef file to be added to the local PlayFabTestingEditor.asmdef file

pgilmorepf:
It's possible that these will be fixed with the addition of an asmdef file that I missed.
Please try again.

---
In reply to: [337448269](https://github.com/PlayFab/SDKGenerator/pull/638#issuecomment-337448269) [](ancestors = 337448269)

pgilmorepf:
PubSubTest.cs
It's an interesting point. We likely could delete it, but it's not the purpose of this pr.
Since it's already so big, I'd rather not take on any changes that I haven't already tested.

targets/unity-v2/source/ExampleTestProject/Assets/Testing/Editor/UUnitTestRunner.cs
pgilmorepf: It's possible that these will be fixed with the addition of an asmdef file that I missed.
Please try again.

pgilmorepf:
retest

pgilmorepf:
retest

pgilmorepf:
retest

pgilmorepf:
retest

SDKGenerator.njsproj
```diff
@@ -461,142 +461,264 @@
...
-    <Content Include="targets\unity-v2\source\Shared\Public\PluginContract.cs.meta" />
-    <Content Include="targets\unity-v2\source\Shared\Public\PluginContractKey.cs" />
-    <Content Include="targets\unity-v2\source\Shared\Public\PluginManager.cs" />
-    <Content Include="targets\unity-v2\source\Shared\Public\PluginManager.cs.meta" />
-    <Content Include="targets\unity-v2\source\Shared\Public\Resources.meta" />
-    <Content Include="targets\unity-v2\source\Shared\Public\Resources\PlayFabSharedSettings.asset" />
-    <Content Include="targets\unity-v2\source\Shared\Public\Resources\PlayFabSharedSettings.asset.meta" />
+    <Content Include="targets\unity-v2\source\ExampleTestProject\Assets\PlayFabEditorExtensions\Editor.meta" />
+    <Content Include="targets\unity-v2\source\ExampleTestProject\Assets\PlayFabEditorExtensions\Editor\PlayFabEditor.cs" />
+    <Content Include="targets\unity-v2\source\ExampleTestProject\Assets\PlayFabEditorExtensions\Editor\PlayFabEditor.cs.meta" />
```
toddbellMSFT: Can we leave the .meta files out of the SDKGenerator? These will prompt Unity to upgrade our example test project every time we open it (There's too many versions of Unity floating around to make this consistent). The meta files will be generated when they first load into Unity automatically. I think we should leverage that ability.

pgilmorepf:
retest

pgilmorepf:
retest

pgilmorepf:
retest

PR #643, opened 2020-01-28
Xamarin android was still referencing a deleted variable

pgilmorepf:


pgilmorepf:
retest

pgilmorepf:
Xamarin iOS failed, but Xamarin Android Passed.
That's exactly what we wanted for this change, so it's good to go.

PR #647, opened 2020-01-31
Get all the Unity tests updated and working

pgilmorepf:
Update all logs to write to WORKSPACE/logs (instead of all over the place)
Fix some log file names to reflect the actual work step, instead of copy-pasta over-write other logs.
Add "TestCompileFlags" which skips some steps for faster iterative testing.
Move unity_copyTestTitleData out of SetProjDefines, which are unrelated, and ensures unity_copyTestTitleData always happens.
The recent default build location was inside the Assets/ proj folder which fails for some platforms.
Change PS4 test builds to Development mode because of a bug
Log a little more thoroughly around the final test sequence (investigating bug - seemed like an ok thing to keep)

### DIFF
--- a/SDKBuildScripts/mac_unity_RunAutoTests.sh
+++ b/SDKBuildScripts/mac_unity_RunAutoTests.sh
@@ -19,16 +19,16 @@ BuildIdentifier=JBuild_${SdkName}_${VerticalName}_${NODE_NAME}_${EXECUTOR_NUMBER
 
 JenkernaughtSaveCloudScriptResults() {
     echo === Save test results to Jenkernaught ===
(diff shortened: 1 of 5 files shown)


pgilmorepf:
The brackets don't actually matter in this case ($WORKSPACE -> ${WORKSPACE}), but can be critical to identify the difference between something like:
$VARIABLE_SUFFIX and ${VARIABLE}_SUFFIX
Which would be interpreted differently.
In this case, WORKSPACE was never ambiguous, but rather just the result of an excessive mass-replace that I decided was ultimately "more correct" and decided to keep.

pgilmorepf:
Passed on switch, ps4, and core, so I'm landing it.
This PR also puts all those into "yellow" state for the current round of maintenance.

PR #653, opened 2020-02-11
Update PlayFabTestContext.h.ejs

toddbellMSFT:
LogPlayFabCommon doesn't link properly on a mac. LogTemp is commonly used throughout the Unreal codebase so this should still log the end test output, it will just label it differently as documented in the unreal wiki: https://wiki.unrealengine.com/Logs,_Printing_Messages_To_Yourself_During_Runtime

The LogPlayFabCommon is not available in the UPlayFabTestContext.h file (this file is stored in the ExampleProject/Plugin/PlayFab/Source and not under PlayFab/Source because we don't want to ship these tests with the plugin that gets distributed through the UnrealMarketplace).

We can move the declaration to the test context, but that will lose the PlayFabCommon definition. This is the quickest fix (and the wiki seems to only show using UE_LOG in a .cpp file, using it in a .h file for customer definitions may not be doable since it doesn't link on the Mac).


targets/UnrealMarketplacePlugin/examplesource/ExampleProject/Plugins/PlayFab/Source/PlayFabCommon/Public/TestFramework/PlayFabTestContext.h.ejs
```diff
@@ -115,7 +115,7 @@ public:
     UFUNCTION()
     void EndTest(PlayFabApiTestFinishState InFinishState = PlayFabApiTestFinishState::PASSED, FString InMessage = "")
     {
-        UE_LOG(LogPlayFabCommon, Log, TEXT("Ending Test: %s."), *testName);
+        UE_LOG(LogTemp, Log, TEXT("Ending Test: %s."), *testName);
```
pgilmorepf: I think that Unreal will possibly reject us if we don't use a properly defined log.
Maybe they won't catch us, but the point being, our customers must be able to identify that a log comes from us.

pgilmorepf:
Create a new log context for the tests if that's what's needed.

targets/UnrealMarketplacePlugin/examplesource/ExampleProject/Plugins/PlayFab/Source/PlayFabCommon/Public/TestFramework/PlayFabTestContext.h.ejs
toddbellMSFT: Note that we already use LogTemp in this file up top.

Since it is our test code, it should never show up in a customer's game runtime environment, so it shouldn't interfere with whatever output they are logging. Unreal shouldn't reject us on this basis, have they mentioned this previously when we were submitting PlayFab to the store?

pgilmorepf: OIC, if it's in the tests and not part of the Mkpl plugin, then that's probably fine.

generate.ts
```diff
@@ -179,7 +179,7 @@ function parseCommandInputs(args, argsByName: { [key: string]: string; }, errorM
     // Parse some other values and defaults
     if (!argsByName.buildidentifier)
         argsByName.buildidentifier = "default_manual_build";
-    argsByName.buildidentifier = argsByName.buildidentifier.toLowerCase(); // lowercase the buildIdentifier
+    argsByName.buildidentifier = argsByName.buildidentifier; // lowercase the buildIdentifier
```
pgilmorepf: Remove the comment.

toddbellMSFT: I'm a bit worried about this change if you were able to build Xplat Android just fine with the .toLowerCase present.

PR #655, opened 2020-02-12
Fix the XPlat Android automated testing

pgilmorepf:
Use the correct gradle commands to build a debug APK, identify the location and name of the release APK, set up some theoretical testing for a release APK, should we want to test it in the future (mostly left over research from fixing the debug build).
Two other issues that don't deserve their own PR.


SDKBuildScripts/xplat_BuildAndTestAndroid.sh
```diff
@@ -6,16 +6,6 @@ popd
 apkPath=$AndroidProjectPath/app/build/outputs/apk/debug/app-debug.apk
 testAssemblyDir="$1"
 
-Usage="./xplat_BuildAndTestAndroid.sh <path to test assemblies>"
```
MichaelMackendMS: why not check parameters?

SDKBuildScripts/xplat_BuildAndTestAndroid.sh
```diff
@@ -33,7 +23,7 @@ BuildAPK() {
     pushd "$AndroidProjectPath"
     ExitIfError
 
-    ./gradlew build
+    ./gradlew assembleDebug
```
MichaelMackendMS: Should parameterize the gradle task so as to facilitate testing different configurations?

pgilmorepf: Not a bad idea for other purposes, but if we want AppCenter to pass, this is the build parameter we must use.
I think if we set up a script for customer use, then perhaps it should be configurable.
But for now, we're just making the test work, so I'm going to say "won't fix"

SDKBuildScripts/xplat_BuildAndTestAndroid.sh
pgilmorepf: I'm not against it, but it's not being invoked, so it's just dead code as it is now.

SDKBuildScripts/xplat_BuildAndTestAndroid.sh
pgilmorepf: On this point, I've done some of the intermediate work for this, including, identifying the releaseApkPath, and setting up all the testing to use that path if it exists.
I'm sticking with assembleDebug for now, since it 1. Works, and 2. is faster.

targets/newTarget/source/exampleSourceFile.txt
```diff
@@ -6,13 +6,3 @@ You can have as many files and subdirectories as you want, and they're all copie
 
 EJS tags are ignored, so, <%- thisChunk %> is not evaluated in any way.
 (If it were, it would probably throw an exception about how "thisChunk" is not a known variable)
-
-
-
-
-***********************************************************************
-If you can see this, then the SDKGenerator is syncing the ghtest branch
```
MichaelMackendMS: just confirming- "ghtest" is not a typo?

pgilmorepf: ghtest was deliberate at the time.
It was short for "Git Hub [Gated checkin] Test"
This was a temporary change that was accidentally committed.
I only just noticed it for the first time.
It's not technically part of this feature, but trivially difficult, and minorly pointless, to separate to its own PR.

PR #665, opened 2020-02-21
Remove UE4 module export from PlayFab::Boxed

braindx:
Templated classes should not use the corresponding module macro to export their classes in Unreal. As an example of the problem, usage of `PlayFab::Boxed<int32>` in another module will cause an unresolved linker error when building Unreal Engine in the "DebugGame Editor" configuration. In this case, it prevents the implementation from being pulled in from the header due to the module macro usage, as it expects the PlayFab module to export the implementation which it does not.


pgilmorepf:
Thanks for your feedback.
We will review this suggestion to determine if your suggestion will cause any issues, and respond back soon.
Thanks.

toddbellMSFT:
This change was accepted into the Unreal Generation template: https://github.com/PlayFab/UnrealMarketplacePlugin/commit/e9e6288cf1d66d6cf65b6e9658548a1f3070bf96

PR #667, opened 2020-02-22
Detect build issues better, and fix the restore step to work properly

pgilmorepf:


SDKBuildScripts/xamarin_buildAppCenterTestIOS.sh
```diff
@@ -104,9 +98,9 @@ InitializeBuildEnvironment() {
     cp -r "$XamarinWorkspaceDirectory/PlayFabSDK" .
     cp -r "$XamarinWorkspaceDirectory/Plugins" .
     cp -r "$XamarinWorkspaceDirectory/XamarinTestRunner" .
-    echo "Loading test title data from $PF_TEST_TITLE_DATA_JSON into $ACB/XamarinTestRunner/XamarinTestRunner/XamarinTestRunner..."
+    echo "Loading test title data from $WORKSPACE/JenkinsSdkSetupScripts/Creds/testTitleData.json into $ACB/XamarinTestRunner/XamarinTestRunner/XamarinTestRunner..."
```
MichaelMackendMS: Why hardcode this?

pgilmorepf: It's dynamically located inside of the workspace for each individual job now.
It's different for every job, so we can't have a global definition for it anymore.

PR #674, opened 2020-03-17
Update to ScreenTimeTracker to expose the GameSessionID

MarcoWilliamsPF:
The intent is that you can use your own Game Session ID or have access to the one set internally to use in other telemetry calls.


pgilmorepf:
retest

PR #675, opened 2020-03-18
New UE Mac build script

pgilmorepf:


targets/UnrealMarketplacePlugin/examplesource/ExampleProject/BuildMac.sh.ejs
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+set -e
+
+DoWork() {
+    echo ========== BUILDING iOS ==========
```
Vicodex: The script name says Mac, but this comment says iOS. Should it be consistent?

targets/UnrealMarketplacePlugin/examplesource/ExampleProject/BuildMac.sh.ejs
```diff
@@ -0,0 +1,19 @@
...
+
+    targetPlatform=Mac
+    projectPath=$PWD
+    scriptsForProject="${projectPath}/ExampleProject.uproject"
+    uePath="/Users/Shared/Epic Games"
+    ueVersion="<%- ueTargetVersion %>"
+    uatPath="${uePath}/UE_${ueVersion}/Engine/Build/BatchFiles/"
+    cd "$uatPath"
```
Vicodex: uatPath already has quotes, no?

pgilmorepf: Not exactly. When you assign a variable in sh, using quotes, it ensures that all spaces and other things get into the variable. The quotes themselves don't actually end up in the variable. It can be very disruptive if they are.
But in this case, if there's spaces in the path, then I need to re-wrap them in quotes, so it doesn't cd into an unexpected location.
Sh is slightly different than bash in regard to quotes.

targets/UnrealMarketplacePlugin/examplesource/ExampleProject/BuildMac.sh.ejs
pgilmorepf: Good catch, I'll fix.

pgilmorepf:
Mac test passed, and that's all this change adds.

PR #681, opened 2020-03-31
Fix a Uobject memory chain issue in Test Framework

pgilmorepf:
As an Interface, rather than a UObject, IPlayFabTestRunner can't maintain the memory uchain
So, convert static fields on that to virtual accessors, and move the actual underlying object responsibility to the child objects, which will implement the interface.
Some objects were created without maintaining a reference to them, so add member properties, and tag them with UPROPERTY

Minor Refactors:
convert some dynamic_cast's to static_cast
Remove a bunch of unnecessary UFUNCTION's
Remove a handful of unnecessary virtual tags
Move some function implementations from .h file to .cpp file for EE.
Slight consistency improvement regarding placement of UPROPERTY tag.
Simplify TestRunner methods for clarity

You can see the template output the changes here:
https://github.com/PlayFab/UnrealMarketplacePlugin/pull/16/files


pgilmorepf:
retest

PR #687, opened 2020-04-04
Lua result calls wrong Entity variable

toddbellMSFT:
we missed an EntityToken change in lua in a previous fix of the same kind (notice the end of the line that was changed in this PR calls out the result's changed member).

 This somehow slipped passed our previous changes (the lua job only runs the generator, it doesn't precisely test lua).

### DIFF
--- a/targets/LuaSdk/make.js
+++ b/targets/LuaSdk/make.js
@@ -182,7 +182,7 @@ function getResultAction(tabbing, apiCall) {
         preCallback = internalTabbing + "PlayFabSettings._internalSettings.entityToken = result.EntityToken\n";
     else if (apiCall.result === "LoginResult") {
         preCallback = internalTabbing + "PlayFabSettings._internalSettings.sessionTicket = result.SessionTicket\n"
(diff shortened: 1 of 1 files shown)


targets/LuaSdk/make.js
```diff
@@ -182,7 +182,7 @@ function getResultAction(tabbing, apiCall) {
         preCallback = internalTabbing + "PlayFabSettings._internalSettings.entityToken = result.EntityToken\n";
     else if (apiCall.result === "LoginResult") {
         preCallback = internalTabbing + "PlayFabSettings._internalSettings.sessionTicket = result.SessionTicket\n"
-            + internalTabbing + "if (result.Entity) then PlayFabSettings._internalSettings.entityToken = result.EntityToken.EntityToken end\n";
+            + internalTabbing + "if (result.EntityToken) then PlayFabSettings._internalSettings.entityToken = result.EntityToken.EntityToken end\n";
```
pgilmorepf, line 185: While I suppose it's a good change, I don't think this would affect anything.
If either Entity or EntityToken is defined, the other should be defined as well.

PR #688, opened 2020-04-06
Modified file for the UnrealMarketplacePlugin to prevent build errors

MaxPlay:
Fixed Unreal build error in files that included "PlayFabResultCommon.h" where the type "UPlayFabJsonObject" wasn't found (due to missing forward declaration) and the "UPROPERTY"-macro wasn't resolved correctly, due to a missing include
 -> Added include for "PlayFab/Classes/PlayFabJsonObject.h" which fixes both errors at once

Note: This only happens when UnityBuild is disabled! So during default compiling it doesn't happen, but there are cases where this could make a build fail.


pgilmorepf:
We moved the Unreal template into its own destination repo.
I've migrated this PR to there:
https://github.com/PlayFab/UnrealMarketplacePlugin/pull/24

PR #692, opened 2020-04-20
Update API.java.ejs

toddbellMSFT:
We are returning null where we catch an exception. A team has reported that this hid an error from them, so we are going to pass along the hopefully helpful exception message to the user instead of leaving them with null.


toddbellMSFT:
I manually ran this change in Jenkins here: https://pf-jenkins.playfab.com/view/SDK/job/SDK_Work_Java/6742/console

PR #694, opened 2020-04-21
Resolve broken Node package

pgilmorepf:


lazerwalker:
I strongly suspect you also want to add `index.d.ts` to the `files` whitelist. The package.json file explicitly states that's the file that contains the TS type definitions for the library, but without it being included in `files` it won't actually exist in the version of the library that npm serves.

PR #699, opened 2020-04-24
Removing check for 404

toddbellMSFT:
We are skipping out on deserializing 404 strings. I'm assuming this was done as an optimization (what could they possibly need out of a "we didn't find the url, you probably should fix that")?

Apparently if you try to get mutliplayer servers with a mis-named server, this also returns 404 (our server couldn't find that particular build alias or match id). But our early cut out check only adds HttpStatus and Code, no other info (in fact, the Error comes out as Success, so it's a successful 404!). This does make sense though, the || statement clearly indicates an empty string will probably not survive the try/catch deserialization. Immediately after that check though, we seem to properly detect any error and pass all info appropriately along.

Fiddler shows there is a lot more info that we are missing (explicitly calls out what the user input as the alias).

### DIFF
--- a/targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabServerApiTest.cs
+++ b/targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabServerApiTest.cs
@@ -246,7 +246,33 @@ namespace PlayFab.UUnit
         [UUnitTest]
         public async void TestForNotFoundWithImportantInfo(UUnitTestContext testContext)
         {
-            PlayFabResult<MultiplayerModels.BuildAliasDetailsResponse> res = await UpdateAlias();
+            PlayFabSettings.staticSettings.TitleId = testTitleData.titleId;
+            PlayFabSettings.staticSettings.DeveloperSecretKey = testTitleData.developerSecretKey;
+
+            var eReq = new AuthenticationModels.GetEntityTokenRequest();
+            eReq.Entity = new AuthenticationModels.EntityKey();
+            eReq.Entity.Type = "title";
+            eReq.Entity.Id = testTitleData.titleId;
+
+            var tokenTask = await PlayFabAuthenticationAPI.GetEntityTokenAsync(eReq);
+
+            if(tokenTask.Error != null)
+            {
+                testContext.EndTest(UUnitFinishState.FAILED, "Failed to retrieve the Title Entity Token");
+                return;
+            }
+
+            MultiplayerModels.UpdateBuildAliasRequest updateBuildAliasRequest = new MultiplayerModels.UpdateBuildAliasRequest()
+            {
+                AliasId = "fakeAliasId",
+                AliasName = "aliasName",
+                AuthenticationContext = new PlayFab.PlayFabAuthenticationContext()
+                {
+                    EntityToken = tokenTask.Result != null ? tokenTask.Result.EntityToken : ""
+                },
+            };
+
+            PlayFab.PlayFabResult<MultiplayerModels.BuildAliasDetailsResponse> res = await PlayFab.PlayFabMultiplayerAPI.UpdateBuildAliasAsync(updateBuildAliasRequest);
 
             string response = PlayFab.PluginManager.GetPlugin<ISerializerPlugin>(PluginContract.PlayFab_Serializer).SerializeObject(res);
 
@@ -258,32 +284,6 @@ namespace PlayFab.UUnit
             {
                 testContext.EndTest(UUnitFinishState.FAILED, "We called the Mutliplayer API expecting to not find anything, but we didn't detect this to be the error.");
             }
-        }
-
-        public static async Task<PlayFabResult<MultiplayerModels.BuildAliasDetailsResponse>> UpdateAlias()
-        {
-            PlayFabSettings.staticSettings.TitleId = testTitleData.titleId;
-            PlayFabSettings.staticSettings.DeveloperSecretKey = testTitleData.developerSecretKey;
-
-            var eReq = new AuthenticationModels.GetEntityTokenRequest();
-            eReq.Entity = new AuthenticationModels.EntityKey();
-            eReq.Entity.Type = "title";
-            eReq.Entity.Id = testTitleData.titleId;
-
-            var tokenTask = await PlayFabAuthenticationAPI.GetEntityTokenAsync(eReq);
-            MultiplayerModels.UpdateBuildAliasRequest updateBuildAliasRequest = new MultiplayerModels.UpdateBuildAliasRequest()
-            {
-                AliasId = "fakeAliasId",
-                AliasName = "aliasName",
-                AuthenticationContext = new PlayFab.PlayFabAuthenticationContext()
-                {
-                    EntityToken = tokenTask.Result.EntityToken // entity token of the title
-                },
-            };
-
-            PlayFab.PlayFabResult<MultiplayerModels.BuildAliasDetailsResponse> res = await PlayFab.PlayFabMultiplayerAPI.UpdateBuildAliasAsync(updateBuildAliasRequest);
-
-            return res;
         }
     }
 }
--- a/targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabServerApiTest.cs
+++ b/targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabServerApiTest.cs
@@ -263,14 +263,12 @@ namespace PlayFab.UUnit
 
             if(tokenTask.Error != null)
             {
-                testContext.EndTest(UUnitFinishState.FAILED, "Failed to retrieve the Title Entity Token, check your playFabSettings.staticPlayer, are they still logged in? (hint no server api should be called through a logged in client)");
-                return;
+                testContext.Fail("Failed to retrieve the Title Entity Token, check your playFabSettings.staticPlayer, are they still logged in? (hint no server api should be called through a logged in client)");
             }
 
             if(aliasId == "")
             {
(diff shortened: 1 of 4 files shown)


targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabServerApiTest.cs
```diff
@@ -237,6 +240,62 @@ public void ParallelRequest(UUnitTestContext testContext)
...
+            eReq.Entity.Type = "title";
+            eReq.Entity.Id = testTitleData.titleId;
+
+            var tokenTask = await PlayFabAuthenticationAPI.GetEntityTokenAsync(eReq);
+
+            var staticPlayer = PlayFabSettings.staticPlayer;
+
+            if(tokenTask.Error != null)
+            {
+                testContext.EndTest(UUnitFinishState.FAILED, "Failed to retrieve the Title Entity Token, check your playFabSettings.staticPlayer, are they still logged in? (hint no server api should be called through a logged in client)");
```
pgilmorepf, line 266: testcontext.Fail(message);

targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabServerApiTest.cs
```diff
@@ -237,6 +240,62 @@ public void ParallelRequest(UUnitTestContext testContext)
...
+                return;
```
pgilmorepf, line 267: Using Fail(message) above won't require a return here.

targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabServerApiTest.cs
```diff
@@ -237,6 +240,62 @@ public void ParallelRequest(UUnitTestContext testContext)
...
+                },
+            };
+
+            PlayFab.PlayFabResult<MultiplayerModels.BuildAliasDetailsResponse> res = await PlayFab.PlayFabMultiplayerAPI.UpdateBuildAliasAsync(updateBuildAliasRequest);
+
+            string response = PlayFab.PluginManager.GetPlugin<ISerializerPlugin>(PluginContract.PlayFab_Serializer).SerializeObject(res);
+
+            if (response.Contains("MultiplayerServerNotFound"))
+            {
+                testContext.EndTest(UUnitFinishState.PASSED, "Detected the Expected MultiplayerServerNotFound PlayFabError");
```
pgilmorepf, line 292: You should also be able to check response.httpCode or something like that too, right?

targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabServerApiTest.cs
```diff
@@ -237,6 +240,60 @@ public void ParallelRequest(UUnitTestContext testContext)
...
+
+            string response = PlayFab.PluginManager.GetPlugin<ISerializerPlugin>(PluginContract.PlayFab_Serializer).SerializeObject(res);
+
+            if (response.Contains("MultiplayerServerNotFound") && res.Error.HttpCode == 404)
+            {
+                testContext.EndTest(UUnitFinishState.PASSED, "Detected the Expected MultiplayerServerNotFound PlayFabError");
+            }
+            else
+            {
+                testContext.Fail("We called the Mutliplayer API expecting to not find anything, but we didn't detect this to be the error.");
```
pgilmorepf, line 294: You should add response, and the http code to this error output, so we have a better idea what happened in this test from the log.
If there's a 1% failure in Jenkins, it will be a beast to debug. This at least starts us off.

targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabServerApiTest.cs
```diff
@@ -237,6 +240,60 @@ public void ParallelRequest(UUnitTestContext testContext)
...
+            PlayFabSettings.staticSettings.DeveloperSecretKey = testTitleData.developerSecretKey;
+
+            var eReq = new AuthenticationModels.GetEntityTokenRequest();
+            eReq.Entity = new AuthenticationModels.EntityKey();
+            eReq.Entity.Type = "title";
+            eReq.Entity.Id = testTitleData.titleId;
+
+            var tokenTask = await PlayFabAuthenticationAPI.GetEntityTokenAsync(eReq);
+
+            var staticPlayer = PlayFabSettings.staticPlayer;
```
pgilmorepf, line 262: Delete this line, it was temp code.

targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabServerApiTest.cs
```diff
@@ -237,6 +240,60 @@ public void ParallelRequest(UUnitTestContext testContext)
...
+
+            if(tokenTask.Error != null)
+            {
+                testContext.Fail("Failed to retrieve the Title Entity Token, check your playFabSettings.staticPlayer, are they still logged in? (hint no server api should be called through a logged in client)");
```
pgilmorepf, line 266: Instead of sending messages to your future self, just write the debug info into the message here.
Appending... staticPlayer.entityType would probably be enough.

pgilmorepf: Also, there's better ways to do checks like this:
testContext.IsNull(tokenTask.Error, "message");

targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabServerApiTest.cs
```diff
@@ -237,6 +240,60 @@ public void ParallelRequest(UUnitTestContext testContext)
...
+            }
+
+            if(aliasId == "")
```
pgilmorepf, line 269: testContext.StringEquals(aliasId, "", "message");

targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabServerApiTest.cs
```diff
@@ -237,6 +240,60 @@ public void ParallelRequest(UUnitTestContext testContext)
...
+                },
+            };
+
+            PlayFab.PlayFabResult<MultiplayerModels.BuildAliasDetailsResponse> res = await PlayFab.PlayFabMultiplayerAPI.UpdateBuildAliasAsync(updateBuildAliasRequest);
+
+            string response = PlayFab.PluginManager.GetPlugin<ISerializerPlugin>(PluginContract.PlayFab_Serializer).SerializeObject(res);
+
+            if (response.Contains("MultiplayerServerNotFound") && res.Error.HttpCode == 404)
+            {
+                testContext.EndTest(UUnitFinishState.PASSED, "Detected the Expected MultiplayerServerNotFound PlayFabError");
```
pgilmorepf, line 290: PASS messages should generally be empty.
We only want to put something here if it's diagnostic to testing (usually other tests, or the cloud-write-back to Jenkins).
In this case, I'd say set a null message here

targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabServerApiTest.cs
```diff
@@ -237,6 +240,60 @@ public void ParallelRequest(UUnitTestContext testContext)
...
+        {
+            PlayFabSettings.staticSettings.TitleId = testTitleData.titleId;
+            PlayFabSettings.staticSettings.DeveloperSecretKey = testTitleData.developerSecretKey;
+
+            var eReq = new AuthenticationModels.GetEntityTokenRequest();
+            eReq.Entity = new AuthenticationModels.EntityKey();
+            eReq.Entity.Type = "title";
+            eReq.Entity.Id = testTitleData.titleId;
+
+            var tokenTask = await PlayFabAuthenticationAPI.GetEntityTokenAsync(eReq);
```
pgilmorepf, line 260: I migrated all tests to not use static interface anymore.
Please convert this to create and use a local instance.
(this would have avoided your earlier issue as well)

targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabServerApiTest.cs
toddbellMSFT: StringEquals is not what we want here. We would need to check for not equals. If these two strings match, we should fail. But how StringEquals works, if my valid aliasId is not "" than we end the test there (which we would only want this case to pass otherwise an empty alias would give us the incorrect information (bad parameter) that we expect back (multiplayer server not found)

targets/csharp/source/PlayFabSDK/source/Uunit/tests/PlayFabServerApiTest.cs
```diff
@@ -237,6 +240,59 @@ public void ParallelRequest(UUnitTestContext testContext)
...
+
+            string response = PlayFab.PluginManager.GetPlugin<ISerializerPlugin>(PluginContract.PlayFab_Serializer).SerializeObject(res);
+
+            if (response.Contains("MultiplayerServerNotFound") && res.Error.HttpCode == 404)
+            {
+                testContext.EndTest(UUnitFinishState.PASSED, null);
+            }
+            else
+            {
+                testContext.Fail("We called the Mutliplayer API expecting to not find anything, but we didn't detect this to be the error. Details: " + res.Error.ErrorMessage + " and http code: "+res.Error.HttpCode);
```
pgilmorepf, line 293:
```suggestion
                testContext.Fail("We called the Mutliplayer API expecting to not find anything, but we didn't detect this to be the error. Details: " + res.Error.GenerateErrorReport() + " and http code: "+res.Error.HttpCode);
```

PR #703, opened 2020-05-04
Stop using string interpolation to be more compatible with older vers…

pgilmorepf:
…ions of msbuild

### DIFF
--- a/JenkinsConsoleUtility/jcuSrc/Commands/CalendarAlertCommand.cs
+++ b/JenkinsConsoleUtility/jcuSrc/Commands/CalendarAlertCommand.cs
@@ -49,9 +49,9 @@ namespace JenkinsConsoleUtility.jcuSrc.Commands
                 var alertTime = expireTime - TimeSpan.FromDays(alertDays);
 
                 if (today >= expireTime)
-                    expiredAlerts.Add($"Expired alert: {eachAlert.name} on: {eachAlert.date}");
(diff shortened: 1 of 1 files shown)


pgilmorepf:
The exact requirements/deficiencies of the older msbuild we're using isn't clear.
It's not the version of .NET, as we're already pinned to 4.5.
It might be possible to identify and test the exact version of C# and #if on the version, but that'll take more research. And it can only be tested/verified on the target machine.

JenkinsConsoleUtility/jcuSrc/Commands/CalendarAlertCommand.cs
```diff
@@ -49,9 +49,9 @@ public int Execute(Dictionary<string, string> argsLc, Dictionary<string, string>
                 var alertTime = expireTime - TimeSpan.FromDays(alertDays);
 
                 if (today >= expireTime)
-                    expiredAlerts.Add($"Expired alert: {eachAlert.name} on: {eachAlert.date}");
+                    expiredAlerts.Add("Expired alert: " + eachAlert.name + " on: " + eachAlert.date);
```
toddbellMSFT, line 52: Aww, but that's such a cool "new" feature :( But yeah, makes back-compat hard

pgilmorepf: /agree

PR #706, opened 2020-05-26
Toddbell/java server test fix

toddbellMSFT:
While updating the Java SDK for minecraft earth. Our tests were originally expecting a null return where we now return a detailed PlayFab error.

### DIFF
--- a/targets/java/source_shared/src/test/java/com/playfab/test/PlayFabApiTest.java.ejs
+++ b/targets/java/source_shared/src/test/java/com/playfab/test/PlayFabApiTest.java.ejs
@@ -444,7 +444,6 @@ public class PlayFabApiTest
         hwRequest.FunctionName = "helloWorld";
         hwRequest.PlayFabId = loginRes.Result.PlayFabId;
         PlayFabResult<PlayFabServerModels.ExecuteCloudScriptResult> hwResult = PlayFabServerAPI.ExecuteCloudScript(hwRequest);
-        
         assertNotNull(hwResult.Result.FunctionResult);
         Map<String, String> arbitraryResults = (Map<String, String>)hwResult.Result.FunctionResult;
         assertEquals(arbitraryResults.get("messageValue"), "Hello " + loginRes.Result.PlayFabId + "!");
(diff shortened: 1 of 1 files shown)


targets/java/source_shared/src/test/java/com/playfab/test/PlayFabApiTest.java.ejs
```diff
@@ -455,10 +460,8 @@ public class PlayFabApiTest
         errRequest.FunctionName = "throwError";
         errRequest.PlayFabId = playFabId;
         PlayFabResult<PlayFabServerModels.ExecuteCloudScriptResult> errResult = PlayFabServerAPI.ExecuteCloudScript(errRequest);
-        VerifyResult(errResult, true);
-        assertTrue(errResult.Result.FunctionResult == null);
-        assertNotNull(errResult.Result.Error);
-        assertEquals(errResult.Result.Error.Error, "JavascriptException");
+        String errorMessage = CompileErrorsFromResult(errResult);
```
pgilmorepf, line 463: Does errResult.GenerateErrorReport not exist?

targets/java/source_shared/src/test/java/com/playfab/test/PlayFabApiTest.java.ejs
```diff
@@ -434,14 +434,19 @@ public class PlayFabApiTest
     @Test
     public void CloudScriptServer()
     {
+        PlayFabServerModels.LoginWithServerCustomIdRequest customIdReq = new PlayFabServerModels.LoginWithServerCustomIdRequest();
+        customIdReq.CreateAccount = true;
+        String serverId = playFabId;
```
pgilmorepf, line 439: No need for a temporary variable.

targets/java/source_shared/src/test/java/com/playfab/test/PlayFabApiTest.java.ejs
```diff
@@ -434,14 +434,19 @@ public class PlayFabApiTest
...
+        customIdReq.ServerCustomId = serverId;
```
pgilmorepf, line 440: This should use the buildIdentifier the same way the client tests use.
If the server uses a fixed ID, then multiple simultaneous runs of this test will break, all trying to simultaneously access the same player.

targets/java/source_shared/src/test/java/com/playfab/test/PlayFabApiTest.java.ejs
toddbellMSFT: Nope, it looks like similar logic is in the CompileErrorsFromResult function (takes the error message and appends all error details to a single string).

PR #709, opened 2020-05-29
Adding singular retries to flaky tests

toddbellMSFT:
possible timeout issue though.

We could instead move this logic up to the test runner level, and explicitly call out ParallelTestRequest and LoginWithAdvert as tests to retry once more, but should probably alot the whole (currently 15 second) timeout again.


toddbellMSFT:
We will proceed with a better way than a one off retry.

PR #713, opened 2020-06-16
It's now possible to define -genconfigprofilename in the generate.js …

pgilmorepf:
…args to specify which profile from genConfig.json to read

### DIFF
--- a/SDKGenerator.njsproj
+++ b/SDKGenerator.njsproj
@@ -18,7 +18,7 @@
     <ProjectGuid>{2ba807ef-310c-43b5-8d42-1abcfff39be9}</ProjectGuid>
     <ProjectHome>.</ProjectHome>
     <ProjectView>ShowAllFiles</ProjectView>
-    <StartupFile>generate.js</StartupFile>
+    <StartupFile>generate.ts</StartupFile>
     <WorkingDirectory>.</WorkingDirectory>
     <OutputPath>.</OutputPath>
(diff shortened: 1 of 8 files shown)


SDKGenerator.njsproj
```diff
@@ -18,7 +18,7 @@
     <ProjectGuid>{2ba807ef-310c-43b5-8d42-1abcfff39be9}</ProjectGuid>
     <ProjectHome>.</ProjectHome>
     <ProjectView>ShowAllFiles</ProjectView>
-    <StartupFile>generate.js</StartupFile>
+    <StartupFile>generate.ts</StartupFile>
```
toddbellMSFT, line 21: OH, that's different

generate.js
```diff
@@ -1,3 +1,6 @@
+/// <reference path="node.d.ts"/>
```
toddbellMSFT, line 1: Should these comments be left in?

generate.ts
```diff
@@ -1,3 +1,7 @@
+/// <reference path="node.d.ts"/>
```
toddbellMSFT, line 1: same comment from previous file, are these comments needed

SDKGenerator.njsproj
pgilmorepf: The sln seems to not create the js file properly.
I was experimenting.
VS should understand what .ts files are though, and it built correctly after I manually regenerated the .js files using tsc.

PR #718, opened 2020-06-18
targets/panda3DSdK: Add Panda3d Python support

thetestgame:
This build target functions similarly to the already existing PythonSdk. However it replaces requests with the engine's built in HTTPClient to allow for async http calls that can nicely coexist with the rest of the engine. In a normal scenario using Panda3D having blocking calls in critical points in the application can cause unwanted application lag due to Python's single threaded nature. 

This was created using what knowledge I have of PlayFab's requirements. I am open to adjusting things as required if there is interest in adding Panda3D support to PlayFab's SDK list.

### DIFF
--- a/targets/Panda3DSdk/Archive/Enum.py.ejs
+++ b/targets/Panda3DSdk/Archive/Enum.py.ejs
@@ -1,0 +1,4 @@
+<%- generateApiSummary("    ", datatype, "description")
+%>class <%- datatype.name %> (enum.Enum):<% for(var i = 0; i < datatype.enumvalues.length; i++)%>
+    <%-  datatype.enumvalues[i].name %> = <%-i%>
+
--- a/targets/Panda3DSdk/Archive/Model.py.ejs
+++ b/targets/Panda3DSdk/Archive/Model.py.ejs
@@ -1,0 +1,23 @@
+<%- generateApiSummary("", datatype, "description")
+%><%- getDeprecationAttribute("", datatype)
+%>class <%- datatype.name %><%- getBaseTypeSyntax(datatype) %>:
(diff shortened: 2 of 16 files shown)


pgilmorepf:
Hello!
We're adopting a new customer supported SDK path, where you can host the template and the SDK on your own GitHub account.

**I'd like to direct you to:**
Templates no longer need to be embedded directly into the SdkGenerator.
https://github.com/PlayFab/UnrealMarketplacePlugin
As an example of how to do this.

**Explanation:**
Assuming {depot} is just the root folder where you keep GitHub repos.
And the specific repo is denoted with <repoName>.
Then, if you sync as follows:
{depot}/<SdkGenerator>
{depot}/SdkGenerator/privateTemplates/<UnrealMarketplacePlugin>
{depot}/sdks/<UnrealMarketplacePlugin>
Then, this SDK will both read and write from the same repo (different copies), and both the templates and the SDK exist in the same repo.
You can keep your Panda3d repo in your own GitHub account as well.

**Alternately:**
You can keep your template and SDK in separate repos like this:
https://github.com/PlayFab/XPlatCoreTemplate and
https://github.com/PlayFab/XPlatCppSdk
Synced as follows:
{depot}/<SdkGenerator>
{depot}/SdkGenerator/privateTemplates/<XplatCoreTemplate>
{depot}/sdks/<XPlatCppSdk>

**Conclusion:**
We encourage you to keep your SDK template, and your SDK in your own GitHub account. They can be separate repos, or even merged in the same repo.
I will be declining this PR, but you can feel free to post the existence of this SDK in our forums:
https://community.playfab.com/index.html
And share with other developers there.

Thank you for your submission, and feel free to post feedback about this new process we're trying out.

PR #720, opened 2020-07-10
Update the Xamarin.UITest version

pgilmorepf:
Step 1 of 2, the TestLauncher app uses Xamarin.UITest, and was set to a particular version.
The other half is installed directly onto the test machines (for now), and that installation either broke or changed (somehow).
Since I'm reconstructing this system, I opted to also update to the latest version.

This commit just updates the TestLauncher app to the latest version.

The other commit is here:
https://github.com/PlayFab/JenkinsSdkSetupScripts/pull/80


pgilmorepf:
Force landing without review since Todd's not available

PR #725, opened 2020-07-22
Remove AC3 & update default build config console params

pgilmorepf:
Remove compilation of some TypeScript that should never have been (should have been read only)
Remove flag sets in the VS build options that won't work from here anymore (simply unused now)


pgilmorepf:
Not a perfect test pass:
One random failure, and existing known failures (in progress in another PR), but sufficient to identify that there's no effect on existing non-as3 jobs.

PR #728, opened 2020-08-12
Update PlayFabEvents.cs.ejs

Soth1985:
Split large methods into batches that process no more than 10 API calls. Fixes https://github.com/PlayFab/UnitySDK/issues/237


toddbellMSFT:
We cannot accept this code change at this time. If you would like to see this in future releases, please post to the playfab developer forums.

PR #732, opened 2020-09-09
Update PlayFabPackager.cs

toddbellMSFT:
Wasn't there no good way to enable/disable these in Jenkins? Somehow we lost this tag, so when we were building in AppCenter we had the Gemfile and script to build it, but the xcode project that unity output never got the links to calabash that were required.  The quickest fix is to just remove the ability to call this code on only UNITY_IOS labels

### DIFF
--- a/targets/unity-v2/source/ExampleTestProject/Assets/Testing/Editor/PlayFabPackager.cs
+++ b/targets/unity-v2/source/ExampleTestProject/Assets/Testing/Editor/PlayFabPackager.cs
@@ -15,7 +15,13 @@ namespace PlayFab.Internal
         public int callbackOrder { get { return 0; } }
         public void OnPostprocessBuild(BuildReport report)
         {
-            OnPostprocessBuildiOS(report);
+            try {
+                OnPostprocessBuildiOS(report);
+            }
+            catch(Exception e)
+            {
+                Debug.Log("Unhandled exception while attempting to add iOS frameworks with: \n" + e.Message);
+            }
         }
 
         private void OnPostprocessBuildiOS(BuildReport report)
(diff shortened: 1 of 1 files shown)


targets/unity-v2/source/ExampleTestProject/Assets/Testing/Editor/PlayFabPackager.cs
```diff
@@ -34,7 +40,7 @@ private void OnPostprocessBuildiOS(BuildReport report)
             var proj = new UnityEditor.iOS.Xcode.PBXProject();
 
             proj.ReadFromString(File.ReadAllText(projectPath));
-            string xcodeTargetGUID = proj.TargetGuidByName("Unity-iPhone");
+            string xcodeTargetGUID = proj.GetUnityMainTargetGuid();
```
MichaelMackendMS, line 43: How do we know the MainTargetGuid will be our desired "Unity-iPhone"?

targets/unity-v2/source/ExampleTestProject/Assets/Testing/Editor/PlayFabPackager.cs
```diff
@@ -15,12 +15,18 @@ class TestAppPostBuildProcessor : IPostprocessBuildWithReport
...
+            catch(Exception e)
+            {
+                Debug.Log("Unhandled exception while attempting to add iOS frameworks with: \n" + e.Message);
+            }
         }
 
         private void OnPostprocessBuildiOS(BuildReport report)
         {
-#if UNITY_IOS
+#if UNITY_IOS || UNITY_MAC
```
MichaelMackendMS, line 29: Why add UNITY_MAC here?

MichaelMackendMS (changes requested):

targets/unity-v2/source/ExampleTestProject/Assets/Testing/Editor/PlayFabPackager.cs
toddbellMSFT: Paul requested it offline, we have 2 jenkins jobs for mac and iOS that may leverage this codepath.

PR #733, opened 2020-09-25
End tests early if we see failures

toddbellMSFT:
This will help ADO, and Jenkins should probably be able to understand an exe failing early without needing to check cloudscript (if it failed, why check?)

Jenkins may not relay this failure back though just because the test ends though (will consoles be able to do the same?) 

We can add this check to only run for iOS since this is the main platform that ado is having trouble with.

### DIFF
--- a/targets/unity-v2/source/ExampleTestProject/Assets/Testing/Tests/Shared/Uunit/UUnitIncrementalTestRunner.cs
+++ b/targets/unity-v2/source/ExampleTestProject/Assets/Testing/Tests/Shared/Uunit/UUnitIncrementalTestRunner.cs
@@ -75,6 +75,12 @@ namespace PlayFab.UUnit
 #else
         private void OnSuiteFinish()
         {
+            var report = suite.GetInternalReport();
+            if(report.failures > 0)
+            {
+                throw new Exception("Tests have failed! Ending our tests early, see this Test Summary\n" + suite.GenerateTestSummary());
+            }
+
             if (postResultsToCloudscript)
                 PostTestResultsToCloudScript();
             else
--- a/targets/unity-v2/source/ExampleTestProject/Assets/Testing/Tests/Shared/Uunit/UUnitIncrementalTestRunner.cs
+++ b/targets/unity-v2/source/ExampleTestProject/Assets/Testing/Tests/Shared/Uunit/UUnitIncrementalTestRunner.cs
@@ -81,15 +81,6 @@ namespace PlayFab.UUnit
                 OnCloudScriptSubmit(null);
         }
 
-        private void FaultRunIfFailed()
-        {
-            var report = suite.GetInternalReport();
-            if(report.failures > 0)
-            {
-                throw new Exception("Tests have failed! Ending our tests early, see this Test Summary\n" + suite.GenerateTestSummary());
-            }
-        }
-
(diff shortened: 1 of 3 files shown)


Vicodex (approved):

targets/unity-v2/source/ExampleTestProject/Assets/Testing/Tests/Shared/Uunit/UUnitIncrementalTestRunner.cs
```diff
@@ -75,6 +75,12 @@ private void OnSuiteFinish()
 #else
         private void OnSuiteFinish()
         {
+            var report = suite.GetInternalReport();
+            if(report.failures > 0)
+            {
+                throw new Exception("Tests have failed! Ending our tests early, see this Test Summary\n" + suite.GenerateTestSummary());
```
pgilmorepf, line 81: PostTestResultsToCloudScript is how the tests leave the device and get back to the Test Coordinator (ADO or Jenkins).
If you throw before it can do that, then you can't ever get the results back.

pgilmorepf (changes requested):

targets/unity-v2/source/ExampleTestProject/Assets/Testing/Tests/Shared/Uunit/UUnitIncrementalTestRunner.cs
toddbellMSFT: It shortcuts certain issues that we have had with cloudscript (upload limits, hanging waiting for builds that never made it up there etc.) 

But to ensure we maintain current behavior, we can move this after both the login success/failure if they happen.

targets/unity-v2/source/ExampleTestProject/Assets/Testing/Tests/Shared/Uunit/UUnitIncrementalTestRunner.cs
```diff
@@ -74,17 +74,20 @@ private void OnSuiteFinish()
...
 #else
         private void OnSuiteFinish()
+        {
+            if (postResultsToCloudscript)
+                PostTestResultsToCloudScript();
+            else
+                OnCloudScriptSubmit(null);
+        }
+
+        private void FaultRunIfFailed()
```
pgilmorepf, line 84: My only remaining comment would be the function-order. It seems that the functions in this file are in sequential order, in terms of the order that they are called in the testing process.
For that reason, it seems FaultRunIfFailed should be moved to the end, after QuitTesting (as it's only called by QuitTesting).

targets/unity-v2/source/ExampleTestProject/Assets/Testing/Tests/Shared/Uunit/UUnitIncrementalTestRunner.cs
pgilmorepf: I accept that the Cloud Script based solution isn't perfect, but we shouldn't make the problem worse by introducing more ways for it to fail. Intermediate solutions shouldn't make the end-goal harder to achieve.

pgilmorepf (approved):
Request: reorder the added functions to maintain "sequence of events" ordering that presently exists.
Otherwise approved.

PR #734, opened 2020-09-29
Removing Accept-Encoding from future web requests

toddbellMSFT:
Unity 2020 will throw a ton of warnings if you include PlayFab anywhere in your Asset folder.

Supposedly Unity has changed how their WebRequests are working under the hood.

We should only need to declare Content-Encoding, Accept-Encoding will either not be looked at by Unity, or certain platforms may experience undefined behavior.

It is recommended we don't fill in Accept-Encoding.


pgilmorepf (changes requested):
We should not disable the gzip encoding.

PR #739, opened 2020-11-09
Allow the write of multiple test result files.

pgilmorepf:
Stop relying on build identifier (ADO build identifier is incorrect atm)
Update tests to run local tests (ancient bug from CSharp nuget conversion).


pgilmorepf:
rebuild

toddbellMSFT (approved):

PR #742, opened 2021-01-12
Parse the days input properly.

pgilmorepf:
Call Builds-List for only one pipeline at a time, in order to ensure we get builds for each pipeline.
Call Pipelines-List in order to find every pipeline
Filter Pipelines-List by the ones that start with "\PlayFabSdk", which hard-locks the folder we consider "finished pipelines", but also matches current expectation
Tweak the task calling a bit.
Cache everything to files a little more thoroughly for testing.

### DIFF
--- a/JenkinsConsoleUtility/jcuSrc/Commands/GetAdoBuilds.cs
+++ b/JenkinsConsoleUtility/jcuSrc/Commands/GetAdoBuilds.cs
@@ -27,7 +27,8 @@ namespace JenkinsConsoleUtility.jcuSrc.Commands
         {
             List<AdoBuildResultRow> rows = new List<AdoBuildResultRow>();
 
-            if (!JenkinsConsoleUtility.TryGetArgVar(out pat, argsLc, "days"))
+            string daysStr;
+            if (!JenkinsConsoleUtility.TryGetArgVar(out daysStr, argsLc, "days") || !int.TryParse(daysStr, out days))
                 days = 3;
             pat = JenkinsConsoleUtility.GetArgVar(argsLc, "pat");
             string kustoConfigFile = JenkinsConsoleUtility.GetArgVar(argsLc, "kustoConfig");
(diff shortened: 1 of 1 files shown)


JenkinsConsoleUtility/jcuSrc/Commands/GetAdoBuilds.cs
```diff
@@ -11,8 +11,10 @@ namespace JenkinsConsoleUtility.jcuSrc.Commands
 {
     class GetAdoBuilds : ICommand
     {
-        private const string buildsUrl = "https://dev.azure.com/PlayFabInternal/Main/_apis/build/builds?api-version=6.1-preview.6";
-        public static readonly List<string> filterIds = new List<string> { "refs/heads/master", "refs/heads/develop" };
+        private const string expectedFolderPrefix = "\\PlayFabSdk";
+        private const string pipelinesUrl = "https://dev.azure.com/PlayFabInternal/Main/_apis/pipelines";
+        private const string buildsUrl = "https://dev.azure.com/PlayFabInternal/Main/_apis/build/builds?definitions={definitionId}";
+        // public static readonly List<string> filterIds = new List<string> { "refs/heads/master", "refs/heads/develop" };
```
Vicodex, line 17: not sure what you policy about commented code, consider removing if it was left accidentally

Vicodex (approved):

JenkinsConsoleUtility/jcuSrc/Commands/GetAdoBuilds.cs
pgilmorepf: It was not accidental, but it may be a required filter that is added at a later time.
At present it's not necessary, but it could become necessary... maybe in the next couple of months.
I'd like to leave it because those strings aren't inherently obvious and when I need it later, I don't want to have to look it up.

PR #744, opened 2021-02-19
Update UUnitTestCase.cs

toddbellMSFT:
Quickest fix to our bug. Don't label this test as Failed if we see a WANT_READ. That just means we have MORE to read, so we need to re do


pgilmorepf (changes requested):
If the goal is a retry, this won't affect that change. Failed results already retry for C#, so this won't change that.

PR #746, opened 2021-05-12
Tb/remove ad

toddbellMSFT:
Recently we removed something from the service that will always make these API's return false. 

This change will fully remove them from the PlayFab SDK.


targets/LuaSdk/templates/PlayFabApi.lua.ejs
```diff
@@ -11,25 +11,7 @@ local PlayFabSettings = require("<%- requirePrefix %>PlayFabSettings")
 local PlayFab<%- api.name %>Api = {
     settings = PlayFabSettings.settings
 }
-<% if (hasClientOptions) { %>
-function PlayFabClientApi.IsClientLoggedIn()
```
pgilmorepf, line 15: We actually want to keep "IsClientLoggedIn()"

pgilmorepf: Consequently, we'll still need to keep the hasClientOptions bracket as well.

targets/javascript/testingFiles/PlayFabApiTest.ts
```diff
@@ -207,35 +206,6 @@ var PlayFabApiTests = {
         assert.ok(typeof loginPromise.then === "function" && Promise.resolve(loginPromise) === loginPromise, "Testing whether the login request returned a promise object");
     },
 
-    /* CLIENT API
-     * Test that the login call sequence sends the AdvertisingId when set
-     */
-    LoginWithAdvertisingId: function (assert): void {
```
pgilmorepf, line 213: Javascript doesn't run automated tests, so you'll have to load the test-web-page manually. Please do that.

targets/objc/templates/PlayFabAPI.m.ejs
```diff
@@ -36,29 +36,6 @@ return deviceModel;
     return !([[PlayFab<%- api.name -%>API GetInstance].mUserSessionTicket length]==0);
 }
 
-#ifdef USE_IDFA
```
pgilmorepf, line 39: USE_IDFA
Do a global search for this, just in case, this is a new variable that might have further use elsewhere.

pgilmorepf: ("New" as in I haven't seen it as part of this changelog until now)

targets/unity-v2/source/ExampleTestProject/Assets/PlayFabSDK/Client/PlayFabDeviceUtil.cs
```diff
@@ -118,10 +91,7 @@ private static void _OnPlayFabLogin(ClientModels.UserSettings settingsForUser, s
             }
 
             // Device attribution (adid or idfa)
-            if (settings.AdvertisingIdType != null && settings.AdvertisingIdValue != null)
-                DoAttributeInstall(settings, instanceApi);
-            else
-                GetAdvertIdFromUnity(settings, instanceApi);
+            GetAdvertIdFromUnity(settings, instanceApi);
```
pgilmorepf, line 94: I'm 99% certain there's no need to call GetAdvertIdFromUnity anymore. It's the same system.
Remove the comment about Device attribution too.

targets/unity-v2/source/ExampleTestProject/Assets/PlayFabSDK/Client/PlayFabDeviceUtil.cs
```diff
@@ -135,27 +105,6 @@ private static void _OnPlayFabLogin(ClientModels.UserSettings settingsForUser, s
             {
                 settings.DisableFocusTimeCollection = true;
             }
-#endif
-        }
-
-        private static void GetAdvertIdFromUnity(PlayFabApiSettings settings, IPlayFabInstanceApi instanceApi)
```
pgilmorepf, line 141: Oh weird... you removed this method, and yet didn't remove the call to it above... Check why this still compiles. It may be #defined out, in which case, you need to enable it to test the removal.

pgilmorepf (changes requested):
Approve:
Doing all the changes en-masse all at once, because they're the same change and it's a lot faster to find it all that way.
The removals themselves look very good.

Request changes:
Lua: Restore the IsClientLoggedIn() method you accidentally deleted.
Global: Recheck for more instances of USE_IDFA variable. I think there are more.
Unity: You deleted the GetAdvertIdFromUnity method, but not a call to it. Find out why this didn't cause a compiler issue.
Global: It would be nice if each SDK were a separate PR. It's a "just in case" scenario where if one of the SDKs is broken by this change, we might only revert just one SDK rather than all of them. It's also "good form" in general to break up the PR's.

pgilmorepf (approved):

PR #749, opened 2021-05-18
Update PlayFabApi.lua.ejs

toddbellMSFT:
Missed an Opening curly bracket.

### DIFF
--- a/targets/LuaSdk/SharedTesting/PlayFabTesting/PlayFabApiTestSuite.lua
+++ b/targets/LuaSdk/SharedTesting/PlayFabTesting/PlayFabApiTestSuite.lua
@@ -120,7 +120,7 @@ function PlayFabApiTestSuite.LoginOrRegisterTest()
     PlayFabClientApi.LoginWithCustomID(loginRequest, AsyncTestSuite.WrapCallback("OnLoginSuccess", PlayFabApiTestSuite.OnLoginSuccess), PlayFabApiTestSuite.OnSharedError)
 end
 function PlayFabApiTestSuite.OnLoginSuccess(result)
-    if (result.PlayFabId) then
+    if (result.PlayFabId) then 
         PlayFabApiTestSuite.playFabId = result.PlayFabId
         AsyncTestSuite.EndTest("PASSED", PlayFabApiTestSuite.playFabId)
     else
--- a/targets/LuaSdk/SharedTesting/PlayFabTesting/PlayFabApiTestSuite.lua
+++ b/targets/LuaSdk/SharedTesting/PlayFabTesting/PlayFabApiTestSuite.lua
@@ -121,6 +121,7 @@ function PlayFabApiTestSuite.LoginOrRegisterTest()
 end
 function PlayFabApiTestSuite.OnLoginSuccess(result)
     if (result.PlayFabId) then 
+        print("Our External On Login Success WAS called!")
         PlayFabApiTestSuite.playFabId = result.PlayFabId
         AsyncTestSuite.EndTest("PASSED", PlayFabApiTestSuite.playFabId)
     else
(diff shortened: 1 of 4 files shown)


targets/LuaSdk/templates/PlayFabApi.lua.ejs
```diff
@@ -11,9 +11,9 @@ local PlayFabSettings = require("<%- requirePrefix %>PlayFabSettings")
 local PlayFab<%- api.name %>Api = {
     settings = PlayFabSettings.settings
 }
-<% if (hasClientOptions) %>
+<% if (hasClientOptions) { %>
 function PlayFabClientApi.IsClientLoggedIn()
-    return (not (PlayFabSettings._internalSettings.sessionTicket == nil))
```
pgilmorepf, line 16: null isn't a thing in Lua. It's nil.

targets/LuaSdk/SharedTesting/PlayFabTesting/PlayFabApiTestSuite.lua
```diff
@@ -348,7 +349,10 @@ function PlayFabApiTestSuite.OnGetEntityToken(result)
     if (result.Entity) then 
         PlayFabApiTestSuite.entityId = result.Entity.Id
         PlayFabApiTestSuite.entityType = result.Entity.Type
-        AsyncTestSuite.EndTest("PASSED", result.Entity.Id)
+        if (result.EntityToken) then
+            PlayFabSettings._internalSettings.EntityToken = result.EntityToken.EntityToken
```
pgilmorepf, line 353: This is probably a case where Lua SDK is so old that it's missing code that has been updated in all other SDKs.
Please remove this EntityToken assignment here, and apply this assignment somewhere appropriate in the PlayFabApi.lua.ejs file. Doing it here means it's only set once for this test, and not properly set "in general" which really defeats the entire point of this test.

targets/LuaSdk/SharedTesting/PlayFabTesting/PlayFabApiTestSuite.lua
```diff
@@ -122,6 +122,7 @@ end
 function PlayFabApiTestSuite.OnLoginSuccess(result)
     if (result.PlayFabId) then 
         PlayFabApiTestSuite.playFabId = result.PlayFabId
+        PlayFabSettings._internalSettings.sessionTicket = result.SessionTicket
```
pgilmorepf, line 125: Same deal here as with entityToken.
Please remove this sessionTicket assignment here, and apply this assignment somewhere appropriate in the PlayFabApi.lua.ejs file. Doing it here means it's only set once for this test, and not properly set "in general" which really defeats the entire point of this test.

targets/LuaSdk/make.js
```diff
@@ -178,7 +178,9 @@ function getResultAction(tabbing, apiCall) {
         preCallback = internalTabbing + "PlayFabSettings._internalSettings.entityToken = result.EntityToken\n";
     else if (apiCall.result === "LoginResult") {
         preCallback = internalTabbing + "PlayFabSettings._internalSettings.sessionTicket = result.SessionTicket\n"
-            + internalTabbing + "if (result.EntityToken) then PlayFabSettings._internalSettings.entityToken = result.EntityToken.EntityToken end\n";
+            + internalTabbing + "if (result.EntityToken) then \n"
```
pgilmorepf, line 181: There's no point of checking result.EntityToken anymore. Once upon a time, EntityToken had to be specifically requested and was sometimes returned as null. Now it's always returned, so let's remove the if-check.

pgilmorepf (changes requested):

pgilmorepf (approved):

PR #752, opened 2021-06-03
Add retryafter value

DrewRomanyk:
Adds the retryafter value for the client to know when to retry a 429 HTTP request after a specified amount of seconds

### DIFF
--- a/targets/csharp/source/PlayFabSDK/source/PlayFabHttp/PlayFabHttp.cs
+++ b/targets/csharp/source/PlayFabSDK/source/PlayFabHttp/PlayFabHttp.cs
@@ -35,6 +35,7 @@ namespace PlayFab.Internal
         public int errorCode;
         public string errorMessage;
         public Dictionary<string, string[]> errorDetails = null;
+        public uint? retryAfterSeconds = null;
     }
 
     public class PlayFabJsonSuccess<TResult> where TResult : PlayFabResultCommon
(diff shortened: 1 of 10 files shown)


targets/csharp/source/PlayFabSDK/source/PlayFabHttp/PlayFabHttp.cs
```diff
@@ -35,6 +35,7 @@ public class PlayFabJsonError
         public int errorCode;
         public string errorMessage;
         public Dictionary<string, string[]> errorDetails = null;
+        public uint? retryAfterSeconds = null;
```
toddbellMSFT, line 38: An issue popped up on line 409 of this file in our Unit tests from the nullable uint here:

"Assets\PlayFabSDK\Shared\Internal\PlayFabHttp\PlayFabHTTP.cs(409,37): error CS0173: Type of conditional expression cannot be determined because there is no implicit conversion between 'uint' and '<null>'"

targets/java/source_shared/src/main/java/com/playfab/internal/PlayFabHTTP.java
```diff
@@ -101,14 +101,15 @@ public static String receive(InputStream in) throws IOException {
         return recieved.toString();
     }
 
-    public static PlayFabError GeneratePfError(int httpCode, PlayFabErrorCode pfErrorCode, String errorMessage, Map<String, List<String>> errorDetails) {
+    public static PlayFabError GeneratePfError(int httpCode, PlayFabErrorCode pfErrorCode, String errorMessage, Map<String, List<String>> errorDetails, Integer retryAfterSeconds) {
```
toddbellMSFT, line 104: Will this create an implicit cast between uint? and Integer?

DrewRomanyk: For Java I left the fields as an Integer, as that is how that lang handles nullable numbers

toddbellMSFT (approved):

PR #755, opened 2021-08-13
remove windows phone target

toddbellMSFT:
Unity 2021 errors out when we load the PlayFab Unity SDK, complaining about the two lines we are omitting in future Unity versions


pgilmorepf (approved):
Approved with caveat:
It'd probably be better to just drop WP8 support entirely at this point, there's no compelling reason to keep it.

PR #762, opened 2021-11-12
CSharp Connection String implementation

toddbellMSFT:


targets/csharp/source/PlayFabSDK/UnittestRunner/UnittestRunner.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>netcoreapp1.1;netcoreapp2.0;net45</TargetFrameworks>
```
amccalib, line 5: Are the netcoreapp additions intentional?

targets/csharp/source/PlayFabSDK/source/PlayFabSDK.csproj.ejs
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;netstandard1.1;net45;portable45-net45+win8+wpa81</TargetFrameworks>
```
amccalib, line 4: Same as above. Are these changes intentional?

targets/csharp/source/PlayFabSDK/source/PlayFabSettings.cs.ejs
```diff
@@ -37,6 +37,8 @@ namespace PlayFab
         public static string TitleId { get { return staticSettings.TitleId; } set { staticSettings.TitleId = value; } }
         #endregion Deprecated staticSettingsredirect properties
 
+        public static string ConnectionString { get { return staticSettings.ConnectionString; } set { staticSettings.ConnectionString = value; } }
```
amccalib, line 40: Given the Obsolete attributes above, should we be adding this?  It sounds like we developers should just be referencing staticSettings instead of this property directly.

amccalib (changes requested):

targets/csharp/source/PlayFabSDK/UnittestRunner/UnittestRunner.csproj
toddbellMSFT: Yup, netcoreapp 1.1 and 2.0 apparently have hit end of life :( 

https://dotnet.microsoft.com/download/dotnet/2.0/runtime?utm_source=getdotnetcore&utm_medium=referral 

I got a hang when running 1.1, and 2.0 spits our this url that mentions this is no longer supported by microsoft.

targets/csharp/source/PlayFabSDK/source/PlayFabSDK.csproj.ejs
toddbellMSFT: portable45 had an issue, and so did the win8 stuff. I put back netstandard 2.0 and 1.1 in this project file and it still ran fine without issues so I can add those back (netcore seems to be what is end of life, standard seems to still be around)

amccalib (approved):

PR #766, opened 2021-11-25
unity connection string

toddbellMSFT:


targets/unity-v2/source/ExampleTestProject/Assets/PlayFabEditorExtensions/Editor/Scripts/PlayFabEditorSDK/PlayFabEditorHttp.cs
```diff
@@ -191,10 +191,18 @@ private static IEnumerator PostDownload(UnityWebRequest www, Action<byte[]> call
             {
                 yield return www.SendWebRequest();
 
+#if UNITY_2021_OR_NEWER
```
amccalib, line 194: Could you explain what this is for? Why the addition of new logic for Unity 2021?

amccalib (approved):

targets/unity-v2/source/ExampleTestProject/Assets/PlayFabEditorExtensions/Editor/Scripts/PlayFabEditorSDK/PlayFabEditorHttp.cs
toddbellMSFT: A warning popped up when I was implementing the connection string which was annoying to debug when running the tests. This should also fix a warning that may become an error in later versions of Unity.

PR #769, opened 2022-01-05
Fix the header and url issue in Postman V2

JayZuo:
Although Postman Schema said the header can be string, but when we use string, the Postman app (I'm using Postman v9.7.1) cannot recognize the header correctly. So, I added a getPostmanHeaderV2 method to set the header to "Header List".

Currently, we are using "apiName" and "apiCall.name" to set up the path of the current url. However, for some APIs, the first segment of the path is not the "apiName". Using "GetDraftItem" for example, it's "apiName" is "economy", however, the correct segment should be "catalog". As the Url can be a string, I'm just using the literal request URL instead of the complete broken-down object.

### DIFF
--- a/targets/postman/make.js
+++ b/targets/postman/make.js
@@ -24,6 +24,7 @@ exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
         apis: apis,
         getPostmanDescription: getPostmanDescription,
         getPostmanHeader: getPostmanHeader,
+        getPostmanHeaderV2: getPostmanHeaderV2,
         getRequestExample: getRequestExample,
         getUrl: getUrl,
         getVerticalTag: getVerticalTag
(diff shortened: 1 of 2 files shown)


targets/postman/make.js
```diff
@@ -84,6 +85,86 @@ function getPostmanHeader(apiCall) {
...
+                "key": "X-PlayFabSDK",
+                "value": "PostmanCollection-" + sdkGlobals.sdkVersion
+            },
+            {
+                "key": "Content-Type",
+                "value": "application/json"
+            },
+            {
+                "key": "X-Authorization",
+                "value": "{{SessionTicket}}"
```
Jelisacooper4, line 101: L

Jelisacooper4 (approved):

toddbellMSFT (approved):

PR #778, opened 2022-02-18
Changes required to generate a CSharp Azure SDK

amccalib:
### Description ###

Making the following changes to support generating a C# SDK for Azure

1. Adding a flag to indicate that we're generating an Azure SDK. This lets us use the same templates. Currently this is driven by the build flags. This also necessitated exposing the build flags to make.js.
2. Many API groups are now empty, so added a check and won't generate empty API and Data Model files.
3. There is a new login API that needs special handling (auto-insert some request fields, cache auth result)
4. Multiple changes to account for the new PlayerAccountPoolId. This includes adding to PlayFabApiSettings, automatically adding it to auth requests, and creating a new client side error for when it's not present.
5. By default, the preprocessor flags are set/cleared such as to remove client, server, and admin api groups.
6. Fixed multiple bugs that arose from specific client apis not being present (generally by reading the appropriate preprocessor flag).


targets/csharp/make.js
```diff
@@ -25,18 +25,28 @@ exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
...
         sdkVersion: sdkGlobals.sdkVersion,
-        sdkYear: sdkGlobals.sdkVersion.split(".")[2].substr(0, 2)
+        sdkYear: sdkGlobals.sdkVersion.split(".")[2].substr(0, 2),
+        azureSdk: false
     };
 
+    if (sdkGlobals.buildFlags.includes("azure")) {
+        locals.azureSdk = true,
+        locals.hasServerOptions = false,
+        locals.extraDefines = ";NETFX_CORE;SIMPLE_JSON_TYPEINFO;DISABLE_PLAYFABCLIENT_API"
```
pgilmorepf, line 35: Nit:
;NETFX_CORE;SIMPLE_JSON_TYPEINFO; could be in some constant at the top, since it's reused now.

targets/csharp/make.js
```diff
@@ -50,7 +60,7 @@ function getBaseTypeSyntax(datatype) {
     //begin classes - only 1 possible
     if (datatype.className.toLowerCase().endsWith("request"))
         parents.push("PlayFabRequestCommon");
-    else if (datatype.className.toLowerCase() === "loginresult" || datatype.className === "RegisterPlayFabUserResult")
+    else if (datatype.className.toLowerCase() === "loginresult" || datatype.className === "AuthenticateIdentityResult" || datatype.className === "RegisterPlayFabUserResult")
```
pgilmorepf, line 63: Nit:
The "toLowerCase" on LoginResult could probably be removed. I think I thought at the time that it'd be more safe, but so many years later, I think it becomes less safe. The chance of names that legitimately differ only by capitalization is probably expanding. And no other part of the pipeline would be limited in this way, making THIS the [only?] point of failure. (And anything else that uses toLowerCase() in this way.)

targets/csharp/make.js
```diff
@@ -92,12 +102,14 @@ function makeDatatypes(apis, sourceDir, apiOutputDir) {
     };
 
     for (var a = 0; a < apis.length; a++) {
-        var modelsLocal = {
-            api: apis[a],
-            makeDatatype: makeDatatype
-        };
+        if (apis[a].calls.length > 0) {
```
pgilmorepf, line 105: Nit:
C++ is the only language that optimizes for diving into if statements, rather than trying to bail out early.

Since this is JS, it would be more stylistically appropriate to invert the if, and continue, without tab-shifting the other lines in the for-block.

pgilmorepf (commented):
I have only nitpicks. Looks good so far!

AlonsoMondal (approved):

PR #779, opened 2022-02-24
Update playfabV2.json.ejs

toddbellMSFT:
Explicitly label body as application/json, as Postman defaults to text now.


toddbellMSFT:
Postman in v2 started interpreting "raw" as just text, which was fine for v1, but making it explicit (even though it means we are saying in both the header and "body" that the Content-Type is application/json

pgilmorepf (approved):

PR #798, opened 2022-05-12
Tb/generate fix

toddbellMSFT:


AlonsoMondal (approved):

generate.ts
```diff
@@ -732,13 +732,40 @@ function templatizeTree(locals: { [key: string]: any }, sourcePath: string, dest
     for (var i = 0; i < filesInDir.length; i++) {
         var filename = filesInDir[i];
         var file = sourcePath + "/" + filename;
+
+        var folderExcluded = false;
```
amccalib, line 736: I favor moving these into the isDirectory if/else below. That way, we only run the excludedFolders check on Directories and the excludedFiles check on files.

generate.ts
```diff
@@ -732,13 +732,40 @@ function templatizeTree(locals: { [key: string]: any }, sourcePath: string, dest
...
+                if(excludeFiles[excludedFileIndex] == filename)
+                {
+                    fileExcluded = true;
+                    break;
+                }
+            }
+        }
+
         if (fs.lstatSync(file).isDirectory()) {
-            if (excludeFolders != null && excludeFolders.includes(filename))
```
amccalib, line 736: Do you know why .includes() wasn't working? Some sort of difference in the execution environment?

amccalib (changes requested):

generate.ts
toddbellMSFT: I think it's a mix of two issues: 1.) the build machines for certain private templates aren't recognizing the includes 2.) since javascript is interpreted, most of the other languages won't hit this issue since excludeFolders is null for most typical cases. 

So one alternative solution is tracking down why the includes javascript function does not work as expected on our build agents (I just updated Node JS and that didn't seem to do the trick).

generate.ts
```diff
@@ -732,13 +732,38 @@ function templatizeTree(locals: { [key: string]: any }, sourcePath: string, dest
...
+                for(var excludedFolderIndex = 0; excludedFolderIndex < excludeFolders.length; excludedFolderIndex++)
+                {
+                    if(excludeFolders[excludedFolderIndex] == filename)
+                    {
+                        folderExcluded = true;
+                        break;
+                    }
+                }
+            }
+            if (excludeFolders != null && folderExcluded)
```
amccalib, line 749: nit: This first clause (excludeFolders != null) can be removed. It's checked above.

generate.ts
```diff
@@ -732,13 +732,38 @@ function templatizeTree(locals: { [key: string]: any }, sourcePath: string, dest
...
+                for(var excludedFileIndex = 0; excludedFileIndex < excludeFiles.length; excludedFileIndex++)
+                {
+                    if(excludeFiles[excludedFileIndex] == filename)
+                    {
+                        fileExcluded = true;
+                        break;
+                    }
+                }
+            }
+            if (excludeFiles != null && fileExcluded)
```
amccalib, line 766: nit: same as above

amccalib (approved):

PR #802, opened 2022-06-06
Adding aadAuth header to postman make file

toddbellMSFT:
We didn't have a valid json output when we return a simple empty string. 

I added a case for aad if we need to do anything specific. 

Otherwise, this new throw is here to ensure we get a quicker fix in the future.

### DIFF
--- a/targets/postman/make.js
+++ b/targets/postman/make.js
@@ -159,6 +159,10 @@ function getPostmanHeaderV2(apiCall) {
             {
                 "key": "Content-Type",
                 "value": "application/json"
+            },
+            {
+                "key": "X-AadToken",
+                "value": "{{AadToken}}"
             }
         ])
     else if (apiCall.auth === "None")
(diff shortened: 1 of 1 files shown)


targets/postman/make.js
```diff
@@ -150,6 +150,17 @@ function getPostmanHeaderV2(apiCall) {
                 "value": "{{EntityToken}}"
             }
         ])
+    else if (apiCall.auth === "AadToken")
```
amccalib, line 153: Was this a new API we were hitting this on? Can you share more details?

toddbellMSFT: 'SetTitleInternalDataAndOverrides', apparently has a .auth set to AadToken, so we may need to also add another group of 

"key": "X-AadToken"
"value": "{{AadToken}}"

But I just want to confirm that it would then require folks to add an AadToken manually to their postman collection

PR #808, opened 2022-06-30
nuget dependency updates

toddbellMSFT:
The latest Newtonsoft json update also relies on other dependencies being updated (and us hard coding it in the past means we will sync to the older version).

### DIFF
--- a/JenkinsConsoleUtility/JenkinsConsoleUtility.csproj
+++ b/JenkinsConsoleUtility/JenkinsConsoleUtility.csproj
@@ -13,6 +13,8 @@
     <FileAlignment>512</FileAlignment>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
     <TargetFrameworkProfile />
+    <NuGetPackageImportStamp>
+    </NuGetPackageImportStamp>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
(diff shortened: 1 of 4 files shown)


JenkinsConsoleUtility/jcuSrc/JUnitXml.cs
```diff
@@ -82,10 +82,14 @@ private static void ParseElementStart(XmlReader reader, bool isEmptyElement)
                     {
                         name = reader.GetAttribute("name")
                     };
-                    int.TryParse(reader.GetAttribute("errors"), out _curSuiteReport.errors);
-                    int.TryParse(reader.GetAttribute("tests"), out _curSuiteReport.tests);
-                    int.TryParse(reader.GetAttribute("failures"), out _curSuiteReport.failures);
-                    int.TryParse(reader.GetAttribute("skipped"), out _curSuiteReport.skipped);
+
+                    int parsedTests = 0;
```
amccalib, line 86: Can you explain the change to the TryParse calls here? Why the change to local variable and why get rid of the call for "errors"?

amccalib (changes requested):

JenkinsConsoleUtility/jcuSrc/JUnitXml.cs
toddbellMSFT: errors no longer exists in the public PlayFabAllSDK UUtil class (unit testing class, so only tests are affected, no end product)

amccalib (approved):

PR #814, opened 2022-07-13
ConnectionString - removing in classic - changing to endpoint in Azure

AlonsoMondal:
Removing connectionString on classic SDKs and changing it to endpoint for Azure SDKs


ExampleTestProject/Assets/Testing/Tests/Client/EndpointTests.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace PlayFab.UUnit
 {
-    public class ConnectionStringTests : UUnitTestCase
+    public class Endpoint : UUnitTestCase
```
toddbellMSFT, line 10: I think this should have a "-Tests" suffix to visually differentiate from the new Endpoint rename

ExampleTestProject/Assets/Testing/Tests/Shared/TestTitleDataLoader.cs
```diff
@@ -102,7 +102,7 @@ private static void SetTitleId(bool setPlayFabSettings)
 #if ENABLE_PLAYFABSERVER_API || ENABLE_PLAYFABADMIN_API || UNITY_EDITOR || ENABLE_PLAYFAB_SECRETKEY
             PlayFabSettings.DeveloperSecretKey = _loadedData.developerSecretKey;
 #endif
-            PlayFabSettings.ConnectionString = _loadedData.connectionString;
```
toddbellMSFT, line 105: Will we not need to add the Endpoint equivalent here?

targets/csharp/source/PlayFabSDK/source/PlayFabApiSettings.cs.ejs
```diff
@@ -21,11 +21,13 @@ namespace PlayFab
         /// <summary> You must set this value for PlayFabSdk to work properly (Found in the Game Manager for your title, at the PlayFab Website) </summary>
         public string TitleId;
 <% if (azureSdk) { %>
+    
         /// <summary> You must set this value in order to call Player Authentication APIs (Found in the Azure Portal as a property of the Player Account Pool resource) </summary>
         public string PlayerAccountPoolId;
-<% } %>
+
         /// <summary> Connects directly to a PlayFab title resource. Terminology subject to change </summary>
```
toddbellMSFT, line 28: We may want to update the comment above to reflect this finalized Terminology

toddbellMSFT (approved):

ExampleTestProject/Assets/Testing/Tests/Shared/TestTitleDataLoader.cs
AlonsoMondal: This file was added by mistake. Checkout the new template in targets/unity-v2/templates/TestTitleDataLoader.cs

toddbellMSFT (approved):

PR #816, opened 2022-07-22
Task 40362273: Generate PFCore SDK within bumblion repository

tculotta:
This is also making a minor change in the next repo to move the template folder from src/playfabcorecsdk_template to src/playfabcorec/playfabcorecsdk_template, so the template and generated sdk will be side by side in src/playfabcorec. If there's any issues with this, lets finalize that in this pr

### DIFF
--- a/SDKGenerator.njsproj
+++ b/SDKGenerator.njsproj
@@ -101,6 +101,11 @@
     <OutputPath>bin\</OutputPath>
     <ScriptArguments>-destpath ..\sdks\XPlatCppSdk -apiSpecGitUrl -buildIdentifier XPlatCppSdk_manual</ScriptArguments>
   </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)' == 'XPlatCSdk' ">
+    <DebugSymbols>true</DebugSymbols>
+    <OutputPath>bin\XPlatCSdk\</OutputPath>
+    <ScriptArguments>-destpath ..\..\src\playfabcorec\playfabcorecsdk -apiSpecGitUrl -buildIdentifier playfabcorecsdk_manual</ScriptArguments>
+  </PropertyGroup>
   <ItemGroup>
     <Content Include="README.md" />
     <Content Include="targets\csharp\make.js" />
--- a/SDKGenerator.njsproj
+++ b/SDKGenerator.njsproj
@@ -104,7 +104,7 @@
   <PropertyGroup Condition=" '$(Configuration)' == 'XPlatCSdk' ">
     <DebugSymbols>true</DebugSymbols>
     <OutputPath>bin\XPlatCSdk\</OutputPath>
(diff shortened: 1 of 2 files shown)


SDKGenerator.njsproj
```diff
@@ -101,6 +101,11 @@
     <OutputPath>bin\</OutputPath>
     <ScriptArguments>-destpath ..\sdks\XPlatCppSdk -apiSpecGitUrl -buildIdentifier XPlatCppSdk_manual</ScriptArguments>
   </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)' == 'XPlatCSdk' ">
+    <DebugSymbols>true</DebugSymbols>
+    <OutputPath>bin\XPlatCSdk\</OutputPath>
+    <ScriptArguments>-destpath ..\..\src\playfabcorec\playfabcorecsdk -apiSpecGitUrl -buildIdentifier playfabcorecsdk_manual</ScriptArguments>
```
jplafonta, line 107: >..\..\src\playfabcorec\playfabcorecsdk [](http://example.com/codeflow?start=31&length=38)

This seems like a strange thing to hardcode into the build config, but I see that its been done that way for the other configurations as well.  Is there a way to override this at build time?

I'd prefer the directory structure to be something like this:

src\PlayFabCore\
  PlayFabCoreCSdk
  PlayFabCoreCSdk_Template

I think its a bit more readable and matches what we've got for the github repo names

jplafonta, line 107: >playfabcorecsdk_manual [](http://example.com/codeflow?start=102&length=22)

Similarly I think this should be PlayFabCoreCSdk_manual because I find it to be more readble

jplafonta (changes requested):
:clock1:

SDKGenerator.njsproj
tculotta: Yeah the path is kinda gross. That's why I was wondering earlier if we'd already had some system to hook down into here given it's a submodule. I'm not sure what you're thinking of with "override this at build time". If we can instead write some kind of build script at the template level that procs SDKGenerator and passes in all the needed paths, that would be nice. It would probably completely remove the need for any change here.

I've got no problem with changing those directory names.

tculotta: To be clear, I intended to explore that build script option a bit before I realized the hardcoded path is apparently the standard design here.

jplafonta: There are some build scripts in /SDKBuildScripts/

Maybe would be a good idea to add one for Core C SDK there (in addition to adding this VS build config probably)

tculotta: Changed folder names. Build script is being modified at the template level as discussed in teams.

jplafonta (approved):
:shipit:

PR #820, opened 2022-08-12
Update ejs

toddbellMSFT:
ejs 1.0 had the function ejs.compile that we relied on to take a template file, return a javascript function that would then be called with a locals JSON object. 

The latest version of ejs does not have this function anymore. Rather, I only found ejs.render to work to our purposes of taking in a readFile string of our .ejs template files and spit back out a properly aligned text file with the expected output of the locals JSON object. 

This new update covers (in order of how I did them): Postman, C#, Unity, Java, JavaScript, Node.

C++ and Unreal are in privateTemplates so will require a different PR.


amccalib:
Are there any changes in output for each of the supported SDKs with these updates? I would expect either no changes or just formatting, maybe. Please confirm.

packages.config
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Microsoft.TypeScript.MSBuild" version="4.7.4" targetFramework="net40" developmentDependency="true" />
```
amccalib, line 3: Do we need to be including the full nuget package in the repo? Can't the build resolve package dependencies?

targets/csharp/make.js
```diff
@@ -1,10 +1,11 @@
+let ejs = require('ejs');
 var path = require("path");
 
 // Making resharper less noisy - These are defined in Generate.js
 if (typeof (copyOrTemplatizeFile) === "undefined") copyOrTemplatizeFile = function () { };
 if (typeof (templatizeTree) === "undefined") templatizeTree = function () { };
 if (typeof (generateApiSummaryLines) === "undefined") generateApiSummaryLines = function () { };
-if (typeof (getCompiledTemplate) === "undefined") getCompiledTemplate = function () { };
+//if (typeof (getCompiledTemplate) === "undefined") getCompiledTemplate = function () { };
```
amccalib, line 8: Would be good to remove commented lines before finalizing.

amccalib, line 8: Is there any point in leaving getCompiledTemplate in Generate.js? Can it be removed?

targets/postman/make.js
```diff
@@ -30,27 +32,38 @@ exports.makeCombinedAPI = function (apis, sourceDir, apiOutputDir) {
...
     try {
-        require(outputFile2); // Read the destination file and make sure it is correctly formatted json
+        require(outputFile2Path); // Read the destination file and make sure it is correctly formatted json
     } catch (ex) {
-        throw "The Postman Collection output was not properly formatted JSON:\n" + outputFile2;
+        throw "The Postman Collection output was not properly formatted JSON:\n" + outputFile2Path;
     }
 }
 
+function testMe() {
```
amccalib, line 63: Needs to be removed?

amccalib (changes requested):

targets/csharp/make.js
toddbellMSFT: Not easily. The function is called by copyOrTemplatizeFile. It's still used for caching files for determining if we have "templatized" (remaining feature: visited) the file or not. What is returned is overwritten with our new changes from the make.js file

toddbellMSFT:
> 

Regarding end result updates, I have found no differences in output of the new ejs files and the older generation:

C# generation examples:
Old ejs 1.0: https://github.com/PlayFab/CSharpSDK/tree/tb/clean2 
With new make.js update: https://github.com/PlayFab/CSharpSDK/tree/tb/new-ejs-gen2 

So we have obsolete tag updates and a local generation id change in both is expected.

Unity generation example: 
Old ejs 1.0:  https://github.com/PlayFab/UnitySDK/tree/tb/cleanse 
New make.js: https://github.com/PlayFab/UnitySDK/tree/tb/ejs-update 
Looks similar to C#, settings file changes mostly
Java:
Old ejs 1.0: https://github.com/PlayFab/JavaSDK/tree/tb/clean 
New make.js: https://github.com/PlayFab/JavaSDK/tree/tb/new-ejs 

Looks similar to C#, settings file updates. Settings files mostly

JavaScript:
Old ejs 1.0: https://github.com/PlayFab/JavaScriptSDK/tree/tb/cleanse 
New make.js:  https://github.com/PlayFab/JavaScriptSDK/tree/tb/new-generationlogic 
No changes between the two

Node:
Old ejs 1.0: https://github.com/PlayFab/NodeSDK/compare/tb/clean?expand=1
New make.js: https://github.com/PlayFab/NodeSDK/compare/tb/ejs-update?expand=1 
No changes between the two.

packages.config
toddbellMSFT: I like the idea of adding the npm install ejs to the build itself so we don't need to copy a dependency module within our own repo. Let me see what we can modify.

amccalib: Would this impact local builds? Don't we still need the package config file to inform my local build of the needed packages to restore?

toddbellMSFT: Looks like we can safely remove it and local building works fine. the solution file wants to run a nuget restore for that typescript package. But it shouldn't be required for our automation. Just for devs who prefer to run in Visual Studio rather than command line.

targets/javascript/make.js
```diff
@@ -1,17 +1,21 @@
+var ejs = require('ejs');
 var path = require("path");
 
 // Making resharper less noisy - These are defined in Generate.js
 if (typeof (templatizeTree) === "undefined") templatizeTree = function () { };
 if (typeof (generateApiSummaryLines) === "undefined") generateApiSummaryLines = function () { };
-if (typeof (getCompiledTemplate) === "undefined") getCompiledTemplate = function () { };
+// if (typeof (getCompiledTemplate) === "undefined") getCompiledTemplate = function () { };
```
amccalib, line 7: Looks like there's a few more commented old lines to remove.

amccalib (approved):

PR #821, opened 2022-08-31
Remove node_modules directory

toddbellMSFT:
ensures we aren't carrying around an older version of ejs.

Removes current node_modules folder

adds scripts to install the latest version of ejs 

(Note: all pipelines may need to call this new initejs script first now before calling any generating code). 

Also fixes some internal build issues in generate.js that prevented us from upgrading.


SDKBuildScripts/initejs.bat
```diff
@@ -0,0 +1,3 @@
+pushd ..
+npm install ejs
```
rgomez391, line 2: In order for the npm install to work you'll need a _package.json_ file defining the minimum ejs version we'll be using, and ideally a _package-lock.json_ file (which will be autogenerated) that will define what's the version we'll keep using moving forward. They both should exist on the root folder. 

Example of a valid _package.json_ file which will install any version of EJS greater than 2.5.7:

```
{
  "dependencies": {
    "ejs": "^2.5.7"
  }
}
```

Or... maybe I'm missing something on the file structure 😅 Is the _initejs.bat_  scipt working locally for you right now?

toddbellMSFT: Added

rgomez391 (approved):

PR #827, opened 2022-09-23
Changing CSharp SDK target to .Net Standard 2.0

amccalib:
This should restore Xamarin and Unity support for the CSharp SDK. As far as I can tell, there are no side effects of this change.


amccalib:
Successful pipeline run: https://dev.azure.com/PlayFabInternal/Main/_build/results?buildId=155392&view=results

AlonsoMondal (approved):

PR #829, opened 2022-10-10
Auth GameServer EntityToken caching & deleting

AlonsoMondal:
This PR goes over the changes on each platform to cache the result of AuthenticateGameServerWithCustomId in a seperate "game server entity token".

Request actions for this specific function should be reviewed, not sure if they should have the same options as GetEntityToken but changes are as if they are. Might want to skip the default EntityToken for the new "GameServerEntityToken".

Result action in caching sometimes grabs the default EntityToken. Might want to default that to the new "GameServerEntityToken" too.

### DIFF
--- a/targets/LuaSdk/make.js
+++ b/targets/LuaSdk/make.js
@@ -143,7 +143,7 @@ function getVerticalNameDefault() {
 function getRequestActions(tabbing, apiCall) {
     var requestAction = "";
 
-    if (apiCall.url === "/Authentication/GetEntityToken")
+    if (apiCall.url === "/Authentication/GetEntityToken" || apiCall.url === "/GameServerIdentity/AuthenticateGameServerWithCustomId")
         requestAction = tabbing + "local authKey = nil\n"
             + tabbing + "local authValue = nil\n"
             + tabbing + "if (PlayFabSettings._internalSettings.entityToken) then\n"
@@ -176,6 +176,10 @@ function getResultAction(tabbing, apiCall) {
     var internalTabbing = tabbing + "    ";
     if (apiCall.url === "/Authentication/GetEntityToken")
         preCallback = internalTabbing + "PlayFabSettings._internalSettings.entityToken = result.EntityToken\n";
+    if (apiCall.url === "/GameServerIdentity/AuthenticateGameServerWithCustomId")
+        preCallback = internalTabbing + "PlayFabSettings._internalSettings.gameServerEntityToken = result.EntityToken.EntityToken\n";
+    if (apiCall.url === "/GameServerIdentity/AuthenticateGameServerWithCustomId")
+        preCallback = internalTabbing + "PlayFabSettings._internalSettings.gameServerEntityToken = nil\n";
     else if (apiCall.result === "LoginResult") {
         preCallback = internalTabbing + "PlayFabSettings._internalSettings.sessionTicket = result.SessionTicket\n"
             + internalTabbing + "PlayFabSettings._internalSettings.entityToken = result.EntityToken.EntityToken\n";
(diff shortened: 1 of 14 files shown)


toddbellMSFT (approved):

targets/LuaSdk/make.js
```diff
@@ -143,7 +143,7 @@ function getVerticalNameDefault() {
 function getRequestActions(tabbing, apiCall) {
     var requestAction = "";
 
-    if (apiCall.url === "/Authentication/GetEntityToken")
+    if (apiCall.url === "/Authentication/GetEntityToken" || apiCall.url === "/GameServerIdentity/AuthenticateGameServerWithCustomId")
```
amccalib, line 146: The handling of the AuthenticateGameServerWithCustomId request needs to look a little different from GetEntityToken. The only auth header that is valid is X-EntityToken. The others won't work so I don't think it needs to be included in this "if" statement. The X-EntityToken header should be automatically attached as it is for other Entity APIs

amccalib: This comment applies across multiple other SDKs as well.

targets/LuaSdk/make.js
```diff
@@ -176,6 +176,10 @@ function getResultAction(tabbing, apiCall) {
     var internalTabbing = tabbing + "    ";
     if (apiCall.url === "/Authentication/GetEntityToken")
         preCallback = internalTabbing + "PlayFabSettings._internalSettings.entityToken = result.EntityToken\n";
+    if (apiCall.url === "/GameServerIdentity/AuthenticateGameServerWithCustomId")
```
amccalib, line 179: A couple of issues with this addition:
1. I don't think it makes sense to store the game server token in a separate setting. part of the point of making this change is that, after authenticating, future calls automatically include the new token in the auth header.
2. You clear it immediately after setting it.

targets/LuaSdk/make.js
```diff
@@ -198,7 +202,7 @@ function getResultAction(tabbing, apiCall) {
 }
 
 function getAuthentication(apiCall) {
-    if (apiCall.url === "/Authentication/GetEntityToken")
+    if (apiCall.url === "/Authentication/GetEntityToken" || apiCall.url === "/GameServerIdentity/AuthenticateGameServerWithCustomId")
```
amccalib, line 205: Similar to above, AuthenticateGameServerWithCustomId can be treated as a normal EntityToken auth method for this purpose.

targets/PythonSdk/source/playfab/PlayFabSettings.py.ejs
```diff
@@ -33,6 +33,11 @@ This is automatically populated by the PlayFabAuthenticationApi.GetEntityToken m
 """
 _internalSettings.EntityToken = None
 
+"""
```
amccalib, line 36: As mentioned above, I don't think a new setting makes sense. We should just be storing the game server entity token in the existing EntityToken setting.

amccalib (changes requested):
Two main sets of changes that need to be made:
1. No special auth logic is needed for the new API
2. The result should be cached in the existing EntityToken setting.

targets/PythonSdk/make.js
```diff
@@ -192,7 +192,7 @@ function getAuthParams(apiCall) {
 }
 
 function getRequestActions(tabbing, apiCall) {
-    if (apiCall.result === "LoginResult" || apiCall.request === "RegisterPlayFabUserRequest")
```
amccalib, line 195: Looks like you removed the wrong clause.

targets/PythonSdk/make.js
```diff
@@ -205,7 +205,7 @@ function getRequestActions(tabbing, apiCall) {
     if (apiCall.auth === "SecretKey")
         return tabbing + "if not PlayFabSettings.DeveloperSecretKey:\n"
             + tabbing + "    raise PlayFabErrors.PlayFabException(\"Must have DeveloperSecretKey set to call this method\")\n\n";
-    if (apiCall.url === "/Authentication/GetEntityToken")
+    if (apiCall.url === "/Authentication/GetEntityToken" || apiCall.url === "/GameServerIdentity/AuthenticateGameServerWithCustomId")
```
amccalib, line 208: This is the one that should have been removed.

amccalib (changes requested):

amccalib (approved):

PR #832, opened 2022-10-30
[CSharpSDK-Draft] Adding Polly as a transport type

maryammadzadeh:
> This is an initial version of how to include Polly polices into the SDK. Without changing too much of the internals of the plugin manager, I've added a PluginContract type *(PlayFab_TransportWithPolly) for Polly that wraps around the existing PlayFabHttp and can have its policy overridden prior to the code being called that gets a plugin by name. I did this because the existing code requires the parameterless constructor (hence not wanting to change too much).

> Note: there are defaults defined for the common resiliency setting which would be good to go over with the SDK team to see is these values make sense, and or to tweak them. 

> My team has a service that can dogfood this in prod and would to generate a prerelease to test it out. Looking to see if that's possible or not a standard followed. Similar tests to the existing ones were added for the new plugin. 

Thanks in advance for the time taken to review this

### DIFF
--- a/targets/csharp/source/PlayFabSDK/source/PlayFabHttp/PlayFabPollyHttp.cs
+++ b/targets/csharp/source/PlayFabSDK/source/PlayFabHttp/PlayFabPollyHttp.cs
@@ -1,0 +1,107 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Threading.Tasks;
+using Microsoft.Xbox.Publishing.PlayFabCSharpSDK.Polly.Interfaces;
+using PlayFab;
+using PlayFab.Internal;
+using Polly;
+using Polly.Wrap;
+
+namespace PlayFab.Internal
+{
+    /// <summary>
+    /// A Polly wrapped version of PlayFab Transport plug in which used
+    /// when making http requests to PlayFab Main Server.
+    /// </summary>
+    public class PlayFabPollyHttp : PlayFabHttp, ITransportPlugin
+    {
+        /// <summary>
+        /// Http requests worth retrying
+        /// </summary>
+        HttpStatusCode[] httpStatusCodesWorthRetrying = {
+           HttpStatusCode.RequestTimeout,       // 408
+           HttpStatusCode.InternalServerError,  // 500
+           HttpStatusCode.BadGateway,           // 502
+           HttpStatusCode.ServiceUnavailable,   // 503
+           HttpStatusCode.GatewayTimeout        // 504
+        };
+
+        /// <summary>
+        /// Gets or set the name of the plug in. Used by the PluginManager when looking up
+        /// plugins upon request.
+        /// </summary>
+        public string Name;
+
+        /// <summary>
+        /// Gets the resilience policies defined.
+        /// </summary>
+        public AsyncPolicyWrap CommonResilience { get; private set; }
+
+        /// <summary>
+        /// Constructor for objects of type PollyTransportPlug.
+        /// <remarks>
+        /// Sets a default resilience policy with the fololwing common settings
+        /// 1) Sets the retry to 3 times and has an embedded backoff.
+        /// 2) > Sets a circuit breaker that will trigger is 25% of collapsed calls within 
+        ///      a 5 second window are failing.
+        ///    > Period for evaluation requires a burst of >=2RPS before evaluating breaker rule,
+        ///      requires a minimum of 10 requests in 5 seconds, and the circuit breaker will
+        ///      will be open for 20 second.
+        /// More information on the client can be found here: https://github.com/App-vNext/Polly
+        ///</remarks>
+        /// </summary>
+        public PlayFabPollyHttp()
+        {
+            var retryPolicy = Policy
+               .Handle<Exception>()
+                .Or<HttpRequestException>()
+                    .OrResult<HttpResponseMessage>(r => httpStatusCodesWorthRetrying.Contains(r.StatusCode))
+                .Or<TaskCanceledException>()
+                .Or<TimeoutRejectedException>()
+                .WaitAndRetry(3,
+                  retryAttempt => TimeSpan.FromSeconds(Math.Pow(2, retryAttempt))  // exponential back-off: 2, 4, 8 etc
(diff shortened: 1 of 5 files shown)


maryammadzadeh:
@microsoft-github-policy-service agree [company="Microsoft"]

maryammadzadeh:
@microsoft-github-policy-service agree company="Microsoft"

maryammadzadeh:
@microsoft-github-policy-service agree

targets/csharp/source/PlayFabSDK/source/PlayFabHttp/PlayFabPollyHttp.cs
```diff
@@ -0,0 +1,92 @@
...
+        ///</remarks>
+        /// </summary>
+        public PlayFabPollyHttp()
+        {
+            var retryPolicy = Policy
+               .Handle<Exception>()
+                .Or<HttpRequestException>()
+                .Or<TaskCanceledException>()
+                .Or<TimeoutRejectedException>()
+                .WaitAndRetryAsync(3, retryAttempt => TimeSpan.FromSeconds(Math.Pow(2, retryAttempt)));
```
maryammadzadeh, line 49: Notes from meeting
Consider using retry header from playfab,
Consider using jitter
Review pf documentation 20 second breaker
Determine if we need a new interface

targets/csharp/source/PlayFabSDK/source/PluginContract.cs
```diff
@@ -3,6 +3,7 @@
     public enum PluginContract
     {
         PlayFab_Serializer,
-        PlayFab_Transport
+        PlayFab_Transport,
+        PlayFab_TransportWithPolly
```
amccalib, line 7: I don't think you need another PluginContract here. The PlayFab_Transport contract should be sufficient (and allows for easy swapping between the two).

targets/csharp/source/PlayFabSDK/source/PlayFabHttp/PlayFabPollyHttp.cs
```diff
@@ -0,0 +1,92 @@
...
+using Polly;
+using Polly.Wrap;
+
+namespace PlayFab.Internal
+{
+    /// <summary>
+    /// A Polly wrapped version of PlayFab Transport plug in which used
+    /// when making http requests to PlayFab Main Server.
+    /// </summary>
+    public class PlayFabPollyHttp : PlayFabHttp, IPlayFabPlugin
```
amccalib, line 16: I think you want to inherit from ITransportPlugin, not IPlayFabPlugin. That should ensure this plugin can be used interchangably with the existing PlayFabSysHttp transport.

targets/csharp/source/PlayFabSDK/source/PluginManager.cs
```diff
@@ -56,6 +56,9 @@ private IPlayFabPlugin GetPluginInternal(PluginContract contract, string instanc
                     case PluginContract.PlayFab_Transport:
                         plugin = this.CreatePlugin<PlayFabSysHttp>();
                         break;
+                    case PluginContract.PlayFab_TransportWithPolly:
```
amccalib, line 59: Building on the other comments. For now, I would prefer if PluginManager didn't have any special knowledge of this plugin. Instead, games could call SetPlugin after creating an instance of PlayFabPollyHttp to override their transport.

targets/csharp/source/PlayFabSDK/source/Uunit/PluginManagerTest.cs
```diff
@@ -54,9 +54,11 @@ public void PluginManagerDefaultPlugins(UUnitTestContext testContext)
         {
             var playFabSerializer = PluginManager.GetPlugin<ISerializerPlugin>(PluginContract.PlayFab_Serializer);
             var playFabTransport = PluginManager.GetPlugin<ITransportPlugin>(PluginContract.PlayFab_Transport);
+            var playFabTransportWithPolly = PluginManager.GetPlugin<ITransportPlugin>(PluginContract.PlayFab_TransportWithPolly);
```
amccalib, line 57: Again, I don't think Polly should be a separate type of default plugin. It could eventually be the default Transport plugin, but when that happens it would be using the PlayFab_Transport contract.

targets/csharp/source/PlayFabSDK/source/Uunit/PluginManagerTest.cs
```diff
@@ -113,6 +115,34 @@ public void PluginManagerMultiplePluginsPerContract(UUnitTestContext testContext
             testContext.True(object.ReferenceEquals(actualTransport2, expectedTransport2));
             testContext.EndTest(UUnitFinishState.PASSED, null);
         }
+
+        /// <summary>
+        /// Test that plugin manager can set and return the polly plugin.
+        /// </summary>
+        [UUnitTest]
+        public void PluginManager_PollyPlugin_Succesful(UUnitTestContext testContext)
```
amccalib, line 123: This is a good test, but needs to be refactored to use the PlayFab_Transport plugin contract. In theory, this should be mimicking what a game would do to change their transport plugin from the default to your polly plugin.

amccalib (changes requested):

targets/csharp/source/PlayFabSDK/source/PluginManager.cs
maryammadzadeh: Sounds good, will adjust. I'll reping you once this has been dogfooded

targets/unity-v2/source/ExampleTestProject/Assets/Testing/Tests/Client/ClientApiTests.cs
```diff
@@ -83,7 +83,9 @@ private void SharedErrorCallback(PlayFabError error)
         public void InvalidLogin(UUnitTestContext testContext)
         {
             if (string.IsNullOrEmpty(_userEmail))
+            {
```
amccalib, line 86: I don't think this file needed to change (it's for Unity).

amccalib (approved):

maryammadzadeh:
Recreating a new PR since a second repo needs to be updated.

PR #833, opened 2022-10-31
Adding pipelines scripts for setup & generation

AlonsoMondal:
Adding pipelines scripts for repo setup & SDK generation to remove dependencies to the private repository JenkinsSdkSetupScripts.

### DIFF
--- a/targets/java/source_shared/src/test/java/com/playfab/test/PlayFabApiTest.java.ejs
+++ b/targets/java/source_shared/src/test/java/com/playfab/test/PlayFabApiTest.java.ejs
@@ -103,6 +103,11 @@ public class PlayFabApiTest
         TitleData resultData = gson.fromJson(testTitleJson, new TypeToken<TitleData>(){}.getType());
         PlayFabSettings.TitleId = resultData.titleId;
         USER_EMAIL = resultData.userEmail;
+        String devSecretKey = env.get("DEVELOPER_SECRET_KEY_JAVA_SDK");
+        if (devSecretKey != null && !devSecretKey.equals("")) {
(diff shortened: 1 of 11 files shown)


targets/java/source_shared/src/test/java/com/playfab/test/PlayFabApiTest.java.ejs
```diff
@@ -103,6 +103,10 @@ public class PlayFabApiTest
         TitleData resultData = gson.fromJson(testTitleJson, new TypeToken<TitleData>(){}.getType());
         PlayFabSettings.TitleId = resultData.titleId;
         USER_EMAIL = resultData.userEmail;
+        String devSecretKey = env.get("DEVELOPER_SECRET_KEY_JAVA_SDK");
```
amccalib, line 106: Was this change intended to be included? I don't see how it's related to the Jenkins script move.

amccalib (approved):

targets/java/source_shared/src/test/java/com/playfab/test/PlayFabApiTest.java.ejs
AlonsoMondal: Yes, JenkinsSdkSetupScripts contained the title data including the developer secret key. In order to remove the dependency to that repo & not have the secret published in a public repo, the pipelines set it as an env var. Suggestions are welcome

amccalib: Oh, I wasn't aware of that. Your solution is fine.

PR #837, opened 2022-11-14
Fixing bug of calling HaveResponse without the needed GetResponse first

AlonsoMondal:


targets/unity-v2/source/ExampleTestProject/Assets/PlayFabSDK/Shared/Internal/PlayFabHttp/PlayFabWebRequest.cs
```diff
@@ -251,6 +251,7 @@ private static void WorkerThreadMainLoop()
                             case HttpRequestState.Idle:
                                 Post(localActiveRequests[i]); break;
                             case HttpRequestState.Sent:
+                                localActiveRequests[i].HttpRequest.GetResponseAsync();
```
amccalib, line 254: Do you know if it's okay to call this multiple times on the same request? Since this is in our worker loop, it's going to get called every update. I'm worried we're creating a new Task with every call, but it's not clear from the documentation. Maybe the CallRequestContainer could be expanded to include tracking of whether we've already called GetRequetAsync?

amccalib (changes requested):

amccalib (approved):

PR #845, opened 2022-11-29
[CSharpSDK] Adding Polly as a transport type

maryammadzadeh:
This is an initial version of how to include Polly polices into the SDK. There are defaults defined for the common resiliency setting which have been covered with the SDK team. Improvements can always be added!

Thanks in advance for the time taken to review this. 

This has been tested against the PlayFabSdk.All-Unit test solution

Note existing testing infrastructure required synchronization for test method, so I conformed. Would like to make this async one day but will let SDK team look into it.


toddbellMSFT (approved):
The pipeline passes, but it is very odd that a test passing twice is outputting a failure to the console. I'll write up a bug to look into the issue, failing tests should ideally fail the pipeline, I think we made an exception for these expected failure tests that adds noise to our test output.

targets/unity-v2/source/ExampleTestProject/Assets/PlayFabSDK/Shared/Public/PluginManager.cs
```diff
@@ -35,12 +36,25 @@ private PluginManager()
...
         }
 
+        /// <summary>
+        /// Add additional custom plugins to the collection
+        /// If a plugin with specified contract and optional instance name already exists, it will be replaced with specified instance.
+        /// </summary>
+        /// <param name="plugin">The plugin instance.</param>
+        /// <param name="contract">The app contract of plugin.</param>
+        /// <param name="instanceName">The instance name. Instance names allow to have multiple plugins with the same contract.</param>
+        public static void Add(IPlayFabPlugin plugin, PluginContract contract, string instanceName)
```
maryammadzadeh, line 52: this is not needed

PR #852, opened 2023-01-03
Update XML output logic

toddbellMSFT:
attempt to write output to the same xml file every time we run a test. Since ADO uploads this file to a particular location and we have this hardcoded limit in place, there have been pipelines that failed due to infrastructure issues rather than failing because of what we are testing.

### DIFF
--- a/JenkinsConsoleUtility/jcuSrc/Commands/CloudScriptListener.cs
+++ b/JenkinsConsoleUtility/jcuSrc/Commands/CloudScriptListener.cs
@@ -136,7 +136,7 @@ namespace JenkinsConsoleUtility.Commands
             if (!callResult || testResults == null)
                 return 1;
             JcuUtil.FancyWriteToConsole(ConsoleColor.Gray, "Writing test results: " + outputFileFullPath);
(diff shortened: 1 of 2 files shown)


JenkinsConsoleUtility/jcuSrc/Commands/CloudScriptListener.cs
```diff
@@ -128,20 +128,15 @@ private int FetchTestResult(string workspacePath, TestTitleData testTitleData)
...
+
+            string outputFileName = "ListenCsResult.xml";
+            
+            outputFileFullPath = Path.Combine(workspacePath, outputFileName);
 
             if (!callResult || testResults == null)
                 return 1;
             JcuUtil.FancyWriteToConsole(ConsoleColor.Gray, "Writing test results: " + outputFileFullPath);
-            return JUnitXml.WriteXmlFile(outputFileFullPath, testResults, true);
+            return JUnitXml.WriteXmlFile(outputFileFullPath, testResults, false);
```
AlonsoMondal, line 139: This will cause for the new results to be appended at the end of the file. Is this desired behavior?

AlonsoMondal (approved):

PR #853, opened 2023-01-04
Add Unity Test jitter

toddbellMSFT:
Our Unit Test title is getting throttled, especially for Unity tests. 

I'm adding a random wait at the end of each test to prevent parallel tests from hitting the test title at the same time every night.

### DIFF
--- a/targets/unity-v2/source/ExampleTestProject/Assets/Testing/Tests/Shared/Uunit/UUnitTestContext.cs
+++ b/targets/unity-v2/source/ExampleTestProject/Assets/Testing/Tests/Shared/Uunit/UUnitTestContext.cs
@@ -55,7 +55,7 @@ namespace PlayFab.UUnit
         {
             // Add a random waiting jitter to each test end to prevent multiplatform parallel tests from throttling the test title.
             Random rand = new Random();
-            System.Threading.Thread.Sleep(rand.Next(250));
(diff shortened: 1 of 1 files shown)


targets/unity-v2/source/ExampleTestProject/Assets/Testing/Tests/Shared/Uunit/UUnitTestContext.cs
```diff
@@ -53,6 +53,10 @@ public void AttemptRetry()
 
         public void EndTest(UUnitFinishState finishState, string resultMsg)
         {
+            // Add a random waiting jitter to each test end to prevent multiplatform parallel tests from throttling the test title.
+            Random rand = new Random();
+            System.Threading.Thread.Sleep(rand.Next(250));
```
kylemc, line 58: 250 milliseconds is not much when throttling is only recalculated each second. Consider something in the 1-3 second range.

toddbellMSFT: Okay, I was noticing a hiccup in the unity UI when I made it 1 second, but it may be what we have to deal with to prevent the parallelism issue.

kylemc (approved):

AlonsoMondal (approved):

PR #857, opened 2023-04-18
enable relative ejs include paths

tculotta:
This enables <%- include('filename') %> usage (without absolute paths) which can prevent bloat in template files and hopefully make them a little more readable by pulling out reused or bespoke chunks of code.

In theory, just adding the filename arg here probably shouldn't mess any current usage up, but I added the switch to ensure other projects are unchanged by this unless they opt in.


jplafonta (approved):
:shipit:

PR #867, opened 2023-10-31
[PlayFab Polly Sdk] Adding new type to ensure the set plugin is initialized correctly

maryammadzadeh:
We have seen in tests that the client can't pick up the correct plugin. A partner is seeing this and we  need to ensure we have retry logic working for their calls to login and lobby

Doing some debugging and i think my original ask to add a new enum for thePluginContract is still required. I am 100% sure the polly code is not hit because it doesnt know about the extension class i added. 

In this change, we explicitly make sure we create a PollyFabPollyHttp plugin

CSharpSDK/PlayFabSDK/source/PlayFabHttp/PlayFabPollyHttp.cs at 8581a8ae707ff331cbc196c19a9f4c03d8cf9c72 · PlayFab/CSharpSDK (github.com)
The manager that cant know about my class, line, line 51
CSharpSDK/PlayFabSDK/source/PluginManager.cs at 8581a8ae707ff331cbc196c19a9f4c03d8cf9c72 · PlayFab/CSharpSDK (github.com)

### DIFF
--- a/targets/csharp/source/PlayFabSDK/source/PlayFabHttp/PlayFabPollyHttp.cs
+++ b/targets/csharp/source/PlayFabSDK/source/PlayFabHttp/PlayFabPollyHttp.cs
@@ -28,7 +28,7 @@ namespace PlayFab.Internal
             500, //HttpStatusCode.InternalServerError
             502, //HttpStatusCode.BadGateway
             503, //HttpStatusCode.ServiceUnavailable
-            504 //HttpStatusCode.GatewayTimeout
+            504  //HttpStatusCode.GatewayTimeout
         };
 
         /// <summary>
@@ -60,20 +60,22 @@ namespace PlayFab.Internal
             var jitterer = new Random();
             var retryPolicy = Policy
                .Handle<Exception>()
-               .OrResult<object>(r => r != null && HttpStatusCodesWorthRetrying.Contains((r as PlayFabError).HttpCode))
+                .OrResult<object>(r => r != null && (r as PlayFabError) != null && HttpStatusCodesWorthRetrying.Contains((r as PlayFabError).HttpCode))
                     .WaitAndRetryAsync(1,
                       retryAttempt => TimeSpan.FromSeconds(Math.Pow(2, retryAttempt))  // exponential back-off: 2, 4, 8 etc
                                 + TimeSpan.FromMilliseconds(jitterer.Next(0, 1000)));  // plus some jitter: up to 1 second
 
             var breakerPolicy = Policy.Handle<Exception>()
-                .OrResult<object>(r => r != null && HttpStatusCodesWorthRetrying.Contains((r as PlayFabError).HttpCode))
+                .OrResult<object>(r => r != null && (r as PlayFabError) != null && HttpStatusCodesWorthRetrying.Contains((r as PlayFabError).HttpCode))
                 .AdvancedCircuitBreakerAsync(
                     failureThreshold: 0.25,
                     samplingDuration: TimeSpan.FromSeconds(5),
                     minimumThroughput: 2,
                     durationOfBreak: TimeSpan.FromSeconds(20));
 
-            CommonResilience = breakerPolicy.WrapAsync(retryPolicy);
+            CommonResilience = retryPolicy.WrapAsync(breakerPolicy);
+
+            Name = "PlayFabWithPolly";
         }
 
         /// <inheritdoc/>
@@ -82,8 +84,7 @@ namespace PlayFab.Internal
             object doPostResult = await CommonResilience
                 .ExecuteAsync(async () =>
                 {
-                    var result = await base.DoPost(fullPath, request, headers) as PlayFabError;
-                    return result;
+                    return await base.DoPost(fullPath, request, headers);
                 });
 
             return doPostResult;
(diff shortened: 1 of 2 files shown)


targets/csharp/source/PlayFabSDK/source/PlayFabHttp/PlayFabPollyHttp.cs
```diff
@@ -82,7 +82,7 @@ public new async Task<object> DoPost(string fullPath, object request, Dictionary
             object doPostResult = await CommonResilience
                 .ExecuteAsync(async () =>
                 {
-                    var result = await base.DoPost(fullPath, request, headers) as PlayFabError;
+                    var result = await base.DoPost(fullPath, request, headers) as PlayFabResultCommon;
```
amccalib, line 85: I'm not sure this works. When base.DoPost succeeds, the return object is a string. It's only a well-formed object on failure (which is admittedly weird). My local fix was to not cast here, but instead check the type in the retryPolicy and breakerPolicy handlers and only apply policy if the type was PlayFabError.

targets/csharp/source/PlayFabSDK/source/PluginContract.cs
```diff
@@ -3,6 +3,7 @@
     public enum PluginContract
     {
         PlayFab_Serializer,
-        PlayFab_Transport
+        PlayFab_Transport,
+        PlayFab_PollyTransport
```
amccalib, line 7: As mentioned offline, I'd prefer to not add an additional contract type (since it would be misaligned with the other languages). Instead, I'd prefer we add comments and possibly deprecation notes around the instanceName to avoid future confusion.

amccalib (changes requested):

targets/csharp/source/PlayFabSDK/source/PlayFabHttp/PlayFabPollyHttp.cs
maryammadzadeh: Actually i should update this test because what's happening now is the test also shows we retry until we succeeds, and tries more than 10 times. Ill add an assertion for that check, too.

targets/csharp/source/PlayFabSDK/source/Uunit/tests/HttpTests.cs
```diff
@@ -81,7 +81,6 @@ public void TestPluginWithoutPolly_OnSuccess_200Response(UUnitTestContext testCo
         [UUnitTest]
         public void TestPluginWithPolly_Success(UUnitTestContext testContext)
         {
-            PluginManager.SetPlugin(mockHttpPluginWithoutPolly, PluginContract.PlayFab_Transport);
             PluginManager.SetPlugin(mockHttpPluginWithPolly, PluginContract.PlayFab_Transport, "PluginWithPolly");
```
amccalib, line 84: This test won't actually test Polly. Since you're providing an instanceName, the clientApi call will just use the default transport plugin.

targets/csharp/source/PlayFabSDK/source/PlayFabHttp/PlayFabPollyHttp.cs
```diff
@@ -92,12 +93,22 @@ public new async Task<object> DoPost(string fullPath, object request, Dictionary
...
+        private async Task<object> PollyDoPostAsync(string fullPath, object request, Dictionary<string, string> headers)
+        {
+            var result = await base.DoPost(fullPath, request, headers);
+            var resultAsError = result as PlayFabError;
+            if (resultAsError != null)
+            {
+                return resultAsError;
+            }
+
+            return result as PlayFabResultCommon;
```
amccalib, line 111: I still have concerns about this. I'm curious if this cast actually works in the success case. The success case test below may not have actually been invoking the polly plugin.

amccalib (changes requested):

targets/csharp/source/PlayFabSDK/source/Uunit/tests/HttpTests.cs
maryammadzadeh: How else do you differentiate the same PluginContract type?

maryammadzadeh: Ok i see we just cant have more that one Contract of type PluginContract.PlayFabTransport.

targets/csharp/source/PlayFabSDK/source/PlayFabHttp/PlayFabPollyHttp.cs
maryammadzadeh: https://learn.microsoft.com/en-us/rest/api/playfab/client/authentication/get-title-public-key?view=playfab-rest  looks like this api doesnt work anymore. Its the only api that requires no auth, which was what i intended to get. 

Found another endpoint i can use.

maryammadzadeh: Updated to check for success case.

targets/csharp/source/PlayFabSDK/source/PlayFabHttp/PlayFabPollyHttp.cs
maryammadzadeh: Ok if we dont use the instancename string, then this works. Thanks!!

amccalib (approved):

PR #869, opened 2023-11-02
[PollyPolicies] Remove circut breaking. We only want retry for the client

maryammadzadeh:
The circuit breaker could cause more issues for a client than we want. Retry is whats important


targets/csharp/source/PlayFabSDK/source/PlayFabHttp/PlayFabPollyHttp.cs
```diff
@@ -11,8 +11,7 @@
 using System.Threading.Tasks;
 
 namespace PlayFab.Internal
-{
-    /// <summary>
+{   /// <summary>
```
amccalib, line 14: unintentional whitespace change

amccalib (approved):
Looks fine. One small whitespace nit

AlonsoMondal (approved):

PR #873, opened 2024-01-17
Adding platform information

NandoCT1014:
Adding platform information to query string using Unity's Application.platform() enum values and GDK platform directives.

### DIFF
--- a/targets/unity-v2/source/ExampleTestProject/Assets/PlayFabSDK/Shared/Public/PlayFabSettings.cs.ejs
+++ b/targets/unity-v2/source/ExampleTestProject/Assets/PlayFabSDK/Shared/Public/PlayFabSettings.cs.ejs
@@ -259,7 +259,7 @@ namespace PlayFab
             }
 
             sb.Append(firstParam ? "?" : "&");
-            sb.Append("engineVersion=").Append(EngineVersion);
+            sb.Append("engine=").Append(EngineVersion);
             sb.Append("&platform=").Append(PlatformString);
 
             return sb.ToString();
--- a/targets/unity-v2/source/ExampleTestProject/Assets/PlayFabSDK/Shared/Public/PlayFabSettings.cs.ejs
+++ b/targets/unity-v2/source/ExampleTestProject/Assets/PlayFabSDK/Shared/Public/PlayFabSettings.cs.ejs
@@ -47,6 +47,7 @@ namespace PlayFab
         public const string BuildIdentifier = "<%- buildIdentifier %>";
         public const string VersionString = "UnitySDK-<%- sdkVersion %>";
         public static string EngineVersion = UnityEngine.Application.unityVersion;
+        public static string PlatformString;
 
         public const string DefaultPlayFabApiUrl = "playfabapi.com";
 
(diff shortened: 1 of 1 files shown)


targets/unity-v2/source/ExampleTestProject/Assets/PlayFabSDK/Shared/Public/PlayFabSettings.cs.ejs
```diff
@@ -29,7 +29,29 @@ namespace PlayFab
 
     public static class PlayFabSettings
     {
-        static PlayFabSettings() { }
+        static PlayFabSettings() 
+        {
+#if UNITY_GAMECORE || UNITY_GAMECORE_XBOXONE || UNITY_GAMECORE_SCARLETT || MICROSOFT_GAME_CORE
+            PlatformString = "GDK";
+#else
+        switch (Application.platform)
```
tculotta, line 37: What about other platforms?

targets/unity-v2/source/ExampleTestProject/Assets/PlayFabSDK/Shared/Public/PlayFabSettings.cs.ejs
```diff
@@ -29,7 +29,29 @@ namespace PlayFab
 
     public static class PlayFabSettings
     {
-        static PlayFabSettings() { }
+        static PlayFabSettings() 
+        {
+#if UNITY_GAMECORE || UNITY_GAMECORE_XBOXONE || UNITY_GAMECORE_SCARLETT || MICROSOFT_GAME_CORE
```
tculotta, line 34: Do we care about differentiating between these at all?

jasonsandlin (approved):

targets/unity-v2/source/ExampleTestProject/Assets/PlayFabSDK/Shared/Public/PlayFabSettings.cs.ejs
NandoCT1014: Other platforms like PS5, PS4, Switch, Android, Windows, and iOS should set by converting the Platform.Runtime enum to a string directly if none of the GDK platform macros are set.

targets/unity-v2/source/ExampleTestProject/Assets/PlayFabSDK/Shared/Public/PlayFabSettings.cs.ejs
NandoCT1014: I didn't differentiate between console and pc in this case as I'm following the pattern in setting the GDK platform macro in PF C SDK

targets/unity-v2/source/ExampleTestProject/Assets/PlayFabSDK/Shared/Public/PlayFabSettings.cs.ejs
tculotta: Oh I see now! I'd misread the default block 👍

tculotta (approved):
:shipit:

PR #876, opened 2024-02-16
Key board focus issue 2

vishnuchoutapelly:
Key board focus indicator issue for multiple controls


targets/unity-v2/source/ExampleTestProject/Assets/PlayFabEditorExtensions/Editor/Scripts/Components/TitleDataViewer.cs
```diff
@@ -14,21 +14,113 @@ public class TitleDataViewer : UnityEditor.Editor
...
+                        case 6:
+                            EditorGUI.FocusTextInControl("refresh");
+                            focusIndex = 0;
+                            break;
+
+                    }
+                }
+                else
+                {
+                    switch (focusIndex)
```
amccalib, line 77: I expected this switch block to basically be the opposite of the one above. Is there a reason it has fewer cases? Why is "new_value_text_field" missing?

amccalib: I think there might be other index numbering issues here as well.

amccalib (changes requested):

targets/unity-v2/source/ExampleTestProject/Assets/PlayFabEditorExtensions/Editor/Scripts/Components/TitleDataViewer.cs
vishnuchoutapelly: Missing code snipped has been added.

amccalib (approved):

PR #881, opened 2024-03-14
Update ValidateEntityTokenAsync to use the instance authenticationContext values

AlonsoMondal:
Added the use of the requestContext local variable that contains the PlayFabAuthenticationContext object.
The PlayFabAuthenticationContext instance object isn't used to get the auth context. If we do not pass the auth context in the method request parameters, it does not fall back to use the auth context from the class instance.
Thanks @fernandezafb for the work: https://github.com/PlayFab/CSharpSDK/pull/125


AlonsoMondal:
Preview of the changes through generator: https://github.com/PlayFab/CSharpSDK/pull/126

amccalib (approved):

PR #884, opened 2024-04-24
Fixed scripting defines for dedicated server

cabbagegod:
Unity has changed how build targets are defined and have set `GetScriptingDefineSymbolsForGroup` and `SetScriptingDefineSymbolsForGroup` for future deprecation. This introduces a bug in PlayFab, causing the editor settings not to update the scripting defines when a "Dedicated Server" build is selected, which was released in 2021.3.

This PR should fix the problem with dedicated servers for versions 2021.3+, and cover backwards compatibility for older Unity versions. Additionally, 2021.2+ should now use the non-deprecated methods for scripting define symbols.

The large use of conditional compilation is due to an oversight(?) from Unity, where this is no easy way to get your current build setting for dedicated servers, because dedicated servers are actually a sub target.


cabbagegod:
@microsoft-github-policy-service agree

cabbagegod:
Fixed a bug. Still needs to be tested on a Unity version before 2021.2

